PR #16, opened 2020-02-23
Add initial prometheus endpoint support. Fixes #4

### LINKED ISSUE #4: Prometheus support
It would be great to have some Prometheus metrics exposed....

Rich7690:
Took a stab at implementing prometheus metrics support asked for in #4. Added some initial counters that seemed to make sense: total queries, total blocked, total cache hit, total cache miss

I'm a fairly novice Golang programmer looking to get some experience in open source so any feedback/suggestions are welcome.


0xERR0R:
@Rich7690: Thank you for the PR! I could have a look at your change, it looks good for me. Question for my understanding: Do actual metrics like totalQuery, totalBlock etc. count indefinitely from start? Or would prometheus do some "magic" and display only values per day? I saw pihole exporter for prometheus (https://github.com/eko/pihole-exporter), its metrics are always per day (dns_queries_today for example), this makes sense for me. Blocky calculates statistics for 24 hours, would it make sense to export this data?

Rich7690:
> @Rich7690: Thank you for the PR! I could have a look at your change, it looks good for me. Question for my understanding: Do actual metrics like totalQuery, totalBlock etc. count indefinitely from start? Or would prometheus do some "magic" and display only values per day? I saw pihole exporter for prometheus (https://github.com/eko/pihole-exporter), its metrics are always per day (dns_queries_today for example), this makes sense for me. Blocky calculates statistics for 24 hours, would it make sense to export this data?

Yea, so generally speaking prometheus counts up indefinitely, but from what I can see its using an atomic uint64 on the backend so it will obviously rollover after that. This is naturally intended by prometheus's design and scrapers are supposed to handle resets. Same behavior when you cycle the process.

As far as the 24 hour stats, the good part about this format is you can have the user side apps do this calculation for you; visualization tools like grafana for example let you do all kinds of fancy aggregations based on your needs. So you can aggregate every 5 minutes or every 24 hours. Its up to the user.

0xERR0R:
Thank you for explanation

PR #24, opened 2020-03-10
Support multiple instances of blocky

billimek:
When running multiple instances of blocky and emitting metrics to prometheus, the grafana dashboard will have trouble when prometheus has multiple instances of blocky collected.

This pull request will remedy some of the problematic queries to properly handle multiple instances.

For example, with 3 instances of blocky running and emitting metrics the proposed changes shows this:
image

Running just one instance with the proposed changes shows this:
image

### DIFF
--- a/docs/blocky-grafana.json
+++ b/docs/blocky-grafana.json
@@ -82,7 +82,7 @@
       "tableColumn": "",
       "targets": [
         {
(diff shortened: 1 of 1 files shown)


onedr0p:
Looks great @billimek thanks!

PR #30, opened 2020-04-16
Update README for correct refreshPeriod

izzeho:
Current refreshPeriod example sets refresh to 1 minute. This results in DNS queries failing during Blocky reload 4 seconds of every minute.


0xERR0R:
You're right, refresh time of 1 minute is not useful for production environment. Thank you for PR!

PR #33, opened 2020-04-26
Dockerfile for adding Prom. access.

lakestonelabs:
Awesome little DNS proxy!  This just adds a line in the Dockerfile to allow Prometheus to pull stats.


0xERR0R:
Thank you!

PR #34, opened 2020-04-27
Expanded compose file on setting-up Prom. and Grafana.

lakestonelabs:


0xERR0R:
Thanks for your work! Your configuration is very useful for people who don't have a running prometheus/grafana stack. I would like to keep the minimalist working example of docker-compose on the first page (readme). The idea to create more examples with advanced configuration is great, is it OK for you, if I move your example to the wiki and link them from README.md as "advanced configuration"?

PR #115, opened 2020-12-09
Log output to stdout instead of stderr

kbadk:
By default, logrus outputs to `stderr`. Other than being semantically incorrect for the output generated, it also makes it hard to grep through the output as `stderr` messages aren't filtered by grep.


0xERR0R:
It makes sense, thanks for your work!

PR #132, opened 2021-02-16
Fix indentation in example config.yml

schlamar:


0xERR0R:
Thank you, there is something went wrong.. The yml version (https://github.com/0xERR0R/blocky/blob/development/docs/config.yml) looks good, but config in the README was indeed broken.

PR #142, opened 2021-02-22
use package logger instance instead of global logrus instance

nicolas-martin:
Created internal logger instance so that it doesn't interfere with the users global instance when used as a package. This was very annoying for us since we wanted our code to have debug level but that would also set blocky's log level which is too verbose and not relevant for us.

Seems like there are some test are failing. Not too sure how to fix them.

### DIFF
--- a/api/api_suite_test.go
+++ b/api/api_suite_test.go
@@ -1,15 +1,15 @@
 package api_test
 
 import (
+	"blocky/log"
 	"testing"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/sirupsen/logrus"
 )
 
 func TestResolver(t *testing.T) {
-	logrus.SetLevel(logrus.WarnLevel)
+	log.NewLogger("Warn", "text")
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "API Suite")
 }
--- a/cmd/blocking.go
+++ b/cmd/blocking.go
@@ -7,7 +7,8 @@ import (
 	"fmt"
 	"net/http"
 
-	log "github.com/sirupsen/logrus"
+	"blocky/log"
+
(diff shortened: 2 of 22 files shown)


0xERR0R:
I'm not sure, if this is a good solution -> it makes the code nor really readable. Maybe you can explain, what are you trying to achieve? Are you trying to embed blocky as external library?

nicolas-martin:
Yes we're running blocky as a package inside our own executable which works fine. The problem is the use of the global logrus variable. It could also be a blocky issue if any of its dependency use the same logrus variable. Obviously having two packages setting the same global variable isn't a good practice and won't get consistent logging.

Most package and libraries either use this approach or allow for a logging interface to be passed as param.

0xERR0R:
I understand your point. Blocky was implemented to be used as a standalone binary. I'll take a look, on your changes. As workaround / other approach, you can also use a dedicated logger in your application. In this case your logging will be independent from other libs which are using logrus and you can configure the log level only for your code

nicolas-martin:
Yup we will also implement a similar approach to what I suggested here. I thought it would be useful for blocky to have that protection.

The changes look bigger than what they are. It's really the same principle but the global variable is held in blocky instead.

0xERR0R:
Some tests check, if logger.Fatal was called (typically in the error case). This is possible by replacing the ExitFunc with one, which only sets the boolean variable. With your logging approach, the logging will be initialized again (for example in server.go) and the ExitFunc will be replaced again with original one. I'll try to find a way around

0xERR0R:
I made some minor refactorings and fixed the tests -> Global logger instance will be initialized only once and can be reconfigured later. Please check, if this approach still meet your needs.

nicolas-martin:
That works, thanks for the update!

PR #171, opened 2021-04-05
Add support of disbale ipv6(AAAA) query

moolex:
Scenes like this https://github.com/AdguardTeam/AdGuardHome/issues/473

Also my network have no IPv6 internet but blocky receive lots AAAA query, i hope to ignore these queries


0xERR0R:
Thanks for your work! It looks good at first glance, I'll test and review it later.

PR #180, opened 2021-04-14
feat(docker): add tini to container image

onedr0p:
Hi :wave: Small change here. I created an `ENTRYPOINT` with [tini](https://github.com/krallin/tini) and updated the `CMD` in the `Dockerfile`. tini is already included in the `Docker` CRI, however some people (like me) use a different CRI like `containerd` or `cri-o` where we don't have access to an `--init` flag. 

`tini` also has other benefits as described in this [Github issue](https://github.com/krallin/tini/issues/8#issuecomment-146135930).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ LABEL org.opencontainers.image.source="https://github.com/0xERR0R/blocky" \
       org.opencontainers.image.url="https://github.com/0xERR0R/blocky" \
       org.opencontainers.image.title="DNS proxy as ad-blocker for local network"
 
-RUN apk add --no-cache bind-tools
+RUN apk add --no-cache bind-tools tini
 COPY --from=build-env /src/bin/blocky /app/blocky
 
 # the timezone data:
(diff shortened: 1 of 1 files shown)


0xERR0R:
Hi, thanks for your suggestion. As far as I know, init process is only needed if the application runs with multiple processes (e.g. started from a bash process or spans own new processes during the runtime). Blocky runs as self-contained binary which is executed directly, without any wrapper scripts. At runtime, no process will be started by the application. IMHO there is no need for an additional init process, this would just make a little overhead (running process with some memory consumption) and introduce another dependency, I don't see any benefit. What do you think?

I tried to find some "best practices" for Go and Docker:
- https://blog.golang.org/docker
- https://github.com/thockin/go-build-template
- https://levelup.gitconnected.com/complete-guide-to-create-docker-container-for-your-golang-application-80f3fb59a15e

They all are using go binary without init process.

onedr0p:
I think it's developer preference, you'll find some projects written in XYZ lang using tini in their Dockerfiles and some that don't. I've been pretty happy including tini in my Dockerfiles as a default because the overhead and memory footprint is really miniscule.

If you don't think their could be benefit here that's fine. :)

0xERR0R:
Thanks for you work!

Issue #182, opened 2021-04-15
Timestamps in log


fmarcia:
When used as a systemd service, timestamps are already available in journal.
Could there be an option to disable those provided by blocky in order to not get them twice?
Thanks for your hard work!

0xERR0R:
closed by #184

PR #184, opened 2021-04-16
#182 -  added config option for logTimestamp

Harnish:
Adds config option to send to the logger and set default to true, making it work exactly the same by default unless you set to false.  #182


0xERR0R:
Thanks!

PR #186, opened 2021-04-18
Make cache for all query types

### LINKED ISSUE #185: Cache more types
Current blocky only cache A/AAAA queries, but other types should also be cached.

refer to https://github.com/AdguardTeam/dnsproxy/blob/48200fa4f8835dc622a06a08f30a3485d77105cf/proxy/cache.go#L80

moolex:
Resolve #185

### DIFF
--- a/util/common.go
+++ b/util/common.go
@@ -177,24 +177,22 @@ func Chunks(s string, chunkSize int) []string {
 	return chunks
 }
 
-// GenerateCacheKey return cacheKey by query type/class/domain
-func GenerateCacheKey(qType uint16, qClass uint16, qName string) string {
-	b := make([]byte, 2+2+len(qName))
+// GenerateCacheKey return cacheKey by query type/domain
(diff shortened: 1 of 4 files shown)


util/common.go
```diff
@@ -175,3 +176,25 @@ func Chunks(s string, chunkSize int) []string {
 
 	return chunks
 }
+
+// GenerateCacheKey return cacheKey by query type/class/domain
+func GenerateCacheKey(qType uint16, qClass uint16, qName string) string {
```
0xERR0R, line 181: Class type can be omitted, IN is used in 99.9%, this would save 2 bytes per cache entry

0xERR0R (commented):
Good job, thanks! Nice idea to put 2 caches together and use one cache with calculated cache key. I think the class attribute can be omitted, since it is always "IN". This would save 2 bytes per cache entry

0xERR0R (approved):

Issue #81, opened 2020-07-02
Enhancement: Configuration of different upstream DNS servers for different clients


adb76:
I really like Blocky and the great configuration options. Because it is so slim and fast I changed my LAN setup from Adguard to Blocky. There is only one feature missing in Blocky for me: The possibility to have different upstream DNS servers for different clients. For my understanding it is currently not possible in Blocky to have something like:

```
externalResolvers:  
      default:     # Default for all requests
         - tcp: 46.182.19.48
         - tcp: 80.241.218.68
      desktop.fritz.box:   # Requests coming from desktop.fritz.box will be forwarded to
          - tcp-tls:fdns1.dismail.de:853
          - https://dns.digitale-gesellschaft.ch/dns-query
```

Or maybe I missed something in the configuration options?

0xERR0R:
This is correct, currently, there is no configuration option to achieve this. Could you describe your use case please? Maybe there exist some other options?

adb76:
The idea behind the use case is:

- I want to use the default upstream DNS Server from my internet provider for most of my devices
- For some devices (e.g. video streaming devices) I want to set a special DNS to solve geo location issues ;-)
- I don't won't to set the special DNS server to all of my devices, because of latency etc.
- Of course it is possible to set that DNS server in the configuration of the client, but it is much more comfortable to do that centrally, the way for example AdGuard offers it.

0xERR0R:
#43

Jogai:
It it something you're considering adding? Where should we vote? Here of in #43?

0xERR0R:
Yes, this feature is useful and I'll implement it in the (near) future. I would like to avoid breaking changes in the configuration (like example in the description -> this will break old installations), so maybe new configuration option is needed. You can vote here, I'll close #43

Jogai:
Sounds good! Maybe follow what you did for the client groups for blocking, then you'd end up with a config like this, which wouldnt break an existing config:
```
upstream:
  externalResolvers:
    - 1.1.1.3
    - 1.0.0.3
  malwareBlockingResolvers:
    - 1.1.1.2
    - 1.0.0.2
resolvers:
    default:
      - externalResolvers
    nsfw*:
      - malwareBlockingResolvers
```

0xERR0R:
I think, grouping upstream resolvers is a bit over-complicated. In case of blocking goups -> it is useful, since your client can be in several groups. For upstream resolvers there should be always 1:1 mapping.

[...]

PR #193, opened 2021-04-21
#81: Configuration of different upstream DNS servers for different clients

0xERR0R:


wwwlde:
I think this implementation is incomplete. At the very least, a separate cache should be formed for the client. Unfortunately, it doesn't work as I expected. Here is my configuration.

```yaml
---
minTlsServeVersion: 1.2
connectIPVersion: v4
bootstrapDns:
  - tcp+udp:1.1.1.1
ports:
  dns: 53
  http: 4000
  https: 8443
log:
  level: debug
  format: json
  timestamp: false
  privacy: true
upstreams:
  timeout: 1s
  groups:
    default:
      - https://1.1.1.3/dns-query
      - https://1.0.0.3/dns-query
    192.168.0.203:
      - tcp-tls:1.1.1.1:853
      - tcp-tls:1.0.0.1:853
prometheus:
  enable: true
  path: /metrics
caching:
  minTime: 30m
  maxTime: 60m
  prefetching: true
```

If the first request is made from an allowed IP, then other clients receive the response from the cache :-(

PR #195, opened 2021-04-29
Added version to web interface.

Harnish:
I broke this out to its own little change since I wanted to get your feedback on using the makefile to put the version into server.go vs passing it along.  

 I am trying to decom my project https://github.com/Harnish/https-dns-proxy and move the fun bits into blocky.


0xERR0R:
I moved version and build time to the util package (to avoid duplication of variables) and tested your PR -> looks good. I added also build time to the web interface, so the web UI shows the same information as the log file.

Thanks for you work!

0xERR0R:
Closes #203

PR #198, opened 2021-05-03
Make prefetch fully configurable

moolex:
add prefetchExpires and prefetchThreshold option (instead of constant value)


0xERR0R:
Thanks, it looks good! 👍

PR #199, opened 2021-05-03
WhitelistOnly Fix for multiple entries

c-f:
## Problem 
WhitelistOnly is only used, if no more than two whitelists are available or all clients include all whitelists. 

Look at the following config snippet. Two Groups are defined, each with their own set of "whitelist" entries. 
```yaml
blocking:
 whiteLists:
   net_one:
      - net_one_whitelist.txt
   net_two:
      - net_two_whitelist.txt
  clientGroupsBlock:
    one-client*:
       - net_one
    two-client*:
       - net_two

clientLookup:
 clients:
   one-client-1:  [127.0.0.1]
   two-client-2:  [127.0.0.2]
```

If multiple WhitelistOnly groups are defined, then every client, which do NOT include ALL whitelists, can resolve non-whitelist items. 
The `reflect.DeepEqual()` list now contains the check if `[net_one] ==  [net_one, net_two]`, which will always return false.  

**Expected Behavior**: Both clients are restricted to the defined whitelistOnly domains.

**Behavior**: None of them have whitelist enabled.

## Possible solution 
- if groupsToCheck contains ONE whitelistOnly group, then the client can only lookup whitelisted domains (this is implemented in the PR)
- if groupsToCheck groups are ALL whitelistOnly group, then the client can only lookup whitelisted domains

I changed the data type from slice to dict for performance reason. 

Best regards,
c-f

### DIFF
--- a/resolver/blocking_resolver.go
+++ b/resolver/blocking_resolver.go
@@ -178,13 +178,15 @@ func (r *BlockingResolver) BlockingStatus() api.BlockingStatus {
 // returns groups, which have only whitelist entries
 func determineWhitelistOnlyGroups(cfg *config.BlockingConfig) (result map[string]bool) {
 	result = make(map[string]bool)
+
 	for g, links := range cfg.WhiteLists {
 		if len(links) > 0 {
 			if _, found := cfg.BlackLists[g]; !found {
-				result[g] = true 
+				result[g] = true
 			}
 		}
 	}
+
 	return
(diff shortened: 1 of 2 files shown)


0xERR0R:
fixes #202

0xERR0R:
@c-f Thank you for your work. I tested your PR and it works like a charm :+1:

PR #219, opened 2021-06-19
Update configuration.md

zewelor:
I believe this is wrong and should be written caching here.


0xERR0R:
Of course, you are right!

PR #224, opened 2021-06-25
Update installation doc with change of depreacted externalResolvers

kshcherban:
Before blocky complained about:

```
[2021-06-25 10:10:20]  WARN parallel_best_resolver: using deprecated 'externalResolvers' as default upstream resolver configuration name, please consider to change it to 'default'
```


0xERR0R:
Thank you!

PR #243, opened 2021-08-10
Docs change with udp and tcp deprecation

FileGo:
If `udp:x.y.z.w` or `tcp:x.y.z.w` is used as a resolver, a warning is recorded upon startup:

```
[2021-08-10 22:32:22]  WARN net prefix udp is deprecated, using tcp+udp as default fallback
```

Standalone use of `tcp` or `udp` alone with resolvers seemed to have been deprecated, so I have changed documentation to reflect this.


0xERR0R:
Thank you!

PR #259, opened 2021-09-04
Allow configuration of block time

grilix:
When an address is blocked, it can be cached by the client. If we
then wish to allow that address, or just disable the blocking
feature, that client is not gonna be able to see that domain until
the previous domain expires.

The units of time for this setting is in seconds, since we might
want to set it to values around 5 or 10 seconds, depending on the
scenario. The default value for it is the value used before, so
ignoring this setting wont result on any change.

Wasn't sure about the names to use, so make sure you comment
if you have some better examples or comment regarding that or
anything else!

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -258,6 +258,7 @@ type BlockingConfig struct {
 	WhiteLists        map[string][]string `yaml:"whiteLists"`
 	ClientGroupsBlock map[string][]string `yaml:"clientGroupsBlock"`
 	BlockType         string              `yaml:"blockType"`
+	BlockTimeSec      int                 `yaml:"blockTTL"`
 	RefreshPeriod     int                 `yaml:"refreshPeriod"`
 }
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -268,6 +268,19 @@ default** block type. Server returns 0.0.0.0 (or :: for IPv6) as result for A an
       blockType: nxDomain
     ```
 
+### Block TTL
+
+TTL for answers to blocked domains can be set to customize the time clients ask for those domains again.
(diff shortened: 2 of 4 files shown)


0xERR0R:
Hey,

thank you for your work! It looks good for me, just one question: currently all duration configuration parameters in config.yaml are in minutes (refreshPeriod, minTime, maxTime, prefetchExpires), maybe it would be better to change the introduced parameter to minutes too (to be consistent with other parameters)? What do you think?

grilix:
Hey, thanks for the input!

Yes, I noticed that, and wasn't really sure. I didn't like using different units that much either, but I genuinely think this value shouldn't be more than half a minute.

The story is that when we disable blocking, we'd expect it to reflect "immediately" (and not having to wait a minute or more.) I normally have to enable/disable blocking almost every day or every other day, so having lower TTL would be a great help for me.

To be able to set it to a minute would still be an improvement, but wouldn't be the best solution IMHO. As a reference I currently have it set to 5 seconds pi-hole seems to be using 2 seconds.

Maybe we could approach this differently? The other way I could think of, would be having a boolean that's just `shortTTL: true` or something like that, which sets the TTL to 10 seconds, for example?

Any idea would be much appreciated, or if you want to just use minutes that's fine too, we can do that and see how it goes from there. Let me know what you think.

0xERR0R:
OK, I understand you point. I think, it would be better to change all duration parameters to Go's duration struct and allow user to enter something like "2s", "2m30s" or "4h50m" (https://pkg.go.dev/time#ParseDuration). This would be more concise and more human-readable for big duration (e.g. 8h instead of 28800 in seconds). I'll create a follow up issue for this release.

grilix:
Oh, that sounds great, thanks! I'll update this one to be in that format later today :+1:

0xERR0R:
I'll merge this one now -> the configuration was introduced and it looks good. I'll create a follow-up issue for duration format

grilix:
Sounds good, thanks!

PR #288, opened 2021-09-21
chore(docker): Update alpine version

onedr0p:
* Updates Alpine version
* Install tzdata and ca-certificates in the final image

I'm not certain why installing tzdata or ca-certificates was excluded and replaced by copying from the build container. Any thoughts @0xERR0R ?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,6 @@ RUN apk add --no-cache \
     make \
     gcc \
     libc-dev \
-    tzdata \
     zip \
     ca-certificates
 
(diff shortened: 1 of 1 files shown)


0xERR0R:
Hi, thank you for your work!

The first version of the Docker file was based on "scratch" base image. This is the reason, why tzdata and certificates were copied from the builder image. Later I introduced a health-check and changed the image to alpine (in order to be able to install dig). I don't see any reasons, why we could'n also install tzdata etc.

Issue #310, opened 2021-10-04
application startup should fail if initial download of a single list failed


0xERR0R:
Startup should fail if at least one list download fails after all retry attempts

PR #313, opened 2021-10-10
application startup should fail if initial download of a single list failed (#310)

kwitsch:
Changes:

- server.go:NewServer -> returns multpiple errors instead of one
- Resolver interface -> new function GetInitErrors returns slice of errors which occured during resolver initialisation
- list_cache.go:NewListCache -> additionaly returns slice of errors which occured during resolver initialisation
- list_cache.go -> modified error handling


0xERR0R:
Let me know if you need help or if I can take a task or a part

Issue #333, opened 2021-11-08
Option to disable the query log


0xERR0R:
It should be possible to completely disable the query log.

PR #334, opened 2021-11-09
Implemented the option to disable query logging

FileGo:
This implements the option to disable query logging, as requested in #333. A new `NoneWriter` is implemented, which requires almost no change in code to implement the new feature.

The default behaviour is unchanged; if there is no `queryLog` configured, the output is to console. To disable query logging, the following configuration is required:

```
queryLog:
  type: none
```

Documentation has been modified to reflect the changes.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -29,7 +29,8 @@ import (
 type NetProtocol uint16
 
 // QueryLogType type of the query log ENUM(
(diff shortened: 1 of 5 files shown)


0xERR0R:
Thanks for your work! :handshake:

0xERR0R:
closes #333

Issue #251, opened 2021-08-23
Define different client names for the same group


0xERR0R:
It should be possible to define more than one client name (IP, wildcard etc) for the group. Example:

```yaml
  clientGroupsBlock:
    # multiple client names are separated by ','
    laptop,desktop*:
      - ads
    192.168.178.1/24:
      - special
```

PR #337, opened 2021-11-12
Define different client names for the same group (#251)

kwitsch:
different client names for the same group


0xERR0R:
Thank you for your work! :handshake:

0xERR0R:
closes #251

PR #349, opened 2021-11-21
Remove Bias Language from blocky

j0sh3rs:
The goal of this PR is to address the majority of the findings for bias language removal as identified by [Splunk's bias language linter](https://github.com/splunk/biased-lang-linter), primarily for the usage of whitelist/blacklist within this repo.

In cases where the term `master` was identified, but was associated with a branch or an external link, those were left in place so as not to break existing flows/docs.

In all cases where `blacklist` was found, it was replaced with `denylist`. Similarly for `whitelist` to `allowlist`.

Regardless, I'd just like to say thank you for this project and its mission. I'd been looking for a lower-weight alternative for pihole and have been very happy with my usage of blocky so far!

### DIFF
--- a/docs/additional_information.md
+++ b/docs/additional_information.md
@@ -60,7 +60,7 @@ Some links/ideas for lists:
 
 !!! warning
 
-    Use only blacklists from the sources you trust!
+    Use only denylists from the sources you trust!
 
 ### Whitelists
 
@@ -74,7 +74,7 @@ Some links/ideas for lists:
 
     Please read the description before using the DNS server as upstream. Some of them provide already an ad-blocker, some
     filters other content. If you use external DNS server with included ad-blocker, you can't choose which domains should be
-    blocked, and you can't use whitelisting.
+    blocked, and you can't use allowlisting.
 
 This is only a small excerpt of all free available DNS servers and should only be understood as an idee.
 
--- a/metrics/metrics_event_publisher.go
+++ b/metrics/metrics_event_publisher.go
@@ -28,14 +28,14 @@ func registerApplicationEventListeners() {
 }
 
 func versionNumberGauge() *prometheus.GaugeVec {
-	blacklistCnt := prometheus.NewGaugeVec(
+	denylistCnt := prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "blocky_build_info",
 			Help: "Version number and build info",
 		}, []string{"version", "build_time"},
 	)
 
-	return blacklistCnt
+	return denylistCnt
 }
 
 func registerBlockingEventListeners() {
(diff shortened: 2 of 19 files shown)


0xERR0R:
Hi, thank you for you work. I'll review the changes today. We can change the documentation, but I would like not to change any configuration / grafana dashboards etc. to avoid breaking of existing user configuration.

kwitsch:
> We can change the documentation, but I would like not to change any configuration / grafana dashboards etc. to avoid breaking of existing user configuration.

Maybe we could duplicate problematic endpoints and configurations and mark the old ones as deprecated?
Users would get a transition time of some releases to change there setup and the depricated endpoints could be removed later.

0xERR0R:
> > We can change the documentation, but I would like not to change any configuration / grafana dashboards etc. to avoid breaking of existing user configuration.
> 
> Maybe we could duplicate problematic endpoints and configurations and mark the old ones as deprecated? Users would get a transition time of some releases to change there setup and the deprecated endpoints could be removed later.

Sure, this would be the way for configuration. For Grafana dashboard, maybe temporary produce metrics with both keys (old and new) and later remove the old one.

metrics/metrics_event_publisher.go
```diff
@@ -51,23 +51,23 @@ func registerBlockingEventListeners() {
...
-	RegisterMetric(blacklistCnt)
-	RegisterMetric(whitelistCnt)
+	RegisterMetric(denylistCnt)
+	RegisterMetric(allowlistCnt)
 	RegisterMetric(lastListGroupRefresh)
 
 	subscribe(evt.BlockingCacheGroupChanged, func(listType lists.ListCacheType, groupName string, cnt int) {
 		lastListGroupRefresh.Set(float64(time.Now().Unix()))
 		switch listType {
 		case lists.ListCacheTypeBlacklist:
```
0xERR0R, line 67: It should be "ListCacheTypeDenylist"

metrics/metrics_event_publisher.go
```diff
@@ -51,23 +51,23 @@ func registerBlockingEventListeners() {
...
-			blacklistCnt.WithLabelValues(groupName).Set(float64(cnt))
+			denylistCnt.WithLabelValues(groupName).Set(float64(cnt))
 		case lists.ListCacheTypeWhitelist:
```
0xERR0R, line 69: It should be "ListCacheTypeAllowlist"

metrics/metrics_event_publisher.go
```diff
@@ -82,26 +82,26 @@ func enabledGauge() prometheus.Gauge {
...
 }
 
-func whitelistGauge() *prometheus.GaugeVec {
-	whitelistCnt := prometheus.NewGaugeVec(
+func allowlistGauge() *prometheus.GaugeVec {
+	allowlistCnt := prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Name: "blocky_whitelist_cache",
-			Help: "Number of entries in the whitelist cache",
+			Name: "blocky_allowlist_cache",
```
0xERR0R, line 99: renaming the metric will break existing grafana dashboards. Better here: create new metric with new name and mark "old" as deprecated. Both metrics should count the same value

server/server_endpoints.go
```diff
@@ -237,7 +237,7 @@ func configureRootHandler(cfg *config.Config, router *chi.Mux) {
 		pd.Links = []HandlerLink{
 
 			{
-				URL:   "https://htmlpreview.github.io/?https://github.com/0xERR0R/blocky/blob/master/docs/swagger.html",
```
0xERR0R, line 240: This link does not exist. This should show to master

docs/additional_information.md
```diff
@@ -60,7 +60,7 @@ Some links/ideas for lists:
 
 !!! warning
 
-    Use only blacklists from the sources you trust!
+    Use only denylists from the sources you trust!
 
 ### Whitelists
```
0xERR0R, line 65: Here and line 51 too?

0xERR0R (changes requested):
It looks almost good, but there are some compiler errors. Please see also my comments.

j0sh3rs:
Hi @0xERR0R -- thanks for your comments, back to work today so I'll review them and make sure CI passes.

Overall, I agree with the approach -- it would be beneficial to duplicate the metrics with some planned deprecation date so as to phase out the language. Having docs updated to use the non-biased language will get new users setup appropriately. Only updating docs and never dropping support for the original language doesn't ensure the overall issue gets addressed

PR #267, opened 2021-09-06
Change golang build constraint syntax

grilix:
It took me a bit to make the `golangci-lint` tool work on my end, but it seems like using the new format fixes the issues.

PR #361, opened 2021-12-11
Make go:build compatible with go 1.16

ThinkChaos:
#267 broke building on older go versions. This fixes it.
I also made sure it still builds with 1.17.


ThinkChaos:
Looks like an unrelated failure. Can you retry?

0xERR0R:
Thanks for your work. I don't understand, why there is a problem with golangci-lint. I'll merge it

Issue #344, opened 2021-11-15
FR: DNS cache sync between multiple blocky instances


kwitsch:
If blocky is deployed on multiple instances for concurrency and/or failsafe the cache most likely will differ.
This will cause spikes in response time during instance switches.

I'd like to propose an external second level cache for blocky.
Redis would be logical solution as it's already been used in similar scenarios(unbound cache db).

If activated this feature would include:
- populate blocky cache from redis during startup
- query redis after cache miss
- update redis entry on cache insertion/update

0xERR0R:
Hey,

which cache do you mean: cache with black/whitelists or cache with DNS responses (positive/negative)?

kwitsch:
Hi,
This would be the DNS response cache.
It was inspired by the section "Cache DB Module Options" in the [unbound manual](https://www.nlnetlabs.nl/documentation/unbound/unbound.conf/).

0xERR0R:
I think, this would be a nice feature, but is should also run without redis. I don't like current cache implementation (which I forked and patched:  https://github.com/0xERR0R/go-cache), maybe it is possible to run something redis compatible in memory for single instance and optionally with external redis for multiple instances to implement only one API.

kwitsch:
My idea was to keep the current cache and add the new redis cache as seperate(optional) resolver between cache and parallel_best.
As it depends on external services it will certainly be slower than an in-memory cache but potentionally way faster than an internet request.
Removing the internal blocky cache would decrease the resolution performance a lot in my opinion.
Therfore i wouldn't suggest that.

An alternative solution may be to let blocky broadcast cache insertions to other instances.
If such a broadcast is recieved the same cache insertion is done on the recieving end.
I guess this would be even better since ther wouldn't be a seperate resolver, server and request neccesarry.
Drawback on this solution would be potentionally a little more network traffic.

0xERR0R:
I think, at runtime only one cache should be there, either external redis or internal. My idea was to use something like "embedded redis", maybe a kind of in-memory cache, which is compatible to redis API (I'm not sure something exists, but I hope so). In this case we could implement caching against redis API and user can either configure external redis or to use "internal" one.

kwitsch:
This would have the contrary effect to my proposal. 😅

Speed in my home environment where DNS resolution is done locally behind blocky:
- after clean start ~35ms(no cache)
- blocky miss, unbound hit ~15ms(unbound cache)
- blocky hit ~7ms(blocky cache)

[...]

Issue #290, opened 2021-09-25
Caching doesn't seem to work properly


agneevX:
I have caching/prefetching set up pretty generously, however it seems that some queries are slow to look up when it should already be cached. 

![A624CA4B-024E-4E27-823D-9216D620C325](https://user-images.githubusercontent.com/19761269/134778484-269190f3-b2cf-455f-9bca-24bf07f91f48.jpeg)

```yml
logLevel: warn

upstream:
  default:
    - https://8.8.8.8/dns-query
    - https://8.8.4.4/dns-query

caching:
  minTime: 5m
  maxTime: 30m
  prefetching: true
  prefetchExpires: 24h
  prefetchThreshold: 2
```

0xERR0R:
Do you have any log entries or do you use csv or database logger? You can only observe a longer time (80ms), but it doesn't mean that this query wasn't cached.

agneevX:
I don't use any type of logging due to this being run on a Raspberry Pi w/ SD card.

agneevX:
Does that high response times are possible even though it's cached? Shouldn't it be sub 1ms?

0xERR0R:
I think, it depends on your network and server load. I have 1 - 80ms for cached entries. I use only udp dns. If you use doh, there is more overhead for encryption

kwitsch:
I tested the logging issue with a test stack.
[test-stack](https://github.com/0xERR0R/blocky/files/7230290/test-stack.txt) [blocky-config](https://github.com/0xERR0R/blocky/files/7230288/blocky-config.txt)
The min TTL is set to one minute and the same url is requested every two minutes.

It appears that after the prefetch initialisation every third request isen't delivered from cache.
[docker log snippet](https://github.com/0xERR0R/blocky/files/7230289/docker-log.txt)

agneevX:
@kwitsch thanks for the investigation! That’s a neat app you have there 👍

kwitsch:
@0xERR0R thanks, the app is used in my current DNS stack to keep often but infrequently used domains cached.

0xERR0R:
I tried to reproduce this behavior. There is a small gap between the log entry expiration and prefetch loading. Let me explain, how it works:
[...]

PR #365, opened 2021-12-11
FR: DNS cache sync between multiple blocky instances (#344)

kwitsch:
Pull request for ticket #344 

~~Sync tested with testdata/docker-compose.yml~~

Currently missing:
- ~~config documentation~~
- unittests
- ~~TTL sync~~
- ~~go routine reduction~~
- ~~stored data size reduction~~
- ~~sync message size reduction~~
- ~~buffer messaging channels~~
- ~~cleanup testdata~~


0xERR0R:
Hey, thanks for you work! I'll take a look this week on this PR. Are you using this already on your "productive" environment? If yes, how does it perform?

kwitsch:
Sorry didn't get enough time for a proper benchmarking.
I'm currently trying to figure out how to unit test the redis client(evaluating minredis).

I'll report back when I got benchmark results.

kwitsch:
Did a quick benchmark today:

Empty cache in test vs trained productive(dual blocky)
```
Test            |  Min  |  Avg  |  Max  |Std.Dev|Reliab%|
----------------+-------+-------+-------+-------+-------+
+ Cached Name   | 0,001 | 0,002 | 0,002 | 0,000 | 100,0 |
+ Uncached Name | 0,032 | 0,059 | 0,197 | 0,043 | 100,0 |
+ DotCom Lookup | 0,031 | 0,037 | 0,049 | 0,004 | 100,0 |
---<-------->---+-------+-------+-------+-------+-------+

Productive      |  Min  |  Avg  |  Max  |Std.Dev|Reliab%|
----------------+-------+-------+-------+-------+-------+
+ Cached Name   | 0,003 | 0,004 | 0,005 | 0,000 | 100,0 |
+ Uncached Name | 0,032 | 0,091 | 0,287 | 0,079 | 100,0 |
+ DotCom Lookup | 0,035 | 0,043 | 0,054 | 0,005 | 100,0 |
---<-------->---+-------+-------+-------+-------+-------+
```

The results are pretty equal.
Test environment was virtualized locally.
Productive(blocky 0.17) is a multi node swarm in the local network.

After training restarterted one test instance new: fully trained cache in ~1-2s

kwitsch:
@0xERR0R sorry I'm a little stuck unit testing the caching resolver.
Committed my attempt so far.
Seems like i'm missing something.

0xERR0R:
> @0xERR0R sorry I'm a little stuck unit testing the caching resolver. Committed my attempt so far. Seems like i'm missing something.

ok, thanks. I started yesterday with review and testing, but I'll need some time. We can discuss and plan open tasks later. Regarding the tests, if it is too complicated with redis, maybe we can just mock it.

kwitsch:
> > @0xERR0R sorry I'm a little stuck unit testing the caching resolver. Committed my attempt so far. Seems like i'm missing something.
> 
> ok, thanks. I started yesterday with review and testing, but I'll need some time. We can discuss and plan open tasks later. Regarding the tests, if it is too complicated with redis, maybe we can just mock it.

Thanks for don't getting annoyed by the constantly crashing workflows 😅
I stop committing for now until you got time reviewing it.

Got the unit test running and did minor changes as result of a 24h test.

resolver/caching_resolver.go
```diff
@@ -66,6 +76,19 @@ func configurePrefetching(c *CachingResolver, cfg *config.CachingConfig) {
...
+			if rc != nil {
+				logger.Debug("Key: ", rc.Key, " Value: ", rc.Response)
+				c.putInCache(rc.Key, rc.Response, false)
+			}
+		}
+	}()
+}
+
 // check if domain was queried > threshold in the time window
 func (r *CachingResolver) isPrefetchingDomain(cacheKey string) bool {
```
0xERR0R, line 93: I'm thinking about prefetching: if we do not publish cache reloads from prefetching -> caches on different blocky instances will be different. But on the other side: each blocky instance will prefetch different domains.
What do you think?

0xERR0R:
I did some tests, it works well! I commented some code fragments, maybe not everything is necessary, we can discuss it later.

I think we have an issue with TTL, here is my expectations:

Case 1:

- fresh blocky instance A and B, fresh redis
- I query A for "google.com" and receive response with IP and TTL of 300 sec. Blocky makes upstream request.
- 10 sec later, I query B for "google.com" and receive response from cache, TTL 290
- 10 sec later, I restart B and query B again for "google.com". Now I receive again the response from cache, but TTL is 300.

I would expect, TTL should be 280. 

Case 2:

- fresh blocky instance A and B, fresh redis
- I query A for "google.com" and receive response with IP and TTL of 300 sec. Blocky makes upstream request.
- I wait 10 min and restart blocky
- Blocky loads staled cache result from redis
- I query A for "google.com" and receive response with TTL 300 from cache

I would expect, after 300 sec the cache entry is expired.

The problem is, TTL is "coded" in response and this response is in the cache. I think we should use TTL in redis. I understand it correct, the entry will expire in redis. If we load the entry from redis, we should also check the actual TTL (it is possible to check TTL, but with separate command (https://redis.io/commands/ttl), I'm not sure if it is possible with "scan".

What do you think?

model/models.go
```diff
@@ -41,3 +42,57 @@ type Request struct {
 	Log             *logrus.Entry
 	RequestTS       time.Time
 }
+
+// ResponseCache struct holding key and response for cache synchronization
+type ResponseCache struct {
```
0xERR0R, line 47: This structure is used as redis payload. I didn't use redis before, but I can imagine, that JSON has some overhead here: more network traffic and more memory consumption on redis side.
It the key attribute necessary here, it is kinda redundant to redis key? Would it make sense to use some binary format for encoding here to reduce the memory footprint?

redis/redis.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// PublishCache publish cache to redis async
+func (c *Client) PublishCache(key string, response *model.Response) {
+	msg, errConv := response.ConvertToCache(key)
+	if errConv == nil {
+		binMsg, errMar := msg.MarshalBinary()
+		if errMar == nil {
+			c.client.Publish(c.ctx, CacheChannelName, binMsg)
+			c.client.Set(c.ctx, prefixKey(key), binMsg, time.Duration(0))
```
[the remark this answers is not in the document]
kwitsch: Thought about it too but would solve it a little different:
- create a configuration option to let the user configure redis.cacheTTL
- set the dns result ttl to 0 upon redis insertion
- set the prefetch flag(if enabled config wise) upon cache insertion following a sync message

Otherwise we would create the same issue as in #290 in redis.

Maybe you got a different idea to combat this issue?

resolver/caching_resolver.go
```diff
@@ -66,6 +76,19 @@ func configurePrefetching(c *CachingResolver, cfg *config.CachingConfig) {
 	})
 }
 
+func setupRedisSubscribers(c *CachingResolver) {
+	logger := logger("caching_resolver_redis")
+
+	go func() {
+		for rc := range c.redisClient.Channel {
+			if rc != nil {
+				logger.Debug("Key: ", rc.Key, " Value: ", rc.Response)
```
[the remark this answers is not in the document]
kwitsch: Agreed this was purely for initial debugging purpose.

redis/redis.go
```diff
@@ -0,0 +1,192 @@
...
+		Addr:     cfg.Address,
+		Password: cfg.Password,
+		DB:       cfg.Database,
+	})
+	ctx := context.Background()
+
+	var err error
+
+	attempt := 1
+	for attempt <= cfg.ConnectionAttempts {
```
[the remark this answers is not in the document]
kwitsch: I'll look into it.

model/models.go
kwitsch: We could drastically reduce the payload if we only store Res in the prefixed key and clean it upon retrieval.
I think this would be sufficient information wise for an blocky instance to generate its cache entry.

My first idea was to also cache negative responses but later it occured to me that this would be a huge waste of resources.

I agree that currently that there is a huge overhead:
image

redis/redis.go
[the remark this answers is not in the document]
0xERR0R: hm, I don't understand your approach. Each DNS response has a TTL time. This TTL should be propagated via redis to other blocky instance otherwise there would be a cache entry in blocky with wrong TTL. We can try to use TTL in redis -> this would also cleanup expired cache entries in redis. Or we put "expirationTimestamp" into the cache struct, so the other blocky instance can calculate TTL by something like "expirationTimestamp - now())

model/models.go
0xERR0R: Reason and RType are used only for logging purposes (also query log). Maybe we can remove them from caching and log in this case "reason: EXTERNAL_CACHE" or something similar

kwitsch:
Thanks for the input.

I looked into it and build a prototype that includes most suggestions. It performed well with a slight performance increase.

Will cleanup the code and commit&push it tomorrow.

redis/redis.go
```diff
@@ -0,0 +1,264 @@
...
+	Response *model.Response
+}
+
+// Client for redis communication
+type Client struct {
+	config       *config.RedisConfig
+	client       *redis.Client
+	l            *logrus.Entry
+	ctx          context.Context
+	id           []byte
```
0xERR0R, line 53: 👍 : thats nice, that own messages can be ignored

0xERR0R:
I tested it again and it works well! Great job, thank you! Do you want to add/change something or is it ready to merge?

kwitsch:
> I tested it again and it works well! Great job, thank you! Do you want to add/change something or is it ready to merge?

Thanks 😊

No changes intended at the moment.
Ready to be merged according to me.

PR #370, opened 2021-12-14
Allow configuring custom DNS TTL

ThinkChaos:
I changed the tests to use a random but common TTL on each run. Maybe you'd prefer a separate test for the TTL explicitly.
Also, I haven't updated the docs.


0xERR0R:
Thanks for your work! It looks good. Do you want to add a description to `configuration.md` and `docs/config.yml`? If not, it is ok, I can do it.

ThinkChaos:
I added the docs!

PR #372, opened 2021-12-17
 Allow configuring multiple listen addresses for each port

ThinkChaos:
I use this to bind to localhost and my LAN interface, but not the WAN one.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -35,8 +35,8 @@ Complete documentation is available at https://github.com/0xERR0R/blocky`,
 	}
 
 	c.PersistentFlags().StringVarP(&configPath, "config", "c", "./config.yml", "path to config file")
-	c.PersistentFlags().StringVar(&apiHost, "apiHost", "localhost", "host of blocky (API)")
-	c.PersistentFlags().Uint16Var(&apiPort, "apiPort", 4000, "port of blocky (API)")
+	c.PersistentFlags().StringVar(&apiHost, "apiHost", "localhost", "host of blocky (API). Default overridden by config and CLI.") // nolint:lll
+	c.PersistentFlags().Uint16Var(&apiPort, "apiPort", 4000, "port of blocky (API). Default overridden by config and CLI.")
 
 	c.AddCommand(newRefreshCommand(),
 		NewQueryCommand(),
@@ -64,8 +64,12 @@ func initConfig() {
 	if len(config.GetConfig().HTTPPorts) != 0 {
 		split := strings.Split(config.GetConfig().HTTPPorts[0], ":")
 
+		lastIdx := len(split) - 1
+
+		apiHost = strings.Join(split[:lastIdx], ":")
+
 		var p uint64
-		p, err := strconv.ParseUint(strings.TrimSpace(split[len(split)-1]), 10, 16)
+		p, err := strconv.ParseUint(strings.TrimSpace(split[lastIdx]), 10, 16)
 
 		if err != nil {
 			util.FatalOnError("can't convert port to number (1 - 65535)", err)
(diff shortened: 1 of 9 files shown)


docs/configuration.md
```diff
@@ -11,20 +11,20 @@ configuration properties as [JSON](config.yml).
...
-| port            | [IP]:port         | no                    | 53                 | Port and optional bind ip address to serve DNS endpoint (TCP and UDP). If you wish to specify a specific IP, you can do so such as 192.168.0.1:53. Example: 53, :53, 127.0.0.1:53  |
-| tlsPort         | [IP]:port         | no                    |                    | Port and optional bind ip address to serve DoT DNS endpoint (DNS-over-TLS). If you wish to specify a specific IP, you can do so such as 192.168.0.1:853. Example: 83, :853, 127.0.0.1:853  |
-| httpPort        | int (1 - 65535)   | no                    |                    | HTTP listener port and optional bind ip address . If > 0, will be used for prometheus metrics, pprof, REST API, DoH ...If you wish to specify a specific IP, you can do so such as 192.168.0.1:4000. Example: 4000, :4000, 127.0.0.1:4000 |
-| httpsPort       | int (1 - 65535)   | no                    |                    | HTTPS listener port and optional bind ip address . If > 0, will be used for prometheus metrics, pprof, REST API, DoH... If you wish to specify a specific IP, you can do so such as 192.168.0.1:443 |
```
ThinkChaos, line 19: I removed the part about `> 0` because as far as I can tell, 0 isn't treated specially. Maybe that's a bug, in which case I could add the docs back and fix the code.

server/server.go
```diff
@@ -52,30 +51,34 @@ func NewServer(cfg *config.Config) (server *Server, err error) {
...
-		if httpsListener, err = net.Listen("tcp", getServerAddress(cfg.HTTPSPort)); err != nil {
-			return nil, fmt.Errorf("start https listener on port %s failed: %w", cfg.HTTPSPort, err)
-		}
+	httpsListeners, err := newListeners("https", cfg.HTTPSPorts)
+	if err != nil {
+		return nil, err
+	}
 
+	if len(httpListeners) != 0 || len(httpsListeners) != 0 {
 		metrics.Start(router, cfg.Prometheus)
```
ThinkChaos, line 82: I haven't tested metrics as I don't have the setup for it yet, so please be mindful of this part.

cmd/root.go
```diff
@@ -61,11 +61,15 @@ func initConfig() {
...
 	log.ConfigureLogger(config.GetConfig().LogLevel, config.GetConfig().LogFormat, config.GetConfig().LogTimestamp)
 
-	if config.GetConfig().HTTPPort != "" {
-		split := strings.Split(config.GetConfig().HTTPPort, ":")
+	if len(config.GetConfig().HTTPPorts) != 0 {
+		split := strings.Split(config.GetConfig().HTTPPorts[0], ":")
+
+		lastIdx := len(split) - 1
+
+		apiHost = strings.Join(split[:lastIdx], ":")
```
ThinkChaos, line 69: This isn't completely related, but I added this behavior as I noticed the discrepancy while fixing the code to work with a list.
I can remove commit 79dcf57e6689fbbdbe85ef45a8c19abc2988c69d if you prefer.

resolver/client_names_resolver.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// ClientNamesResolver tries to determine client name by asking responsible DNS server vie rDNS (reverse lookup)
```
0xERR0R, line 18: :+1:

server/server.go
0xERR0R: I tested it -> works

0xERR0R:
Thanks for you work! It looks good for me, works as expected! :+1:

PR #385, opened 2021-12-28
fix: update dashboard for new grafana versions

coolguy1771:
This update would update all of the data inputs and sources to be compatible with newer versions of Grafana. 

- Updated Pie charts to use the new Pie Charts plugin
- Updated Time Series to use the new Time Series plugin
- Updated stat panels to use Table instead of Time Series for input

### DIFF
--- a/docs/blocky-grafana.json
+++ b/docs/blocky-grafana.json
@@ -1,65 +1,4 @@
 {
-  "__inputs": [
-    {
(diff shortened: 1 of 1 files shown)


0xERR0R:
Thanks, it looks good. Your JSON was not prepared for importing (it was not exported for "sharing externally"), but I exported it again, now it can be imported and grafana asks for variable and datasource value

Issue #362, opened 2021-12-11
[Feature] read /etc/hosts 


dnalor:
dnsmasq (and others) can automatically read /etc/hosts and additional hostfile(s)/a directory of hostfiles (dnsmasq's `addn-hosts` parameter).
While customDNS>mapping can be used to archive something similar, the hosts method is very handy for having a single file (format) for static LAN IPs to distribute.
I don't have that many static IPs, so it's no big deal to update several files, but it might still be useful.

kwitsch:
If the hosts are configured in blocky why do they need a identical host file on their own?
Are they using different dns server?

dnalor:
> If the hosts are configured in blocky why do they need a identical host file on their own? 

It's a matter of convenience - static IP->host mappings are already in /etc/hosts for all servers (so they still work if DNS is unavailable), so I basically have to copy/paste any changes/additions there to blocky's config.
I used to run dnsmasq before, which reads /etc/hosts by default, so I didn't need to configure static host->IP mappings for DNS resolution.

PR #393, opened 2022-01-02
Enable resolving hosts file

FileGo:
This implements functionality desired in #362. It creates a new resolver, which looks up entries in hosts file and resolves them, if found, otherwise query is passed along to next resolver in queue.

In order to avoid any OS compatibility issues, a `hostsFile` configuration entry is required to point out to the hosts file location. If left empty, hosts resolving is disabled.

PS: Sorry I made a bit of a mess with multiple commits, I'm still learning Git :)


resolver/hosts_file_resolver.go
```diff
@@ -0,0 +1,182 @@
...
+	if question.Qtype == dns.TypePTR {
+		response := new(dns.Msg)
+		response.SetReply(request.Req)
+
+		for _, host := range r.hosts {
+			raddr, _ := dns.ReverseAddr(host.IP.String())
+
+			if raddr == question.Name {
+				ptr := new(dns.PTR)
+				ptr.Ptr = dns.Fqdn(host.Hostname)
```
0xERR0R, line 35: I'm not sure: maybe also aliases should be considered here. Example:

`10.0.0.1                router0 router1 router2`

Should reverse DNS for the IP return only router0 as A record or 3 A records (rourter0, router1 and router2)?

resolver/hosts_file_resolver.go
```diff
@@ -0,0 +1,182 @@
...
+	"os"
+	"strings"
+
+	"github.com/0xERR0R/blocky/model"
+	"github.com/0xERR0R/blocky/util"
+	"github.com/miekg/dns"
+	"github.com/sirupsen/logrus"
+)
+
+const hostsDNSTTL = 3600
```
0xERR0R, line 15: Please externalize this parameter like "CustomTTL" with defined default value. See https://github.com/0xERR0R/blocky/blob/9259c5cdcdccbda7fd243d33aa564f0153d8c0ef/config/config.go#L313

resolver/hosts_file_resolver_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	. "github.com/0xERR0R/blocky/helpertest"
+	. "github.com/0xERR0R/blocky/model"
+	"github.com/miekg/dns"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/stretchr/testify/mock"
+)
+
+var _ = Describe("HostsFileResolver", func() {
```
0xERR0R, line 15: :+1: 100% coverage

0xERR0R (changes requested):
Thank you for your work! I reviewed your changes and tested a bit, it looks really good! I fixed some lint issues and added some comments/suggestions as review. Feel free to comment.

0xERR0R:
> 
> PS: Sorry I made a bit of a mess with multiple commits, I'm still learning Git :)

It is ok, I will squash all commits to one on merging this PR

resolver/hosts_file_resolver.go
FileGo: Done

resolver/hosts_file_resolver.go
FileGo: Ah, I've thought you should only get one PTR record, but DNS does allow for more. That's been fixed.

resolver/hosts_file_resolver.go
```diff
@@ -0,0 +1,178 @@
...
+
+func (r *HostsFileResolver) Resolve(request *model.Request) (*model.Response, error) {
+	logger := withPrefix(request.Log, "hosts_file_resolver")
+
+	if r.HostsFilePath == "" {
+		logger.WithField("resolver", Name(r.next)).Trace("go to next resolver")
+		return r.next.Resolve(request)
+	}
+
+	if !r.parsed {
```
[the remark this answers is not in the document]
FileGo: Makes sense, I was thinking of the same, but didn't know how to implement it. I've look how blocky does it elsewhere and found some good examples.

Refresh period is also user-defined now.

0xERR0R:
Thanks! It looks good, I added one test case and improved logging a bit. We can merge it :+1:

PR #402, opened 2022-01-10
Do not run docker build on fork

### LINKED ISSUE #401: Development docker build fails on fork
I've got this repo forked and the development docker build GitHub action always fails on forked repo with every push, as it can't login to Docker Hub.

Perhaps it could be worth adding something to the likes of condition below, to prevent Docker build from running on forks?

```
if: github.repository == ‘0xERR0R/blocky’
```

FileGo:
Fixes #401.

### DIFF
--- a/.github/workflows/development-docker.yml
+++ b/.github/workflows/development-docker.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   docker:
+    if: github.repository_owner == '0xERR0R'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout


0xERR0R:
Do we have the same problem with "Upload results to codecov" step?

FileGo:
> Do we have the same problem with "Upload results to codecov" step?

It works fine, but not sure if that's the desired behaviour? Is there any benefit in codecov upload running on forks?

0xERR0R:
I was just wondering, why this PR has changed the code coverage. I think, this step can run on forks, but codecov should not consider any results from fork build. Not sure, how it currently works 🤷

FileGo:
Yeah, not sure either really, it's only one line in a GH yaml 😐

Issue #377, opened 2021-12-22
Sync blocking enable/disable over redis


0xERR0R:
Blocking enable/disable event should be synced over redis (pub/sub).

PR #403, opened 2022-01-10
Sync blocking enable/disable over redis (#377)

kwitsch:
~~still needs some unittests~~


kwitsch:
@0xERR0R sorry it seem i still don't really get unit test. 😞
According to redis_test.go line 105-161 the lines codecov marks as untested should be covered.

0xERR0R:
Thank you for your work! I'll try to review and test this PR this week. I see, you are using the same pub/sub channel as for the cache synchronization (by using coded message type as message field). What was the reason to reuse this channel and not to create a separate one only for enable/disable state (2 separate channels for 2 different messages). I guess you wanted to reuse the logic with client uuid?

kwitsch:
> Thank you for your work! I'll try to review and test this PR this week. I see, you are using the same pub/sub channel as for the cache synchronization (by using coded message type as message field). What was the reason to reuse this channel and not to create a separate one only for enable/disable state (2 separate channels for 2 different messages). I guess you wanted to reuse the logic with client uuid?

Yes i figured utilizing different channels would duplicate functionality on the client side.
Thought that ~3 byte in communication would cost less time than expensive logic on the client side.

Issue #319, opened 2021-10-20
queryLog: log instance hostname or identifier to distinguish log entries in multi-instance installation


0xERR0R:
If more than 1 blocky is running, it is not possible to distinguish query log entries. It would be nice to log at least the host name as additional column.

PR #410, opened 2022-01-16
queryLog: log instance hostname or identifier to distinguish log entries in multi-instance installation (#319)

kwitsch:
Implemented a per start instance uuid wich is used in redis and logging.
Logging can be activated with the config value "logInstanceId: true".


0xERR0R:
The idea was to have an instance id in the database query log. If multiple instances are running, it would be useful to distinguish between them. In prometheus metrics, there is always a job id but not in the database table. So basically, there must be something like "instance name" string for a blocky instance (maybe default just a host name, but configurable). This name should also not change if blocky gets restarted.

kwitsch:
> The idea was to have an instance id in the database query log. If multiple instances are running, it would be useful to distinguish between them. In prometheus metrics, there is always a job id but not in the database table. So basically, there must be something like "instance name" string for a blocky instance (maybe default just a host name, but configurable). This name should also not change if blocky gets restarted.

Ah ok sorry i got this feature request wrong.
I thought of it to distinguish different log entries in an combined stack log.

As the hostname of all instances in an swarm is the same and one instance could possibly be on different hosts the identification of instances would be troublesome without an external storage.

Got an solution in my mind utilizing redis for this, but it would only work if there is an active connection.

I try coming up with a more independent solution.

0xERR0R:
> > The idea was to have an instance id in the database query log. If multiple instances are running, it would be useful to distinguish between them. In prometheus metrics, there is always a job id but not in the database table. So basically, there must be something like "instance name" string for a blocky instance (maybe default just a host name, but configurable). This name should also not change if blocky gets restarted.
> 
> Ah ok sorry i got this feature request wrong. I thought of it to distinguish different log entries in an combined stack log.
> 
> As the hostname of all instances in an swarm is the same and one instance could possibly be on different hosts the identification of instances would be troublesome without an external storage.
> 
> Got an solution in my mind utilizing redis for this, but it would only work if there is an active connection.
> 
> I try coming up with a more independent solution.

👍 I think, we can use the "instance name" also for logging. This should be done automatically, I think we don't need a flag to activate it. Hostname should be the default value (or something generated, if you think it is better).

PR #420, opened 2022-01-22
development-docker workflow rework

kwitsch:
- enabled image creation on fork repository
- dynamic tag and repo selection for main repository(needs to be tested on @0xERR0R repository)
~~- docker layer caching~~
- trivy scan & result upload


0xERR0R:
if I understand correctly, there a 3 major changes:

- trivy scan -> 👍 
- caching for docker 
  - -> not sure if we need it. Key contains the commit sha, does it mean this cache will be used only if I start the job again for the same commit? Personally, I would try to avoid any caches for docker build. Each time it runs "apk install" it should get the latest package version. What do you think?
- docker build on forks
  - works only with ghcr.io -> this is ok, but the special handling for docker hub is ugly. Maybe we should split the development-docker.yml in 2 different jobs: one for github and one for ghcr.io?
  - This means also, everybody who forks the repository and commits some change (maybe just fixing a typo or playing around) will automatically deploy the docker image in his repo under ghcr.io -> Don't you think, the deployment should be more explicit? Maybe it would be better to introduce new repo variable "docker_push" or something similar, which is false by default? If someone forks this repo and wants to deploy the docker image -> in this case the variable must be set.

.github/workflows/development-docker.yml
```diff
@@ -6,39 +6,104 @@ on:
...
+        run: |
+          RESPONSE="${{steps.get_gh.outputs.tag}}"
+          if [ "${{github.repository_owner}}" == '0xERR0R' ]
+          then
+            RESPONSE+="\nspx01/blocky:${{steps.extract_branch.outputs.branch}}"
+          fi
+          echo "##[set-output name=tag;]$(echo ${RESPONSE})"
+        id: get_all
+
+  docker-gitlab:
```
0xERR0R, line 36: gitlab?

kwitsch: sorry should be github, had to many projects open 😐

kwitsch:
> if I understand correctly, there a 3 major changes:
> 
> * trivy scan -> 👍
> * caching for docker
>   
>   * -> not sure if we need it. Key contains the commit sha, does it mean this cache will be used only if I start the job again for the same commit? Personally, I would try to avoid any caches for docker build. Each time it runs "apk install" it should get the latest package version. What do you think?
> * docker build on forks
>   
>   * works only with ghcr.io -> this is ok, but the special handling for docker hub is ugly. Maybe we should split the development-docker.yml in 2 different jobs: one for github and one for ghcr.io?
>   * This means also, everybody who forks the repository and commits some change (maybe just fixing a typo or playing around) will automatically deploy the docker image in his repo under ghcr.io -> Don't you think, the deployment should be more explicit? Maybe it would be better to introduce new repo variable "docker_push" or something similar, which is false by default? If someone forks this repo and wants to deploy the docker image -> in this case the variable must be set.

The cache is restored from "${{ runner.os }}-buildx-" so any previous runs should act as it's source.
It was an attempt to reduce the build time.
The caching itself works quite well but doesn't make a big difference since the go build cannot be cached and it takes the most time.
I agree that it can be removed.

I would like to have images on my fork since manually building and pushing them is a bit troublesome.
Introducing a variable to enable this would be great. 😁

Also agreeing that the solution for mutiple tags is a bit messy. 
Couldn't come up with a better solution. 😕
Opted against multiple workflows since the buildtime would possibly differ. 
This would create different images on GitHub and DockerHub.

Maybe pass the build time as arg from job to Dockerfile and build both images on different jobs after each other 🤔

kwitsch:
Will close this pull request.
My use case isn't really common and would burden others.

Issue #418, opened 2022-01-22
Example configuration uses not available blacklists


ikarisan:
Hi!
In the example configuration blocklist URLs are used, that are not available.
At least

https://zeustracker.abuse.ch/blocklist.php?download=domainblocklist
https://mirror1.malwaredomains.com/files/justdomains

ist not working. Maybe the example config should be cleaned up.

PR #424, opened 2022-01-28
Example configuration uses not available blacklists #418

kwitsch:
removed dead links from configuration example #418


0xERR0R:
Thanks!

PR #447, opened 2022-02-28
Update to gingko v2

ThinkChaos:
I did this to be able to use the gingko CLI tool.


go.mod
```diff
@@ -3,32 +3,38 @@ module github.com/0xERR0R/blocky
 go 1.17
 
 require (
-	github.com/alicebob/miniredis/v2 v2.18.0
+	github.com/alicebob/miniredis/v2 v2.17.0
```
0xERR0R, line 6: downgrade?

go.mod
```diff
@@ -3,32 +3,38 @@ module github.com/0xERR0R/blocky
...
-	github.com/hashicorp/golang-lru v0.5.4
-	github.com/miekg/dns v1.1.46
+	github.com/mattn/go-colorable v0.1.12 // indirect
+	github.com/mgutz/ansi v0.0.0-20200706080929-d51e80ef957d // indirect
+	github.com/miekg/dns v1.1.43
 	github.com/mroth/weightedrand v0.4.1
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.18.1
-	github.com/prometheus/client_golang v1.12.1
+	github.com/prometheus/client_golang v1.11.0
```
0xERR0R, line 20: downgrade?

ThinkChaos:
Ah good catch, I glossed over that file.  
I'll double check later but I did do a lot of branches and rebasing cause I started this to be able to write tests for a bigger change I made. I'll get back to you ASAP.

Yeah dependabot doesn't seem to detect this. I've never set it up myself so no clue if there's an option for it.

ThinkChaos:
I manually reverted the downgrades and ran `go mod tidy`. Should be good now!

PR #449, opened 2022-03-04
Add rewrite support to custom DNS

ThinkChaos:
Main commit message:
> This commit extracts rewriting logic from `ConditionalUpstreamResolver`
> into the new `RewriterResolver`, and uses that to enable rewriting for
> the `CustomDNSResolver`.
> `RewriterResolver` wraps a resolver and applies the rewrite to the
> request that is forwarded to the inner resolver.
> 
> It also introduces a new optional interface: `NamedResolver`.
> This allows a `Resolver` to choose what its user friendly name is,
> instead of always being its type name.

I use this feature to have an alternative way of spelling my domain which is a bit long.
Comes in handy when manually poking around.

Also did some smaller changes to tests such as unifying `MockResolver` and `resolverMock`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Blocky is a DNS proxy and ad-blocker for the local network written in Go with fo
     * Supports modern DNS extensions: DNSSEC, eDNS, ...
     * Free configurable blocking lists - no hidden filtering etc.
     * Provides DoH Endpoint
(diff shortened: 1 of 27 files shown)


ThinkChaos:
Force pushed to fix `Configuration()` output.
The failed test looks random and unrelated: I haven't seen this in any previous CI or local run.

PR #457, opened 2022-03-16
docs/config.yml: default for tlsPort is 853

johanneskastl:
default for `tlsPort` is 853, hence adjusting the (commented out) example


0xERR0R:
Makes absolutely sense, thank you!

PR #472, opened 2022-03-30
[Enhancement] Add kafka querylog functionality

rmellmer:
For my use-case, I wanted to be able to log DNS queries to a Kafka stream for further analysis. Figured others might find this feature useful as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,11 +5,13 @@ RUN apk add --no-cache \
     make \
     gcc \
     libc-dev \
+    librdkafka-dev \
+    musl-dev \
     zip \
     ca-certificates
 
 ENV GO111MODULE=on \
-    CGO_ENABLED=0
+    CGO_ENABLED=1
     
 WORKDIR /src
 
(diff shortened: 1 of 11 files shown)


0xERR0R:
Hi, thanks for you work! I think it can be useful if someone has the kafka stack running. Just curious: what is your toolset to analyse queries from blocky?

The build fails currently, I assume this is the issue with "musl" flag and different linux distros? Did you try to use another go library for Kafka (for example kafka-go), which are not relying on libkafka/musl?

rmellmer:
@0xERR0R Yeah, I was hoping that the official confluent kafka library would work, but it seems that as you mentioned, that the reliance on librdkafka is going to be an issue with the build. My use case was to run it in alpine linux, which the musl build flag is appropriate for, but seems that it can't easily be cross-compatible without specifying different build flags depending on the platform.

I haven't tried using a different Kafka library yet, but I'll take a stab at that here and will update the branch if I can get it working with the kafka-go library.

victoraugustolls:
Hi! New here and just found this issue! Sorry for the intrusion, just a minor comment suggesting to take a look at [franz-go](https://github.com/twmb/franz-go), so far the best experience I had with a Golang kafka driver! Cheers!

PR #475, opened 2022-03-31
Exclude resolvers without config from the chain

ThinkChaos:
The code in this branch makes it so only actually configured resolvers are created and added to the chain.
This should simplify reading trace logs, and add a small speedup to queries (didn't attempt to benchmark).

The PR is not ready: tests do not pass yet, and that's a fair bit of work.
Is it worth it I spend time fixing the tests, is this something you'd be interested in merging?

### DIFF
--- a/resolver/blocking_resolver.go
+++ b/resolver/blocking_resolver.go
@@ -90,6 +90,10 @@ type BlockingResolver struct {
 
 // NewBlockingResolver returns a new configured instance of the resolver
 func NewBlockingResolver(cfg config.BlockingConfig, redis *redis.Client) (ChainedResolver, error) {
+	if len(cfg.ClientGroupsBlock) == 0 {
+		return nil, nil
+	}
+
 	blockHandler := createBlockHandler(cfg)
 	refreshPeriod := time.Duration(cfg.RefreshPeriod)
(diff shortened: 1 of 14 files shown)


0xERR0R:
IMHO, I don't see any advantages for this:
- each resolver implementation has more code now (check if passed configuration must be processed or is the configuration for each resolver "does nothing"), but the result is the same (returning nil and exclude the resolver from chain vs. resolver delegates to next without any operation)
- Each resolver has a constructor function "NewXXX", which implies that the function returns a new instance. Returning nil is IMHO against this principle
- regarding performance improvements: I think, it is not really measurable, if resolver does "nothing" it simple delegates to the next one
- Regarding debugging: It depends, what do you mean with debugging. If you analyze log files -> OK, there may be more log entries, because not active resolvers log something. If you mean code debugging with debugger -> it depends on your tool. For example in Goland (aka. IntelliJ) you can create conditional breakpoints.

ThinkChaos:
I understand your concerns. Closing this is fine with me.
(Debugging I did mean end user via logs)

I think the `ChainBuilder` would be useful for #476. I can open a PR with just that if you agree.

Issue #471, opened 2022-03-30
Weird behavior for www.eitf.org


ThinkChaos:
Hi,

I've noticed a weird behavior when querying `www.eitf.org`: blocky returns the `RRSIG`s for the `CNAME` and the domain it points to, even though I'd expect it to handle those internally and just forward the `CNAME` and `A` records.
To get the results I ran the following queries (some output elided for brevity):
```bash
$ dig www.ietf.org @127.0.0.1
[...]
;; QUESTION SECTION:
;; www.ietf.org.       		IN	A

;; ANSWER SECTION:
www.ietf.org.       	1194	IN	CNAME	www.ietf.org.cdn.cloudflare.net.
www.ietf.org.       	1194	IN	RRSIG	CNAME 5 3 1800 20230328000424 20220327230707 40452 ietf.org. MNYE8[...]==
www.ietf.org.cdn.cloudflare.net. 1194	IN	A	104.16.44.99
www.ietf.org.cdn.cloudflare.net. 1194	IN	A	104.16.45.99
www.ietf.org.cdn.cloudflare.net. 1194	IN	RRSIG	A 13 6 300 20220331024317 20220329004317 34505 cloudflare.net. teE5[...]==
[...]
```

This is causing an issue for me because with systemd's resolved pointing to blocky the domain ends up not resolving!
```bash
$ resolvectl query www.ietf.org
www.ietf.org: resolve call failed: Received invalid reply
```

<details>
<summary>Extra logs (that don't add too much information)</summary>

```
$ sudo resolvectl log-level debug
$ journalctl -efu systemd-resolved.service # run the query in another shell
[...]
Cache miss for www.ietf.org IN AAAA
Firing regular transaction 30781 for <www.ietf.org IN AAAA> scope dns on */* (validate=yes).
Using feature level TLS+EDNS0+D0 for transaction 30781.
Using DNS server 127.0.0.1#dns.[redacted] for transaction 30781.
Sending query via TCP since UDP isn't supported or DNS-over-TLS is selected.
Using feature level TLS+EDNS0+D0 for transaction 30781.
Announcing packet size 512 in egress EDNS(0) packet.
Cache miss for www.ietf.org IN A
Firing regular transaction 26063 for <www.ietf.org IN A> scope dns on */* (validate=yes).
Using feature level TLS+EDNS0+D0 for transaction 26063.
Using DNS server 127.0.0.1#dns.[redacted] for transaction 26063.
[...]

Issue #464, opened 2022-03-21
Support of DoH/DoT as bootstrap DNS


0xERR0R:
Currently, bootstrap DNS can be plain DNS only (UDP/TCP via 53). It should be possible to use DoT/DoH as bootstrap DNS. If host system has no valid DNS configuration, bootstrap should be defined as IP address only.

PeterDaveHello:
As non-IP DoT/DoH bootstrap DNS needs "another bootstrap DNS", which is the DNS used by system in this case, will limit to use IP address in bootstrap DNS be easier for the implementation?

0xERR0R:
Yes, there is only one bootstrap DNS in blocky. The system will use the default DNS resolver if needed. Ideally, bootstrap DNS should be only IP address or empty if host is configured properly.

0xERR0R:
I would prefer to use the same syntax here as for upstream resolvers (for example tcp+udp:example.com). Otherwise we have 2 different syntax for the same thing.

0xERR0R:
If host has no working DNS resolver, blocky will need one if bootstrap DNS is a host address. Workaround would be to use some ip addresses (like 1.1.1.1) which do have also a valid certificate (with SAN for the IP address).

If I understand it right, your approach would be to define an IP address for the resolver and to specify the awaited server name in the certificate. So technically, blocky will establish the connection to the DNS server (DoT or DoH) with the IP, but it will bypass the standard certificate chain validation and check if server certificate contains the defined "server name" in SAN or CN field?

ThinkChaos:
Yeah exactly.
A simpler way to see it is "hardcoded DNS response" à la /etc/hosts IMO, and that's what I was envisioning for the implementation.

0xERR0R:
ok, got it. In this case it would be probably better to create this "special" case only for bootstrap DNS.

ThinkChaos:
Also if you don't like the idea of a string format, we could do something like:
```yaml
bootstrapDns:
  address: x.x.x.x
  hostname: xxxx.com
```

I think it's doable with a new type that implements unmarshal (haven't looked at the code name probably wrong) by trying from a string first for backwards compatibility and then this format.

What's your thoughts on how this should look in the config?

0xERR0R:
> Also if you don't like the idea of a string format, we could do something like:
> 
> ```yaml
> bootstrapDns:
>   address: x.x.x.x
>   hostname: xxxx.com
> ```
> 
[...]

PR #477, opened 2022-04-01
Add cd bit to cache key to fix DNS poisoning

### LINKED ISSUE #455: DNS cache poison of DNSSEC?
By default, if the upstream resolver is doing DNSSEC validation, the query of the record with broken signature will return [SERVFAIL](https://dnssec.vs.uni-due.de/dig-sigfail), but using `dig` with the `+cd` flag(_to set the CD (checking disabled) bit in the query_) to query domain will ask the resolver to bypass the DNSSEC check, return the result without validation.

Just noticed that the result of a CD bit enabled query will be cached by blocky, and then served to the client no matter if they enabled the CD bit, once the cache is created, the following query without CD bit will retrieve the same result without DNSSEC validation as the CD bit enabled.

The test result using [Quad9](https://quad9.net/):

```
$ dig +short sigfail.verteiltesysteme.net @9.9.9.9
$ dig +short sigfail.verteiltesysteme.net @9.9.9.9
$ dig +cd +short sigfail.verteiltesysteme.net @9.9.9.9
134.91.78.139
$ dig +short sigfail.verteiltesysteme.net @9.9.9.9
```

The test result using Block to proxy queries to Quad9 via DoH:

```
$ dig +short sigfail.verteiltesysteme.net @::1
$ dig +short sigfail.verteiltesysteme.net @::1
$ dig +cd +short sigfail.verteiltesysteme.net @::1
134.91.78.139
$ dig +short sigfail.verteiltesysteme.net @::1
134.91.78.139
```

Not sure if this is something should be fixed? Thanks.

ThinkChaos:
The CheckingDisabled flag tells the resolver to skip DNSSEC signature
verification.

Fixes #455

### DIFF
--- a/resolver/caching_resolver.go
+++ b/resolver/caching_resolver.go
@@ -95,7 +95,7 @@ func (r *CachingResolver) isPrefetchingDomain(cacheKey string) bool {
 }
 
 func (r *CachingResolver) onExpired(cacheKey string) (val interface{}, ttl time.Duration) {
-	qType, domainName := util.ExtractCacheKey(cacheKey)
+	_, qType, domainName := util.ExtractCacheKey(cacheKey)
 
 	logger := logger("caching_resolver")
 
@@ -160,7 +160,7 @@ func (r *CachingResolver) Resolve(request *model.Request) (response *model.Respo
 
 	for _, question := range request.Req.Question {
 		domain := util.ExtractDomain(question)
-		cacheKey := util.GenerateCacheKey(question.Qtype, domain)
+		cacheKey := util.GenerateCacheKey(&request.Req.MsgHdr, question.Qtype, domain)
 		logger := logger.WithField("domain", util.Obfuscate(domain))
 
 		r.trackQueryDomainNameCount(domain, cacheKey, logger)
--- a/resolver/caching_resolver_test.go
+++ b/resolver/caching_resolver_test.go
@@ -526,7 +526,7 @@ var _ = Describe("CachingResolver", func() {
 			It("load", func() {
 				request := newRequest("example2.com.", dns.TypeA)
 				domain := util.ExtractDomain(request.Req.Question[0])
-				cacheKey := util.GenerateCacheKey(dns.TypeA, domain)
+				cacheKey := util.GenerateCacheKey(&dns.MsgHdr{}, dns.TypeA, domain)
 				redisMockMsg := &redis.CacheMessage{
 					Key: cacheKey,
 					Response: &Response{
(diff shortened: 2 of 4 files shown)


ThinkChaos:
Also relates to #471.

I think before merging we should determine if there are other fields missing from the key.
One thing that comes to mind but I haven't checked is that we might cache truncated responses.

0xERR0R:
Hey, thanks for your PR! I skimmed over the code: the usage of the CD bit in the cache key will potentially increase the number of cached entries if some of the clients use CD and some not (maybe this is only a theoretical problem). My idea for the solution (not tired yet, maybe this can't work this way) was:
The cache key remain. We have the complete DNS response in the cache and we can determine if the response is valid (EDNS/OPT section). On new request, we check if CD flag is set, if yes, we return the cached item (valid or invalid). If not, we check if cached item is valid and return it only if it is.
As I mentioned, I didn't check this approach, but if it works, we can keep the cache size. The only change is, we "mark" invalid cache entries as "bad data" and return this data only if explicitly desired.

What do you think about it?

ThinkChaos:
I did this change as a quick way to fix the DNS poisoning issue, but indeed there's more work to be done.
As you said, this causes duplication for responses that are DNSSEC valid since the CD flag doesn't impact the result in that case we should use a single cache entry for both. I think your approach sounds good for that.

I also think we first need to go over the RFCs and determine what other changes we need to make, and probably other stub resolver code, to have a complete picture of what special care is needed before making bigger changes.

For instance, the DO flag tells resolvers they should include `RRSIG`s, and if space is an issue, set TC (truncate) in the response. Blocky doesn't take into account if DO is set when using the cache. That's why I was seeing `RRSIG`s in responses from blocky even if I wasn't using DNSSEC (testing with dig).
In the case of DO, it's possible to use the same entry whether the flag is set or not, as long as you strip the `RRSIG`s when it is not.
RFC4035 has [a section](https://www.ietf.org/rfc/rfc4035.html#section-4.9) on how stub resolvers should interact with DNSSEC related flags.

Later I'll start by working on #464  and then will do some more reading related to this.
Do you want to open a new issue, or maybe expand the scope of #455 to include the other changes we need to make to caching, including determining exactly what they are?

0xERR0R:
Yes, you're right, blocky doesn't care actually about request flags and simple caches responses. This will work for the most cases, but it should be fixed.
Regarding the scope expansion or creating of new issue - please feel free to do as it fits best.

PeterDaveHello:
Not sure if considering the AD (authentic data) bit in the implementation will help?

PeterDaveHello:
Should keep it opened.

PR #480, opened 2022-04-02
Update reference configuration file in the docs

PeterDaveHello:
To prevent the warning:

> net prefix tcp is deprecated, using tcp+udp as default fallback


0xERR0R:
Thanks!

PR #498, opened 2022-04-16
Change swagger URL to match blocky's version

ThinkChaos:
Just a small change so the link to Swagger is versioned.


0xERR0R:
Thanks, this makes sense!

PR #518, opened 2022-05-13
Hide DoH client http header "User-Agent"

### LINKED ISSUE #446: Hide http user agent for DNS over HTTPS like what Firefox and iOS do?
Just found that currently blocky is using the default HTTP user agent string: `Go-http-client/1.1`, though we know that there're tons of user profiling mechanisms to find out the tools and details about the users, but maybe it'll be great if we can give one less information, like user agent string? That's what Firefox and iOS doing, both known as privacy protected products, is this something blocky would like to do 😃

PeterDaveHello:
I'm newbie to Go, please help verify the patch, thanks!

close #446

### DIFF
--- a/resolver/upstream_resolver.go
+++ b/resolver/upstream_resolver.go
@@ -117,7 +117,15 @@ func (r *httpUpstreamClient) callExternal(msg *dns.Msg,
 		return nil, 0, fmt.Errorf("can't pack message: %w", err)
 	}
 
-	httpResponse, err := r.client.Post(upstreamURL, dnsContentType, bytes.NewReader(rawDNSMessage))
+	req, err := http.NewRequest("POST", upstreamURL, bytes.NewReader(rawDNSMessage))
+
+	if err != nil {
+		return nil, 0, fmt.Errorf("can't create the new request %w", err)
+	}
+
+	req.Header.Set("User-Agent", "")
+	req.Header.Set("Content-Type", dnsContentType)
+	httpResponse, err := r.client.Do(req)
 
 	if err != nil {
 		return nil, 0, fmt.Errorf("can't perform https request: %w", err)


0xERR0R:
Thanks for your work! Your code change looks good, we can merge it. One question: your implementation sets empty string as user agent. The proposed solution (see https://github.com/0xERR0R/blocky/issues/446#issuecomment-1054665140) on #446 was to make the user-agent string configurable via config.yaml. It is absolutely ok for me to take your approach, I just want to avoid some misunderstandings. Do you think, it makes not really sense to make user-agent configurable?

0xERR0R:
closes #446

PeterDaveHello:
@0xERR0R set it to empty looks good enough to me, configurable seems to be another level nice to have feature. Just I'm new to Golang, so make it configurable will be much more complex to me to finish it at this moment. Maybe we can do it later in the future. Or anyone who'd like to make the patch, well also be great.

FileGo:
@PeterDaveHello Should be a fairly easy fix, but if you are too busy, let me know and I can do the configuration bit.

0xERR0R:
> @0xERR0R set it to empty looks good enough to me, configurable seems to be another level nice to have feature. Just I'm new to Golang, so make it configurable will be much more complex to me to finish it at this moment. Maybe we can do it later in the future. Or anyone who'd like to make the patch, well also be great.

ok, got it, thanks!

0xERR0R:
> @PeterDaveHello Should be a fairly easy fix, but if you are too busy, let me know and I can do the configuration bit.

@FileGo I'll merge the fix from @PeterDaveHello, if you are interested in making the user-agent header configurable -> your contributions are always welcome 🤝

FileGo:
Yeah, might do it later today.

PR #520, opened 2022-05-13
Improve DoT server TLS cipher suites

PeterDaveHello:
This removes some VULNERABLE, or potentially VULNERABLE ciphers, like Triple DES and Obsoleted CBC ciphers, for the DoT server.

For the cipher suggestions, we can refer to Mozilla SSL Configuration Generator:
- https://ssl-config.mozilla.org/#server=go

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -190,6 +190,14 @@ func createTLSServer(address string, certFile string, keyFile string) (*dns.Serv
 		TLSConfig: &tls.Config{
 			Certificates: []tls.Certificate{cer},
 			MinVersion:   tls.VersionTLS12,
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
Not sure why the cipher change would break the test 🤔

0xERR0R:
> Not sure why the cipher change would break the test thinking

There are some tests which test cache expiration, sometimes is github environment too slow. I can't reproduce it locally, I want to rewrite those tests in the near future

0xERR0R:
Thanks for your work! I did not know the site from mozilla, very cool!

Issue #343, opened 2021-11-15
FR: Generate self signed certificate if cert key and file are not specified


samip5:
I would like to address the fact that if the container requires the certificate's key and the certificate file itself, it's rather troublesome to deploy in Kubernetes.

Could it be done in a way where Blocky generates an self signed certificate if those are not specified?

0xERR0R:
Hey,

I think, there are some pros and cons regarding this question:

- On one side: yes, this would be cool, if blocky would generate and renew certificates automatically. But the question is, how should it work? Only with Let's encrypt or also with other services? Only DNS challenge or also HTTP? I created https://github.com/0xERR0R/dns-proxy as proof-of-concept project and I use it as public DoT resolver with blocky.

- On the other side: I think, the certificate management is out-of-scope (KISS principle) and can be done outside. In k8s, you can simple use Ingress with automatic certificate creation. Or it is always possible to use e.g. traefik as reverse proxy (traefik can TLS). There are some disadvantages, like passing the client name is not possible if DoT is proxied.

I'm not against the certificate creation, but I want to reduce the maintenance effort later. In https://github.com/0xERR0R/dns-proxy I used https://github.com/go-acme/lego library to create certificates. I can imagine, we can integrate this library in blocky and generate only Lets Encrypt certificates, with only DNS challenge and only with DNS providers which are supported by lego. Would it fit your needs?

samip5:
No, I don't mean that Blocky would automatically renew any other certificate than it's self signed one, if that would be done in Blocky so that the ingress controller could actually connect to it without having to mount the k8s secret certficate inside it.
I don't want that Blocky were to get Lego as an dependency too as I don't think that would be needed.

In a Kubernetes deployment, Blocky will most of the time be either poxied via an ingress controller, aka nginx/traefik or setup as a LoadBalancer, in which case traffic hits Blocky directly without being proxied.

How/why does it matter to DoT if it's proxied or not, in the aspect of getting the client name for example?

0xERR0R:
Sorry, I don't get your point: do you want, that blocky creates a self signed certificate or a valid let's encrypt certificate?

[...]

PR #532, opened 2022-05-23
Self-signed certificate generation

FileGo:
This PR adds generation of a self-signed certficate, as proposed in #343.

It changes the functionality quite a lot - server start will no longer fail if HTTPS or TLS port is defined, but certificate/key files aren't. Blocky will instead create its own CA, which will in turn sign the newly created certificate.

As using self-signed certificates is insecure by the nature of it (compared to using a trusted CA), I didn't see the point of adding an option for configuring the DNS name for the certificate, as any software using blocky's DoT/DoH features in this configuration would have to use something similar to `InsecureSkipVerify` anyway.

I'd appreciate any feedback on this.


0xERR0R:
Thanks for your work!

> As using self-signed certificates is insecure by the nature of it (compared to using a trusted CA), I didn't see the point of adding an option for configuring the DNS name for the certificate, as any software using blocky's DoT/DoH features in this configuration would have to use something similar to `InsecureSkipVerify` anyway.
> 

Absolutely agree, this should be used only for testing purposes or in trusted local networks. I also think, it is ok, that the certificate will be regenerated on each start -> it doesn't make sense to import this generated cert in any trust store. The only pain point is the start time, on my laptop it needs about 8 sec to generate the certificate, I think on a raspberry pi 1/2 it would take a noticeably amount of time, but in this case it is always possible to generate the certificate externally.

FileGo:
> The only pain point is the start time, on my laptop it needs about 8 sec to generate the certificate, I think on a raspberry pi 1/2 it would take a noticeably amount of time, but in this case it is always possible to generate the certificate externally.

I have noticed this during development, although my dev VM is fortunately slightly faster (about 4-5 seconds for start).

I have thought about using ECDSA keys, and blocky with slightly modified code launches nearly instantaneously on Pi Zero W 2, but I decided against it in favour of greater compatibility that use of standard RSA keys provide. I think this feature has a quite niche use, and is likely not going to used by many, and certainly not in production.

If you think the use of ECDSA doesn't pose any significant compatibility issues, let me know, and I can do a PR.

Issue #512, opened 2022-05-11
Support for multiple configuration files


Qarasique:
Can i use "include" directive for config file, like in nginx ?

Example:
Separate *.conf file for

```
conditional:
  mapping:
```
and
```
customDNS:
  mapping:
```

I want to use a lot of static A entries, it would be nice to divide service settings and DNS entries

ThinkChaos:
For static A entries, you could use the hosts file resolver to achieve this ([docs](https://0xerr0r.github.io/blocky/configuration/#hosts-file)).
There is no generic config include support though.

0xERR0R:
So basically the idea is to provide a path to the directory with several YAML files, blocky should read all files (alphabetically order) and "merge" them together, right?

PR #534, opened 2022-05-24
Support for multiple configuration files

FileGo:
This adds support for multiple configuration files (in a single directory), as proposed in #512.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Blocky is a DNS proxy and ad-blocker for the local network written in Go with fo
   * Various REST API endpoints
   * CLI tool
 
-- **Simple configuration** - single configuration file in YAML format
+- **Simple configuration** - single or multiple configuration files in YAML format
 
     * Simple to maintain
     * Simple to backup
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -40,7 +40,7 @@ Complete documentation is available at https://github.com/0xERR0R/blocky`,
 		SilenceUsage: true,
 	}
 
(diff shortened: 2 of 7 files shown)


FileGo:
Interesting, test fails on something that hasn't been changed.

@0xERR0R Any ideas?

0xERR0R:
> Interesting, test fails on something that hasn't been changed.
> 
> @0xERR0R Any ideas?

Is a flaky test which I can't reproduce locally ;(

0xERR0R:
Thanks for your work! If I understand your idea right, splitted files should contain distinct part of the target YAML. I mean, basically, all YAMLs in the directory will be concatenated together to a big target file. That means, there is no logical "merge" with attribute overwrite between pieces, right?

So following would (probably?) fail:
config1.yaml
```yaml
upstream:
  default:
    - 5.9.164.112
    - 1.1.1.1
```

config2.yaml
```yaml
upstream:
  someclient:
    - 1.2.3.4
```

I think, it can be a useful use case to separate config files per client or to split the blacklist configuration, but it could be technically difficult to merge the parts correct together (we have nested maps of maps in the config). What do you think?
 
Maybe we can leave it (take your implementation) but improve the documentation, so that it becomes clear that all config parts should have no overlapping. In this case it doesn't matter if the files are read in the alphabetically order or not.

FileGo:
Exactly, current functionality assumes no parts of config will overlap and as you point out, it will fail the YAML unmarshal, if it does overlap.

I'll amend the documentation, to make it more clear.

yunwu-jnpr:
I think it would be better to support `include` directive than blindly merging files under a directory.
You have much more control.

PR #536, opened 2022-05-24
Remove net.Error.Temporary

FileGo:
This fixes #533 to remove functionality, deprecated in Go 1.18.

PR #540, opened 2022-05-25
Update to Go 1.18

FileGo:
Updated things to Go 1.18. Also upgraded the Alpine Linux version in Dockerfile, as 3.16 [has been released recently](https://alpinelinux.org/posts/Alpine-3.16.0-released.html).

Lint action fails, but this has been addressed in #536, so should be fine once it's merged.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -7,10 +7,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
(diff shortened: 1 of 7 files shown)


ThinkChaos:
I think you need to run `go mod tidy` to update the `go.mod` and `go.sum` files, 1.18 now splits direct and indirect dependencies in multiple `require` blocks.

0xERR0R:
Thanks!

PR #545, opened 2022-05-29
Add HTTP Secure Headers on HTTPS server response

PeterDaveHello:
Secure Header reference:
- https://owasp.org/www-project-secure-headers/


0xERR0R:
It looks good, thanks for your work!

PR #546, opened 2022-05-29
 Make DoH/DoT server mininum TLS version configurable

PeterDaveHello:
This allows us to set the minimum TLS version for the DoH & DoT server, without modify the source code and recompile it.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -55,6 +55,20 @@ func logger() *logrus.Entry {
 	return log.PrefixedLog("server")
 }
 
+func minTLSVersion() uint16 {
+	minTLSVer := config.GetConfig().MinTLSServeVer
+	switch minTLSVer {
+	case "1.2":
+		return tls.VersionTLS12
+	case "1.3":
+		return tls.VersionTLS13
+	default:
+		logger().Warn("Not allowed or supported mininum TLS version ", minTLSVer, ", fallback to TLS 1.3")
+
+		return tls.VersionTLS13
+	}
+}
+
 func tlsCipherSuites() []uint16 {
 	tlsCipherSuites := []uint16{
 		tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
@@ -232,7 +246,7 @@ func createTLSServer(address string, cert tls.Certificate) (*dns.Server, error) 
 		Net:  "tcp-tls",
 		TLSConfig: &tls.Config{
 			Certificates: []tls.Certificate{cert},
-			MinVersion:   tls.VersionTLS12,
+			MinVersion:   minTLSVersion(),
 			CipherSuites: tlsCipherSuites(),
 		},
 		Handler: dns.NewServeMux(),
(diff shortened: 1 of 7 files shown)


PeterDaveHello:
Not sure if we can disable this check `G402: TLS MinVersion too low. (gosec)`?

server/server.go
```diff
@@ -54,6 +54,20 @@ func logger() *logrus.Entry {
...
+	minTLSVer := config.GetConfig().MinTLSServeVer
+	switch minTLSVer {
+	case "1.2":
+		return tls.VersionTLS12
+	case "1.3":
+		return tls.VersionTLS13
+	default:
+		logger().Warn("Not allowed or supported mininum TLS version ", minTLSVer, ", fallback to TLS 1.3")
+
+		return tls.VersionTLS13
```
ThinkChaos, line 67: IMO it's a bit weird/surprising that the fallback is not the same as the default specified in the struct field's annotation (1.2).

server/server.go
```diff
@@ -224,9 +238,10 @@ func createTLSServer(address string, cert tls.Certificate) (*dns.Server, error)
 	return &dns.Server{
 		Addr: address,
 		Net:  "tcp-tls",
+		//nolint:gosec
 		TLSConfig: &tls.Config{
 			Certificates: []tls.Certificate{cert},
-			MinVersion:   tls.VersionTLS12,
+			MinVersion:   minTLSVersion(),
```
ThinkChaos, line 244: Does the annotation work if you move to the end of line 244?  
`MinVersion:   minTLSVersion(), // nolint:gosec`

I'm asking cause that could make the linter complain for other issues in the `tls.Config` if any arise in the future.

server/server.go
```diff
@@ -471,8 +486,9 @@ func (s *Server) Start(errCh chan<- error) {
 
 			server := http.Server{
 				Handler: s.httpMux,
+				//nolint:gosec
 				TLSConfig: &tls.Config{
-					MinVersion:   tls.VersionTLS12,
+					MinVersion:   minTLSVersion(),
```
ThinkChaos, line 491: Same as other `nolint`.

server/server.go
PeterDaveHello: because by default you won't need to specify it unless you don't want TLS v1.2

ThinkChaos: Yeah I get your logic, but that only works because there's only two values for now.
Anyways, either way is fine, I don't think this is really an issue.

server/server.go
PeterDaveHello: Not working 😅 
```
server/server.go:247:15: G402: TLS MinVersion too low. (gosec)
                TLSConfig: &tls.Config{
                        Certificates: []tls.Certificate{cert},
                        MinVersion:   minTLSVersion(), // nolint:gosec
                        CipherSuites: tlsCipherSuites(),
                },

server/server.go:249:35: directive `// nolint:gosec` is unused for linter "gosec" (nolintlint)
                        MinVersion:   minTLSVersion(), // nolint:gosec
                                                       ^
```

server/server.go
0xERR0R: Fallback to the highest (1.3) makes sense :+1:

0xERR0R:
Thanks for your work @PeterDaveHello  and also thanks @ThinkChaos for your review? Can it be merged or is there some todos?

PeterDaveHello:
Yeah I believe so!

PeterDaveHello:
But not sure if you guys whould like tp make the default option to TLS 1.3? Those users who are using DoT should have new enough environment I think.

ThinkChaos:
Yeah everything looks good to me!

PR #553, opened 2022-06-03
Rename securityHeader -> secureHeader

PeterDaveHello:
Just noticed that the naming was not so consistent in #545 when I was working on some future works, should correct it as early as possible.


0xERR0R:
:+1:

PR #561, opened 2022-06-10
Option to handle FQDN only requests

kwitsch:
added fqdnOnly to config(default false)
if enabled all non FQDN requests will immediately respond with NXDOMAIN and no further processing will be done


kwitsch:
idk why race test failed 😕
worked a few minutes ago...

0xERR0R:
This race occurs from time to time, but only on github. I opened an issue to track it

resolver/fqdn_only_resolver.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (r *FqdnOnlyResolver) Resolve(request *model.Request) (*model.Response, error) {
+	if r.enabled {
+		domainFromQuestion := util.ExtractDomain(request.Req.Question[0])
+		if !strings.Contains(domainFromQuestion, ".") {
+			response := new(dns.Msg)
+			response.Rcode = dns.RcodeNameError
+
+			reason := fmt.Sprintf("NOTFQDN (%s)", domainFromQuestion)
```
0xERR0R, line 31: Reason is used as a label in prometheus metric (https://github.com/0xERR0R/blocky/blob/e313ef7fe868601b3b3757d41c3677954d0518b2/resolver/metrics_resolver.go#L48), this should be a limited amount of possible values (for example the group name for blocking or the address from upstream resolver). I think it is fine enough just to return "NOTFQDN" here

0xERR0R (changes requested):
Thanks for your work, it looks very good. Please see my comment, the reason should be changed, otherwise we get problems with prometheus due to unlimited set of data

0xERR0R (approved):

Issue #537, opened 2022-05-24
Fix flaky test


0xERR0R:
Sometimes we can observe following error (only with github action):

```bash
• [FAILED] [0.192 seconds]
Downloader
/home/runner/work/blocky/blocky/lists/downloader_test.go:21
  Download of a file
  /home/runner/work/blocky/blocky/lists/downloader_test.go:73
    when DNS resolution of passed URL fails
    /home/runner/work/blocky/blocky/lists/downloader_test.go:196
      [It] Should perform a retry until max retry attempt count is reached and return DNSError
      /home/runner/work/blocky/blocky/lists/downloader_test.go:203

  Expected
      <bool>: false
  to be true
  In [It] at: /home/runner/work/blocky/blocky/lists/downloader_test.go:207
````

kwitsch:
Got way better but I think this issue should stay open as it still gets flaky from time to time

shizunge:
This may not be the test's problem. Instead it may be an indicator to a problem of the downloader used by blocky.

I recently switched to the following setup: (1) a script to download lists via `wget` (2) a static file server running locally to expose these lists to blocky. (3) blocky.

Comparing to the blocky only setup:
Blocky only:
total time spent on downloading and parsing is about 15 minutes.
File downloading randomly failed as well.

With the local static file server:
downloading: about 30 seconds.
blocky parsing: about 5 minutes.

You can see that using `wget` not only is much faster than letting blocky download the files, but also fixes the random failures.

PR #562, opened 2022-06-11
use ginkgo for test executions (#537)

kwitsch:
replaced "go test" with "ginkgo --flake-attempts=5" in "make test" and "make race" which should reduce the workflow fails significantly

this should be considderd as a workaround and not a fix for the flaky test


0xERR0R:
As a workaround it is a good solution. The root cause should be fixed, but you're absolutely right, it should not block other CI jobs. Your approach activates flaky attempts for all tests. Would it not be better to use the FlakeAttempts(N) decorator only for failing test (see https://onsi.github.io/ginkgo/#repeating-spec-runs-and-managing-flaky-specs)?

kwitsch:
> As a workaround it is a good solution. The root cause should be fixed, but you're absolutely right, it should not block other CI jobs. Your approach activates flaky attempts for all tests. Would it not be better to use the FlakeAttempts(N) decorator only for failing test (see https://onsi.github.io/ginkgo/#repeating-spec-runs-and-managing-flaky-specs)?

Ultimately the goal should be to never use it. 😅 
It was just considered to be a quick fix and seems to be the fastest solution. 

I investigated the problem a bit further and got it on different tests. 
In my opinion one of the main problems is the usage of the os and ioutil packages as the read/write behavior seems a bit flaky in the github environment. 

Currently still evaluating possible solutions.

0xERR0R:
I thought about this fix: setting of "flake-attempts" will definitely fix all problems with flaky tests. This is good for new PRs. But I see the danger of something being overlooked. For example, some new feature will be implemented and there is something wrong with feature code or test code and this results in a flaky test. With flake-attempts for all tests, we have to check the log to see if there were test failures -> this is always a time effort.

Setting of "flake-attempts" for race detection is more critical -> if a race detection test is flaky -> there is always a race condition.

I would suggest to remove "flake-attempts" for all tests and to use the "FlakeAttempts(5)" annotation only for hand chosen tests in test code. Those tests can be found more easily later (and probably fixed ;) )

FlakeAttempts will work with "go test" and "ginkgo", but we can keep your change to ginkgo - this will let us to use more features later and ginkgo produces nicer outputs (for example it aggregates all stdout togethere).

What do you think?

kwitsch:
During my tests i got the impression that some tests are executed in a fashion that are logically correct but don't consider the underlying system. 

For example file_writer_test line [181](https://github.com/0xERR0R/blocky/blob/01b7947ff2107ff3d8f28bd90b1bb4ea66d2082f/querylog/file_writer_test.go#L181]) to [188](https://github.com/0xERR0R/blocky/blob/01b7947ff2107ff3d8f28bd90b1bb4ea66d2082f/querylog/file_writer_test.go#L188) is logically correct but doesn't consider that the underlying io operations can take some ms to take effect. 
Which explained why it behavied flaky on my computer. 
Tried to solve it with Eventually as it performs the test with a 50ms delay by default. 
```Go
Eventually(func(g Gomega) int {
	files, err := ioutil.ReadDir(tmpDir)
	g.Expect(err).Should(Succeed())
	return len(files)
}).Should(Equal(2))
writer.CleanUp()

Eventually(func(g Gomega) int {
	files, err := ioutil.ReadDir(tmpDir)
	g.Expect(err).Should(Succeed())
	return len(files)
}).Should(Equal(1))
```
Got a robust test result after this change. 

Will be quite time intensive to find and fix all of these i guess.  😅

0xERR0R:
IMHO  if a test runs not on all environments (for example only on "fast" machines) -> the test is wrong and must be fixed. This situation can also appear on a "productive" hardware. It looks like a github acton test execution is sometimes slow, I think we should take it as a chance to improve the code.

I execute all tests on my i7 machine, but a lot of users run blocky on edge devices (I run blocky on a raspberry pi 3). This will be a challenge to write robust tests, but it is possible ;)

you basically agree?

kwitsch:
Absolutely agree 👍

I run blocky mostly on different PIs and never get any crashes but got flaky tests during development.
As already mentioned it looks more like an issue in some tests than in blocky itself.

I continuously executed the test inside WSL and produce load on the windows host.
Using this i was able to reproduce flaky behavior on some tests.
Mainly those which do io read directly after io write.

I had a similar problem in the past on slow single core machines and thread sleep between write and read fixed the flaky behavior every time.
As the GitHub runner speed can be rather slow at times the problem looks somewhat related to me. 😅

Running the tests in a VM with really low specs should also get them flaky I guess. 🤔
... haven't tried it yet though...

PR #571, opened 2022-06-22
Make UpstreamResolver re-use IPSet between Resolves

ThinkChaos:
Tried implementing this idea again because of https://github.com/0xERR0R/blocky/discussions/557.

Do you think this is worth the extra complexity?


0xERR0R:
Do we need this after #589 is merged?

ThinkChaos:
Not necessarily, but it's not exclusive either.
I'm not sure how much of an improvement this would be as I don't have any way of making stats on requests (yet).

PR #575, opened 2022-06-28
feature: hide client ip/name in querylog

axelrindle:
I'd like to hide the Client IP and Name in the querylog for privacy reasons.

- [x] Extend config
- [x] Fix tests
- [x] Implement for CSV etc. as well
- [x] Move logic to QueryLoggingResolver.writeLog~~Resolve or LogEntry~~

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -517,6 +517,7 @@ type QueryLogConfig struct {
 	LogRetentionDays uint64       `yaml:"logRetentionDays"`
 	CreationAttempts int          `yaml:"creationAttempts" default:"3"`
 	CreationCooldown Duration     `yaml:"creationCooldown" default:"2s"`
+	HideClient       bool         `yaml:"hideClient" default:"false"`
 }
 
 // RedisConfig configuration for the redis connection
--- a/querylog/database_writer.go
+++ b/querylog/database_writer.go
@@ -100,16 +100,16 @@ func (d *DatabaseWriter) Write(entry *LogEntry) {
 	domain := util.ExtractDomain(entry.Request.Req.Question[0])
 	eTLD, _ := publicsuffix.EffectiveTLDPlusOne(domain)
 
-	var clientIp string
-	var clientName string
+	var clientIP, clientName string
+
 	if !d.hideClient {
-		clientIp = entry.Request.ClientIP.String()
+		clientIP = entry.Request.ClientIP.String()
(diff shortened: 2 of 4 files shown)


axelrindle:
Sounds good. I will try to implement this for CSV etc. as well.

Also moving the logic to a more central place is surely a good approach.

config/config.go
```diff
@@ -517,6 +517,7 @@ type QueryLogConfig struct {
 	LogRetentionDays uint64       `yaml:"logRetentionDays"`
 	CreationAttempts int          `yaml:"creationAttempts" default:"3"`
 	CreationCooldown Duration     `yaml:"creationCooldown" default:"2s"`
+	HideClient       bool         `yaml:"hideClient" default:"false"`
```
ThinkChaos, line 520: There's an existing option called `logPrivacy`, maybe we should extend that instead of creating a new option?

And if a new option is created think naming this something that starts with "log" would be nicer to make it clear it impacts the logs, and would follow the existing convention (logLevel, logPrivacy, etc.).

axelrindle: I totally agree to follow the existing naming convention. But what exactly do you mean by extending that option? I am thinking of two possibilities here:

- Use that option instead of `HideClient` in the check in `resolver/query_logging_resolver.go`
- Introduce different "levels" of privacy similar to Pi-hole (e.g. "hide client", "hide targets", "hide all")

0xERR0R: logPrivacy is only for logging purposes. If true, all sensitive data (requested domain, response data) will be obfuscated. Purpose: admin should not see, which domains were resolved.

I think it makes sense to introduce new configuration parameter for query logging. Currently we log client IP, client names, response reason, question (typically domain name), response code, answer and duration.

To make it more configurable, we can introduce a list of "columns" to log, something like:

```yaml
queryLog:
  columns:
    - clientIP
    - clientNames
    - duration
```

column can be an enum. If not specified, we log all possible columns. If only a subset is set, all other columns are empty.

What do you think?

axelrindle: > To make it more configurable, we can introduce a list of "columns" to log, something like:

I think that is awesome.

ThinkChaos: > Introduce different "levels" of privacy similar to Pi-hole (e.g. "hide client", "hide targets", "hide all")

I was thinking of something like this, but what @0xERR0R suggested is great!

0xERR0R:
replaced with #766

PR #589, opened 2022-07-01
Add connectIPVersion to choose how Blocky connects to other hosts

ThinkChaos:
This impacts both DNS lookups and lists downloading.

See https://github.com/0xERR0R/blocky/discussions/557.

I cherry-picked [Fix using wrong context in Bootstrap.resolveUpstream](https://github.com/0xERR0R/blocky/commit/89fe50222633a1829ce25e84a9277d253ddfea42) from #571.
On that topic, that MR can still be useful IMO, but up to you to decide!


resolver/bootstrap.go
```diff
@@ -145,14 +137,16 @@ func (b *Bootstrap) NewHTTPTransport() *http.Transport {
...
 			var qTypes []dns.Type
 
 			switch {
-			case strings.HasSuffix(network, "4") || filteredQTypes.Contains(dns.Type(dns.TypeAAAA)):
+			case upstreamIPVersion != config.IPVersionDual: // ignore `network` if a specific version is configured
+				qTypes = upstreamIPVersion.QTypes()
+			case strings.HasSuffix(network, "4"):
 				qTypes = []dns.Type{dns.Type(dns.TypeA)}
-			case strings.HasSuffix(network, "6") || filteredQTypes.Contains(dns.Type(dns.TypeA)):
+			case strings.HasSuffix(network, "6"):
```
ThinkChaos, line 149: This is a bit of a weird edge case and I'm not sure what's best: the config is less precise so making `network` higher priority makes sense, but `network` is most likely not a end-user choice.  
Should the config override the requested `network` or should it be the other way around?

larivierec:
imo, if I'm the only one having the issue this PR may not be worth.

Although, I'm glad that you find a solution if ever.
I do believe that this would be a good addition

resolver/bootstrap.go
0xERR0R: Thanks for your work!

I think, if user sets "UpstreamIPVersion", this should have precedence.

This property affect as mentioned by you the list download and the resolution of external upstream resolvers. I'm thinking about the naming, maybe we should rename it to something more generic, maybe "NetworkIPVersion" or "NetworkMode"?

Can you please also describe the new configuration property in docs/configuration.md and docs/config.yml?

0xERR0R: @ThinkChaos any thoughts regarding my previous comment?

ThinkChaos:
Rebased on `development`

resolver/bootstrap.go
ThinkChaos: Sorry had completely missed this!  
Second force-push updates the MR with new name and docs.

For the name I went with `connectIPVersion` because I think it makes it clear it doesn't prevent from listening on an IP version you disable for outgoing connections.  
Let me know if you prefer something else.

larivierec:
Is there a chance this will be merged?
😊

ThinkChaos:
It was delayed by my own fault, I missed @0xERR0R's comment in the review (see thread at the top).

0xERR0R (approved):
LGTM :+1:

Issue #476, opened 2022-03-31
Let's make Blocky more flexible


ThinkChaos:
While working on #449 and #475, I touched the resolver chaining and adding resolver nesting. This has got me thinking it would be nice to be able to configure arbitrary resolver chains, or rather trees, in the config.
This means making `createQueryResolver` in server.go configuration driven.

I haven't done any code work on this, but have been thinking about it for a while and think I have a decent vision for an implementation.

Here's an example of what I mean by configuration driven,
Please don't pay too much attention to the syntax and names, it's nothing concrete, just a draft to explain what I'm thinking.
```yaml
resolvers:
- log:
- metrics:
- rewrite:
    rules: [...]
    resolvers:
    - hostsFile:
    - cache:
    - customDNS:
      mapping: [...]
    - conditional:
        home: <router ip>
    - parallel:
      upstreams: [...]
```

This would make blocky much more flexible, and, I think, easy to understand without looking at `createQueryResolver`.
My main confusion source while using blocky has been the order in which resolvers run. Having the user decide would help alleviate that IMO.

What do you think of this idea?

P.S. please do not take this as me not liking blocky. On the contrary, I really enjoy using it, and want to help build it into something even greater!

EDIT: sorry accidentally pressed Ctrl-Enter. Filled out the rest!

0xERR0R:
I have thought about it several times in the past. Making the resolver order configurable can help "power users" with solving of some special problems. But on the other side, they need deep technical understanding about blocky's internal structure and it can lead to unpredictable results. For example: if you move loggingResolver or metricsResolver, you'll get wrong metrics result or incomplete query logs.

I don't know, what is the right way. I tried always to keep things simple (KISS principle), but I see, there are different use cases in the wild and each user has unique network infrastructure.

Perhaps the point has also come where we can think about a completely new concept, maybe some kind of plugins or microservice approach, where the user can decide, which parts of the system he wants to run or to customize.

I am completely open to constructive suggestions or further discussion.

ThinkChaos:
Yeah I understand your concern. I think the defaults should stay unchanged, and that this should be opt-in to respect KISS without limiting power users.
With that in mind I think generally the people that setup their own infra like this are technical enough to understand that the logging must come early in the chain. Especially with docs and example that would explain how the resolvers are actually called.

I think the plugin idea is good, but I don't personally have the need for it right now, and think it's another major step on top of the custom chain/tree feature.
[...]

PR #593, opened 2022-07-06
Add fallbackUpstream to RewriterResolver

adyanth:
Introducing a `fallbackOnFail` flag which passes the query to the next resolver if the answer from the RewriterResolver is nil.

Fixes #587

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -449,7 +449,8 @@ type UpstreamConfig struct {
 
 // RewriteConfig custom DNS configuration
 type RewriteConfig struct {
-	Rewrite map[string]string `yaml:"rewrite"`
+	Rewrite        map[string]string `yaml:"rewrite"`
+	FallbackOnFail bool              `yaml:"fallbackOnFail" default:"false"`
 }
 
 // CustomDNSConfig custom DNS configuration
--- a/config/config.go
+++ b/config/config.go
@@ -449,8 +449,8 @@ type UpstreamConfig struct {
 
 // RewriteConfig custom DNS configuration
 type RewriteConfig struct {
-	Rewrite        map[string]string `yaml:"rewrite"`
-	FallbackOnFail bool              `yaml:"fallbackOnFail" default:"false"`
+	Rewrite          map[string]string `yaml:"rewrite"`
+	FallbackUpstream bool              `yaml:"fallbackUpstream" default:"false"`
 }
 
 // CustomDNSConfig custom DNS configuration
--- a/docs/config.yml
+++ b/docs/config.yml
@@ -35,6 +35,10 @@ customDNS:
 # optional: definition, which DNS resolver(s) should be used for queries to the domain (with all sub-domains). Multiple resolvers must be separated by a comma
 # Example: Query client.fritz.box will ask DNS server 192.168.178.1. This is necessary for local network, to resolve clients by host name
 conditional:
+  # optional: if false (default), return empty result if after rewrite, the mapped resolver returned an empty answer. If true, the original query will be sent to the upstream resolver
+  # Example: The query "blog.example.com" will be rewritten to "blog.fritz.box" and also redirected to the resolver at 192.168.178.1. If not found and if `fallbackOnFailure` was set to `true`, the original query "blog.example.com" will be sent upstream.
+  # Usage: One usecase when having split DNS for internal and external (internet facing) users, but not all subdomains are listed in the internal domain.
+  fallbackOnFailure: false
   # optional: replace domain in the query with other domain before resolver lookup in the mapping
   rewrite:
     example.com: fritz.box
--- a/docs/config.yml
+++ b/docs/config.yml
@@ -36,9 +36,9 @@ customDNS:
 # Example: Query client.fritz.box will ask DNS server 192.168.178.1. This is necessary for local network, to resolve clients by host name
 conditional:
   # optional: if false (default), return empty result if after rewrite, the mapped resolver returned an empty answer. If true, the original query will be sent to the upstream resolver
-  # Example: The query "blog.example.com" will be rewritten to "blog.fritz.box" and also redirected to the resolver at 192.168.178.1. If not found and if `fallbackOnFailure` was set to `true`, the original query "blog.example.com" will be sent upstream.
+  # Example: The query "blog.example.com" will be rewritten to "blog.fritz.box" and also redirected to the resolver at 192.168.178.1. If not found and if `fallbackUpstream` was set to `true`, the original query "blog.example.com" will be sent upstream.
   # Usage: One usecase when having split DNS for internal and external (internet facing) users, but not all subdomains are listed in the internal domain.
-  fallbackOnFailure: false
+  fallbackUpstream: false
   # optional: replace domain in the query with other domain before resolver lookup in the mapping
   rewrite:
     example.com: fritz.box
--- a/resolver/rewriter_resolver.go
+++ b/resolver/rewriter_resolver.go
@@ -72,23 +72,26 @@ func (r *RewriterResolver) Resolve(request *model.Request) (*model.Response, err
 	logger.WithField("resolver", Name(r.inner)).Trace("go to inner resolver")
 
 	response, err := r.inner.Resolve(request)
-	if err != nil {
-		return response, err
-	}
+	// Test for error after checking for fallbackOnFail
 
 	// Revert the request: must be done before calling r.next
 	request.Req = original
 
-	if response == NoResponse {
-		// Inner resolver had no response, continue with the normal chain
-		logger.WithField("next_resolver", Name(r.next)).Trace("go to next resolver")
+	fallbackCondition := err != nil || (response != NoResponse && response.Res.Answer == nil)
+	if r.fallbackOnFail && fallbackCondition {
+		// Inner resolver had no answer, configuration requests fallback, continue with the normal chain
+		logger.WithField("next_resolver", Name(r.next)).Trace("fallback to next resolver")
 
 		return r.next.Resolve(request)
 	}
 
-	if r.fallbackOnFail && response.Res.Answer == nil {
-		// Inner resolver had no answer, configuration requests fallback, continue with the normal chain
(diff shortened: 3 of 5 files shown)


adyanth:
## Config:

```yaml
port: 8053
upstream:
  default:
    - 1.1.1.1

conditional:
  fallbackOnFail: true
  rewrite:
    public.site: private.lan
  mapping:
    private.lan: 10.10.10.4
```

## Status

`uptime.private.lan` does not exist on `10.10.10.4`
`uptime.public.site` exists on `1.1.1.1`

## Flag set to true

### Result

```bash
adyanth@ubuntu-nuc:~/workspace/blocky$ dig @127.0.0.1 -p 8053 uptime.public.site

; <<>> DiG 9.16.1-Ubuntu <<>> @127.0.0.1 -p 8053 uptime.public.site
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 4328
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;uptime.public.site.           IN A

;; ANSWER SECTION:
uptime.public.site.    255     IN AXXX.XXX.XX.XX
uptime.public.site.    255     IN AXXX.XXX.XX.XX

;; Query time: 0 msec
;; SERVER: 127.0.0.1#8053(127.0.0.1)
;; WHEN: Wed Jul 06 14:20:16 IST 2022
;; MSG SIZE  rcvd: 69
```

### Logs

```
[2022-07-06 14:19:32]  INFO queryLog: query resolved answer=A (XXX.XXX.XX.XX), A (XXX.XXX.XX.XX) client_ip=127.0.0.1 client_names=127.0.0.1 duration_ms=220 question=A (uptime.public.site.) response_code=NOERROR response_reason=RESOLVED (tcp+udp:1.1.1.1)
```

## Flag set to false

### Result
```bash
adyanth@ubuntu-nuc:~/workspace/blocky$ dig @127.0.0.1 -p 8053 uptime.public.site

; <<>> DiG 9.16.1-Ubuntu <<>> @127.0.0.1 -p 8053 uptime.public.site
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 59768
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;uptime.public.site.           IN A

;; Query time: 0 msec
;; SERVER: 127.0.0.1#8053(127.0.0.1)
;; WHEN: Wed Jul 06 14:23:39 IST 2022
;; MSG SIZE  rcvd: 48
```

### Logs

```
[2022-07-06 14:23:39]  INFO queryLog: query resolved answer= client_ip=127.0.0.1 client_names=127.0.0.1 duration_ms=1 question=A (uptime.adyanth.site.) response_code=NOERROR response_reason=CONDITIONAL
```

adyanth:
This also works on the `customDNS` block as an added feature!? :)

adyanth:
@0xERR0R can you review it please? I am a bit new to testing methodologies in Go, so not sure if repeating the test cases is the way to do it.

0xERR0R:
Hi, thanks for your work! I'll review and test it, but it  take some time ;(

Can you please describe the new configuration property in `docs/configuration.md` and `docs/config.yml`?

adyanth:
Sure, no worries! Added documentation and fixed the linter warning.

resolver/rewriter_resolver.go
```diff
@@ -70,13 +72,23 @@ func (r *RewriterResolver) Resolve(request *model.Request) (*model.Response, err
...
 	response, err := r.inner.Resolve(request)
-	if err != nil {
-		return response, err
-	}
+	// Test for error after checking for fallbackOnFail
 
 	// Revert the request: must be done before calling r.next
 	request.Req = original
 
+	fallbackCondition := err != nil || (response != NoResponse && response.Res.Answer == nil)
```
ThinkChaos, line 80: IMO querying the upstream when there's an error is not desirable, returning the error to the client seems more logical to me. A concrete case is we're going to leak private DNS requests when the server is down.  
If I understand your use case properly, falling back only when we get `NXDOMAIN` should be enough.  
Is that correct? What do you think about doing it that way?

It would also make the code simpler: the error handling would go back to the classic pattern of being handled immediately. And you could move the fallback block below the `NoResponse`.  
With both of those, the `fallbackCondition` would just be `response.Res.Answer == nil`.

resolver/rewriter_resolver_test.go
```diff
@@ -53,6 +58,7 @@ var _ = Describe("RewriterResolver", func() {
 
 	When("has rewrite", func() {
 		var request *model.Request
+		var fqdnEmptyResponse bool
```
ThinkChaos, line 61: I think `expectEmptyResponse` would be clearer since this changes the behavior for the expect and isn't related to the test's FQDNs.

ThinkChaos (commented):
Did a quick review!  
I think what you arrived at is already really good and just suggested some minor changes :)

resolver/rewriter_resolver_test.go
```diff
@@ -126,6 +137,110 @@ var _ = Describe("RewriterResolver", func() {
...
+
+				return &model.Response{Res: &dns.Msg{Question: req.Req.Question, Answer: nil}}, nil
+			}
+
+			// Resolver after RewriterResolver should not be called `fqdnOriginal`
+			mNext.AssertNotCalled(GinkgoT(), "Resolve", mock.Anything)
+		})
+	})
+
+	When("has rewrite and fallback on failure", func() {
```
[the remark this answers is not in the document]
adyanth: Ahh, I am new to testing in Go, so thanks for that! Will nest it under has rewrite and overwrite the config.

Okay, I will just leave out the repetitive ones and just keep the needed test then.

docs/config.yml
```diff
@@ -35,6 +35,10 @@ customDNS:
 # optional: definition, which DNS resolver(s) should be used for queries to the domain (with all sub-domains). Multiple resolvers must be separated by a comma
 # Example: Query client.fritz.box will ask DNS server 192.168.178.1. This is necessary for local network, to resolve clients by host name
 conditional:
+  # optional: if false (default), return empty result if after rewrite, the mapped resolver returned an empty answer. If true, the original query will be sent to the upstream resolver
+  # Example: The query "blog.example.com" will be rewritten to "blog.fritz.box" and also redirected to the resolver at 192.168.178.1. If not found and if `fallbackOnFailure` was set to `true`, the original query "blog.example.com" will be sent upstream.
+  # Usage: One usecase when having split DNS for internal and external (internet facing) users, but not all subdomains are listed in the internal domain.
+  fallbackOnFailure: false
```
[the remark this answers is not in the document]
adyanth: My bad! Thanks!

config/config.go
```diff
@@ -449,7 +449,8 @@ type UpstreamConfig struct {
 
 // RewriteConfig custom DNS configuration
 type RewriteConfig struct {
-	Rewrite map[string]string `yaml:"rewrite"`
+	Rewrite        map[string]string `yaml:"rewrite"`
+	FallbackOnFail bool              `yaml:"fallbackOnFail" default:"false"`
```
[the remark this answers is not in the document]
adyanth: I know, right! I spent like 15 minutes, could not come up with anything better. `fallbackUpstream` makes sense to me. I will switch to that.

resolver/rewriter_resolver.go
adyanth: I had done that before, the problem is, my upstream for the internal domain times out slower than blocky. I do not want to increase blocky's timeout (unless there is a separate timeout for the conditional block) since that would slow down the resolution for other domains as well. Not just that, I would like a fallback when the resolver is down too.

My argument for your concern is that as soon as you opt for `fallbackUpstream`, you are forgoing the DNS leak protection. It actually makes sense because if you in any case want your DNS queries to go upstream for a domain, it means that the domain is already known to the upstream, just not particular subdomains. Remember that we are not forwarding the query after the mapping, we are forwarding the initial request itself. Meaning, public.site is the one that is sent out, not private.lan which is resolved by the internal resolver only.

In my usecase, I would _*prefer*_ that the queries be sent upstream when the internal resolver is down, so that I can access my services, albeit with a slower, through the internet, connection as opposed to a local one.

resolver/rewriter_resolver_test.go
adyanth: Changed it to `expectNilAnswer` since that is what we are checking for in `AfterEach`.

resolver/rewriter_resolver.go
ThinkChaos: I understand, I think it's fine as is then.

Maybe something to tackle as part of #476.

resolver/rewriter_resolver_test.go
```diff
@@ -126,6 +141,54 @@ var _ = Describe("RewriterResolver", func() {
...
+				Expect(req.Req.Question[0].Name).Should(Equal(fqdnRewritten))
+
+				return &model.Response{Res: &dns.Msg{Question: req.Req.Question, Answer: nil}}, nil
+			}
+
+			// Resolver after RewriterResolver should not be called `fqdnOriginal`
+			mNext.AssertNotCalled(GinkgoT(), "Resolve", mock.Anything)
+		})
+
+		When("has fallback on failure", func() {
```
ThinkChaos, line 164: The old name is left here: "fallback on failure".

ThinkChaos (commented):
Besides the string, everything looks good to me!

resolver/rewriter_resolver_test.go
adyanth: Done!

0xERR0R:
@ThinkChaos: Thanks for review! If it is ok for you, then I can merge it. 👍

PR #595, opened 2022-07-06
Update golang.org/x/crypto to mitigate CVE-2020-9283

PeterDaveHello:
There's a high severity CVE in current golang.org/x/crypto package.

Reference:

- https://www.cve.org/CVERecord?id=CVE-2020-9283


PeterDaveHello:
@0xERR0R @FileGo I'm not quite sure if I'm doing it the right way, it's an indirect dependency, might need your help to take a look at this one. Thanks.

PR #601, opened 2022-07-20
Add .tool-versions file with golang 1.18.4

ashmckenzie:
This PR adds a `.tool-versions` file which can be used by https://asdf-vm.com/ and in this specific case, https://github.com/kennyp/asdf-golang to define the specific version of golang required, which I set as the latest 1.18.x.

The `.tool-versions` file is only used by `asdf`, so for non-asdf users, its existence is a no-op, but for `asdf` users, a simple `asdf plugin add golang && asdf install` will result in golang 1.18.4 being installed 🙂


ashmckenzie:
cc @0xERR0R

0xERR0R:
Hi, thanks for the suggestion! I have not used this before, looks useful. The only concern I have is that we already have multiple places, which golang version is used:

* in GitHub actions for building from CI pipeline (multiple, for example https://github.com/0xERR0R/blocky/blob/88991379cf1a299e96ab0f17abbc7bfe51319252/.github/workflows/ci-build.yml#L13)
* in Dockerfile https://github.com/0xERR0R/blocky/blob/88991379cf1a299e96ab0f17abbc7bfe51319252/Dockerfile#L2 to build the dockerimage
* in go.mod https://github.com/0xERR0R/blocky/blob/88991379cf1a299e96ab0f17abbc7bfe51319252/go.mod#L3

Technically, only go.mod is relevant. If go.mod contains 1.18, you can use golang 1.18 or above to build it. I'm not familiar with asdf, but it looks like it is possible to use the version from go.mod (see https://github.com/kennyp/asdf-golang/pull/32). What do you think?

0xERR0R:
relevant with #618

PR #606, opened 2022-07-21
Add /health endpoint

ashmckenzie:
This PR adds a new `/health` endpoint which for now, just displays the version and is designed to be used to determine API health and inform of the running version for diagnostic/informational purposes:

```sh
$ curl -qs http://127.0.0.1:4000/health | jq .
{
  "version": "v0.18-114-gb9384d5"
}
```

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -30,6 +30,9 @@ const (
 
 	// PathDohQuery DoH Url
 	PathDohQuery = "/dns-query"
+
+	// HealthPath Url
+	HealthPath = "/health"
 )
 
 // QueryRequest is a data structure for a DNS request
@@ -60,4 +63,10 @@ type BlockingStatus struct {
 	DisabledGroups []string `json:"disabledGroups"`
 	// If blocking is temporary disabled: amount of seconds until blocking will be enabled
 	AutoEnableInSec uint `json:"autoEnableInSec"`
+}
+
+// HealthStatus represents the current health status
+type HealthStatus struct {
+	// Display version
+	Version string `json:"version"`
 }
--- a/go.sum
+++ b/go.sum
@@ -98,6 +98,7 @@ github.com/go-redis/redis/v8 v8.11.5/go.mod h1:gREzHqY1hg6oD9ngVRbLStwAWKhA0FEgq
 github.com/go-sql-driver/mysql v1.6.0 h1:BCTh4TKNUYmOmMUcQ3IipzF5prigylS7XXjEkfCHuOE=
 github.com/go-sql-driver/mysql v1.6.0/go.mod h1:DCzpHaOWr8IXmIStZouvnhqoel9Qv2LBy8hT2VhHyBg=
 github.com/go-stack/stack v1.8.0/go.mod h1:v0f6uXyyMGvRgIKkXu+yp6POWl0qKG85gN/melR3HDY=
+github.com/go-task/slim-sprig v0.0.0-20210107165309-348f09dbbbc0 h1:p104kn46Q8WdvHunIJ9dAyjPVtrBPhSr3KT2yUst43I=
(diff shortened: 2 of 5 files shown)


ashmckenzie:
cc @0xERR0R

go.sum
```diff
@@ -98,6 +98,7 @@ github.com/go-redis/redis/v8 v8.11.5/go.mod h1:gREzHqY1hg6oD9ngVRbLStwAWKhA0FEgq
 github.com/go-sql-driver/mysql v1.6.0 h1:BCTh4TKNUYmOmMUcQ3IipzF5prigylS7XXjEkfCHuOE=
 github.com/go-sql-driver/mysql v1.6.0/go.mod h1:DCzpHaOWr8IXmIStZouvnhqoel9Qv2LBy8hT2VhHyBg=
 github.com/go-stack/stack v1.8.0/go.mod h1:v0f6uXyyMGvRgIKkXu+yp6POWl0qKG85gN/melR3HDY=
+github.com/go-task/slim-sprig v0.0.0-20210107165309-348f09dbbbc0 h1:p104kn46Q8WdvHunIJ9dAyjPVtrBPhSr3KT2yUst43I=
```
ashmckenzie, line 101: The change here (and below) were made after running `make test` so am not sure if they should be checked in?

resolver/blocking_resolver.go
```diff
@@ -296,6 +296,12 @@ func (r *BlockingResolver) BlockingStatus() api.BlockingStatus {
 	}
 }
 
+func (r *BlockingResolver) GetHealth() api.HealthStatus {
+	return api.HealthStatus{
+		Version: util.Version,
```
ashmckenzie, line 301: Is `util.Version` the best value?

api/api.go
```diff
@@ -30,6 +30,9 @@ const (
 
 	// PathDohQuery DoH Url
 	PathDohQuery = "/dns-query"
+
+	// HealthPath Url
+	HealthPath = "/health"
```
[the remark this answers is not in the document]
ThinkChaos: I think using `/healthz` would be nice as it's the usual convention for docker and kubernetes.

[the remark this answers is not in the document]
ashmckenzie: Sounds sensible to me, thoughts @0xERR0R ?

0xERR0R (commented):
Thanks for your work!
I'm not sure, what is the purpose of this endpoint... Should it be used as healthcheck in docker/k8s? Or should it just return the version number?

If it should be used as healthcheck -> the implementation of the endpoint will always return 200 if the server is started and no error is occurred, but it doesn't mean that all services, especially the DNS resolution is working. Please check the Dockerfile, we do already have a healthcheck implementation. It uses dig to perform a DNS check with dummy query.

IMHO, if it should be a full-fledged healthcheck, it should also check if all services are working. If it should only check if HTTP endpoints are up, maybe it would be better to rename it to for example "/api/version"?

Can you follow my thougths?

api/api.go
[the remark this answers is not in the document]
0xERR0R: I think, all REST endpoints should start with "/api". But we have already a special endpoint "/dns-query" (because it is the convention) and healtcheck is also something special. I think all 3 variants are Ok.

go.sum
0xERR0R: you can run "go mod tidy", this will update this file. PRs from dependbot with dependency update produce sometimes staled entries.

resolver/blocking_resolver.go
0xERR0R: yes `util.Version" is ok, but the implementation should not be in "blocking_resolver.go", since this operation has nothing to do with blocking. server_endpoints.go would be a better place

PR #608, opened 2022-07-27
Enable start as long as at least one upstream resolver in group is reachable

### LINKED ISSUE #599: Blocky doesn't start when the name of one upstream resolver can't be resolved
I had "https://dns.greyport.space/" in the list of upstream resolver. They seem to have changed their domain and after a restart of blocky the server wouldn't start any more with the following error:

> Error: can't start server: 1 error occurred:
>        * parallel resolver:  no such host dns.greyport.space

Maybe the server could still start with a warning if there is at least one other valid upstream resolver?

FileGo:
Closes #599 by checking that at least one upstream resolver in the group is reachable; if not, blocky will fail to start.

PS: Added VScode folder into `.gitignore`.


resolver/parallel_best_resolver.go
```diff
@@ -38,15 +38,19 @@ type requestResponse struct {
...
 	for name, res := range upstreamResolvers {
 		resolvers := make([]*upstreamResolverStatus, len(res))
 
+		var errResolvers int
+
 		for i, u := range res {
 			r, err := NewUpstreamResolver(u, bootstrap)
 			if err != nil {
-				return nil, err
+				logger.Warnf("upstream group %s: %v", name, err)
```
0xERR0R, line 52: slice in line 41 will be initialized with resolver count. If one resolver throws error, slice will contain a nil resolver. I can reproduce it with following config:
```yaml
upstream:
  default:
  - https://wrong.url1
  - https://wrong.url2
  - https://dns.google/resolve
```

0xERR0R (changes requested):

0xERR0R:
Thanks for your work. If I understand it right, it works only for resolvers defined as hostname/URL and only if DNS resolution fails, right? 
If someone defines only IP addresses of not working resolvers, maybe we should perform a ping or a test query to check if they are alive?

resolver/parallel_best_resolver.go
FileGo: Correct, will fix this.

FileGo:
> Thanks for your work. If I understand it right, it works only for resolvers defined as hostname/URL and only if DNS resolution fails, right?

Looking at it more accurately now, you are right.

> If someone defines only IP addresses of not working resolvers, maybe we should perform a ping or a test query to check if they are alive?

I am thinking about scenario where there is not internet access when blocky starts, then it will fail on start with this kind of logic, but I guess this is a job of system administrator through Docker/systemd config, to ensure system attempts to restart the service, if/when it fails?

What do you think?

0xERR0R:
> 
> I am thinking about scenario where there is not internet access when blocky starts, then it will fail on start with this kind of logic, but I guess this is a job of system administrator through Docker/systemd config, to ensure system attempts to restart the service, if/when it fails?
> 
> What do you think?
In this case it will also fail, because DNS validation (this PR) fails. I would expect, if we check for at least one healthy upstream on start, we should check all of them and not only the DNS resolution of DoH and DoT upstreams. Or we don't check at all, blocky starts and this is a job of a system administrator.

FileGo:
Makes sense, I'll change it so at least one upstream server is reachable and works properly at the time of start.

Thanks.

0xERR0R:
> Makes sense, I'll change it so at least one upstream server is reachable and works properly at the time of start.
> 
> Thanks.

Maybe we should make it configurable? I think, if a server with blocky does a reboot and blocky starts faster than an internal upstream, it would make sense to deactivate this feature.

FileGo:
Interesting, this passes the test suite on my side.

```
Ginkgo ran 11 suites in 18.808065003s
Test Suite Passed
```

Care to take a look? @0xERR0R

0xERR0R:
I just did a merge with development branch.

0xERR0R:
I can reproduce this test error:
```

  Expected success, but got an error:
      <*multierror.Error | 0xc00011c200>: {
          Errors: [
              <*errwrap.wrappedError | 0xc00011c1e0>{
                  Outer: <*errors.errorString | 0xc000436220>{
                      s: "parallel resolver:  unable to reach any DNS resolvers configured for resolver group default",
                  },
                  Inner: <*errors.errorString | 0xc0004361f0>{
                      s: "unable to reach any DNS resolvers configured for resolver group default",
                  },
              },
          ],
          ErrorFormat: nil,
      }
      1 error occurred:
        * parallel resolver:  unable to reach any DNS resolvers configured for resolver group default
      
      
  In [It] at:/server_test.go:699
```

FileGo:
I've fixed the issues in `server_test.go` - I wasn't able to reproduce them in my development environment, as I use a complex setup with DNAT to prevent rogue devices from querying unauthorized servers, so even a query to a random IP address would always work, as it would get translated to my router address :)

I was able to reproduce the issue in `downloader_test.go` by reducing the timeouts to 1 nanosecond, so I guess the test passing will depend on how quick the environment is. I've increased the timeouts generously, and it seems to be passing consistently now. Might be worth into increasing timeouts generally throughout the tests, though.

0xERR0R:
@FileGo 
I tested your changes -> It worked for all my test cases. :+1: 
I changed the default value for "StartVerifyUpstream" to false -> I think we shouldn't change the old behavior for users who install a new version. Is it OK for you? Can you please check my change?

FileGo:
Sure, looks great.

Makes sense not to introduce a breaking change at this stage.

PR #610, opened 2022-07-30
Fix flaky test(537) - second hunt

kwitsch:
Changes to combat flakieness:
- error unwrap to match matcher(lists/downloader)
- Changed some Expected to Eventually with propper timings
- go routine check according to gomega best practice(cmd/serve_test)
- added DeferCleanup to remove temporary files
- changed some (Just)AfterEach to DeferCleanup in (Just)BeforeEach

Affected files:
- /lists/downloader.go
- /lists/downloader_test.go
- /querylog/file_writer_test.go
- /cmd/serve_test.go
- /resolver/blocking_resolver_test.go
- /resolver/query_logging_resolver_test.go

Result:
- sucess rate on my development machine: 9/10 at tests & 8/10 at race
- sucess rate stayes the same even if i throttle the machine with >50% load


lists/downloader_test.go
```diff
@@ -200,7 +200,7 @@ var _ = Describe("Downloader", func() {
 					WithAttempts(3),
 					WithCooldown(time.Millisecond))
 			})
-			It("Should perform a retry until max retry attempt count is reached and return DNSError", FlakeAttempts(5), func() {
+			It("Should perform a retry until max retry attempt count is reached and return DNSError", func() {
```
0xERR0R, line 203: :+1:

lists/downloader.go
```diff
@@ -145,6 +146,16 @@ func (d *HTTPDownloader) DownloadFile(link string) (io.ReadCloser, error) {
 			onDownloadError(link)
 		}))
 
+	var transientErr *TransientError
```
0xERR0R, line 149: Do we need it here? Downloader should just return the error without wrapping/unwrapping. Otherwise the code is polluted with not necessary things. This is a task for caller (in our case the test).

cmd/serve_test.go
```diff
@@ -21,13 +20,18 @@ var _ = Describe("Serve command", func() {
 
 			isConfigMandatory = false
 
+			grClosure := make(chan interface{})
```
0xERR0R, line 23: To be honest, I don't understand my own test. I think the idea was to start the server (without errors) and to shut them down (by sending the signal via "done" channel). But the server starts with error because default upstream is not defined. :confused: 

I think your change is OK here, because the server will not block (since it starts with error).

0xERR0R (commented):
Thanks for your work. I looks good, many improvements! :+1: 

Please check my comments in downloader.go, I think we should discuss it

cmd/serve_test.go
kwitsch: yea seams like it but caused flaky behavior in the following tests as the go routine might still be running depending on the performance & context switches of the testing machine.
I tried to implement th go routine handling acording to [gomega](https://onsi.github.io/gomega/#making-asynchronous-assertions).

lists/downloader.go
kwitsch: Depends on the premise.
I implemented it with the test first premise.
If the premise is implementation first the code can be moved to the test go file.
Main cause of flakiness in this test is that an dnsError is expected but a transient error of type dnsError could be returned if executed in race.

kwitsch: Reworked it by unwrapping transient error in Test instead of downloader.

0xERR0R (approved):
Looks good, thanks!

PR #617, opened 2022-08-03
Managing of external tools with go modules (#616)

### LINKED ISSUE #616: Managing of external tools with go modules
Currently, we use some external tools like ginkgo, swag, go-enum etc. These tools will be downloaded with a fixed version in the Makefile. Updates must be made manually.

There is an interesting way to manage dependencies with go modules:
https://play-with-go.dev/tools-as-dependencies_go115_en/
https://www.jvt.me/posts/2022/06/15/go-tools-dependency-management/

This will allow to use go.mod as the single source of truth. Dependabot will also update those external dependencies

0xERR0R:
closes #616


ThinkChaos:
This is pretty cool!

PR #624, opened 2022-08-06
Dynamic unit test data

kwitsch:
Write test files inside the unittests before they are used.

Added:
 - helpertest.TmpFolder temporary folder for storing files during unit tests
 - helpertest.TmpFile temporary file for usage during unit tests
 - .vscode to .gitignore & .dockerignore

Removed:
 - static testdata folder
 - some os. function usage & replaced with TmpFolder & TmpFile methodes

(first step for creation of test & race docker images)


0xERR0R:
That looks good! Just curious about docker images for test and race: what's your idea?
I want to introduce some end-to-end tests with testcontainers in near future (unfortunately I don't have that much time and have to postpone it again and again.)

kwitsch:
> That looks good! Just curious about docker images for test and race: what's your idea?
> I want to introduce some end-to-end tests with testcontainers in near future (unfortunately I don't have that much time and have to postpone it again and again.)

It's an idea to identify flaky behavior in test&race scenarios faster:
Continuously run both inside multiple docker containers with different cpu&memory limits. The log output should be written to separate log files per container. The images for the containers should be designed to stop if a test or race fails and continually execute them otherwise.

Currently optimizing the default Dockerfile and will develop a separate one for testing in combination with a compose file.

PR #634, opened 2022-08-22
Filter loopback addresses from hosts file

### LINKED ISSUE #633: Filter loopback addresses from hosts file
I have the hostname of my server with the loopback address `127.0.1.1` in my hosts file (which was added by Docker I think) and `127.0.1.1` is one of the addresses returned by blocky when resolving the name of the server.

Blocky does also return the other address of the server and the connection to the server somehow works, but it might be a problem for some devices/applications and the loopback addresses should probably be filtered by blocky when reading the hosts file.

Content of /etc/hosts:
```
127.0.0.1       localhost
::1             localhost ip6-localhost ip6-loopback
ff02::1         ip6-allnodes
ff02::2         ip6-allrouters

127.0.1.1       example.com

192.168.178.2             example.com
fe80::c44a:3704:3be3:3320 example.com
```
The first 4 lines probably come from the OS, the 5th line is probably from Docker and the last 2 where added by me.

Output of `nslookup example.com`:
```
Server:  fritz.box
Address:  :::::fafb

Non-authoritative answer:
Name:    example.com
Addresses:  fe80::c44a:3704:3be3:3320
          127.0.1.1
          192.168.178.2
```

FileGo:
Fixes #633.

### DIFF
--- a/resolver/hosts_file_resolver.go
+++ b/resolver/hosts_file_resolver.go
@@ -150,6 +150,7 @@ type host struct {
 	Aliases  []string
 }
 
+// nolint:funlen
 func (r *HostsFileResolver) parseHostsFile() error {
 	const minColumnCount = 2
 
@@ -197,6 +198,10 @@ func (r *HostsFileResolver) parseHostsFile() error {
 		h.IP = net.ParseIP(fields[0])
 		h.Hostname = fields[1]
 
+		if isLoopbackAddr(h.IP) {
+			continue
+		}
+
 		if len(fields) > minColumnCount {
 			for i := 2; i < len(fields); i++ {
 				h.Aliases = append(h.Aliases, fields[i])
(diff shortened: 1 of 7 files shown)


kwitsch:
Thanks for your work!  
I tested it and it works as intended.  

It could be a breaking change if the hosts file is used to define blocked domains against 127.0.0.1.  

Therefore maybe introduce an config option in the host file section to enable this feature and disable it by default?

resolver/hosts_file_resolver.go
```diff
@@ -211,6 +216,17 @@ func (r *HostsFileResolver) parseHostsFile() error {
 	return nil
 }
 
+func isLoopbackAddr(addr net.IP) bool {
+	_, loopback4, _ := net.ParseCIDR("127.0.0.0/8")
+	loopback6 := net.ParseIP("::1")
```
ThinkChaos, line 221: I think these two vars should be moved out of the function into the file's top level to avoid parsing on every check.

FileGo: Of course, makes sense.

Thanks for the feedback.

kwitsch:
Looks merge ready to me.  
Any objections @ThinkChaos ?

ThinkChaos:
Looks good to me!

PR #636, opened 2022-08-25
Blocky should start resolving DNS traffic as soon as possible(#566)

### LINKED ISSUE #566: [Feature Request] Blocky should start resolving DNS traffic as soon as possible
I have noticed when I first start Blocky, the DNS resolution isn't available until Blocky has parsed all the block and allow lists. On powerful machines such as x86 devices this isn't much of an issue because they can easily parse large blocklists in seconds. However, on the embedded devices (especially routers), I've noticed the initial parsing of large blocklists take quite some time (sometimes over a minute). Until Blocky has done its work, the whole network just stays down.

Would it, at least, be possible to add an option to Blocky, where we can allow DNS resolution to take place before Blocky's DNS filtering abilities kicks in?

P.S.: Since we can now control how many lists are parsed concurrently with `processingConcurrency` attribute. If a person set the value of `processingConcurrency` to `1`; with plenty of large blocklists a device like a low powered router is going to take a long time to get the whole network up and running.

kwitsch:
- marked failStartOnListError as deprecated
- added startStrategy as enum(default, failOnError, fast)
- added documentation

Closes #566


config/config.go
```diff
@@ -46,6 +46,13 @@ type NetProtocol uint16
 // )
 type QueryLogType int16
 
+// StartStrategyType upstart strategy ENUM(
+// default // synchronously download blocking lists on startup
```
0xERR0R, line 50: "default" is not really a meaningful. Maybe "blocking"?

docs/config.yml
```diff
@@ -104,8 +104,8 @@ blocking:
   downloadAttempts: 5
   # optional: Time between the download attempts. Default: 1s
   downloadCooldown: 10s
-  # optional: if true, application startup will fail if at least one list can't be downloaded / opened. Default: false
-  failStartOnListError: false
+  # optional: if failOnError, application startup will fail if at least one list can't be downloaded / opened. Default: false
```
0xERR0R, line 107: "Default: false" -> should be "Default: default"

0xERR0R (commented):
Thanks, this is a cool feature. 👍

0xERR0R (approved):
LGTM :+1:

PR #638, opened 2022-08-26
Docker container rework

kwitsch:
- added healthcheck parameter to blocky
- added build-static to Makefile
- added tzdata and rootcerts to static build
- removed tini in favor of [go-reaper](github.com/ramr/go-reaper) in static build
- changed build to build-static in Dockerfile
- changed healthcheck to use healthcheck parameter in Dockerfile
- changed alpine to scratch in Dockerfile


kwitsch:
Static build still needs some testing regarding DOT & DOH as I had no appropriate test setup on hand.

cmd/healthcheck.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func NewHealthcheckCommand() *cobra.Command {
+	c := &cobra.Command{
+		Use:   "healthcheck",
+		Short: "performs healthcheck",
+		RunE:  healthcheck,
+	}
+
+	c.Flags().Uint16P("port", "p", defaultDNSPort, "healthcheck port 5333")
```
0xERR0R, line 22: should it be port 53?

main_static.go
```diff
@@ -0,0 +1,17 @@
+//go:build linux
+// +build linux
+
+package main
+
+import (
+	_ "time/tzdata"
+
+	_ "github.com/breml/rootcerts"
```
0xERR0R, line 9: I get your idea to embed all external dependencies like tzdata and root certs and I like it. But I have some doubts about security and certificate handling. Can user import own certificates? What is about certificate revocation? Do we need to release in shorter periods?

Dockerfile
```diff
@@ -20,24 +27,26 @@ RUN go mod download
 ADD . .
 
 ARG opts
-RUN env ${opts} make build
+RUN env ${opts} make build-static
+RUN setcap 'cap_net_bind_service=+ep' /src/bin/blocky
+
+RUN adduser -S -D -H -u 10001 -h /app -s /sbin/nologin blocky
```
0xERR0R, line 33: I'm not sure, but I think this will change the user id and it could break installations with CSV query logging

Dockerfile
```diff
@@ -20,24 +27,26 @@ RUN go mod download
...
 USER blocky
 WORKDIR /app
 
-ENTRYPOINT ["/sbin/tini", "--"]
-CMD ["sh", "-c", "/app/blocky --config ${CONFIG_FILE:-/app/config.yml}"]
+ENV BLOCKY_CONFIG_FILE=/app/config.yml
+
+ENTRYPOINT ["/app/blocky"]
+
+HEALTHCHECK --interval=1m --timeout=3s CMD ["/app/blocky", "healthcheck"]
```
0xERR0R, line 52: cool idea 👍

cmd/healthcheck.go
```diff
@@ -0,0 +1,38 @@
...
+
+	return c
+}
+
+func healthcheck(cmd *cobra.Command, args []string) error {
```
0xERR0R, line 27: Maybe we should add some log output here (like OK/NOT), since this command can be executed by user manually?

main_static.go
```diff
@@ -0,0 +1,17 @@
...
+
+	reaper "github.com/ramr/go-reaper"
```
0xERR0R, line 11: Did not know this tool, looks useful. Theoretically, we don't need any init processes, because blocky doesn't spawn any chlild processes. Tini was introduced by a PR to have some kind of a "safety net". Docker also includes "-init" functionality for running containers.

I think we can try go-reaper, I hope it is stable and has no impact at runtime

Dockerfile
```diff
@@ -20,24 +27,26 @@ RUN go mod download
...
 ARG opts
-RUN env ${opts} make build
+RUN env ${opts} make build-static
+RUN setcap 'cap_net_bind_service=+ep' /src/bin/blocky
+
+RUN adduser -S -D -H -u 10001 -h /app -s /sbin/nologin blocky
 
 # final stage
-FROM alpine:3.16
+FROM scratch
```
0xERR0R, line 36: 👍

Dockerfile
```diff
@@ -20,24 +27,26 @@ RUN go mod download
...
-
-HEALTHCHECK --interval=1m --timeout=3s CMD dig @127.0.0.1 -p 53 healthcheck.blocky +tcp +short || exit 1
+COPY --from=build-env /etc/passwd /etc/passwd
 
 USER blocky
 WORKDIR /app
 
-ENTRYPOINT ["/sbin/tini", "--"]
-CMD ["sh", "-c", "/app/blocky --config ${CONFIG_FILE:-/app/config.yml}"]
+ENV BLOCKY_CONFIG_FILE=/app/config.yml
```
0xERR0R, line 48: CONFIG_FILE was changed to BLOCKY_CONFIG_FILE, it could potentially break existing installations

Makefile
```diff
@@ -29,6 +29,10 @@ build:  ## Build binary
 	go generate ./...
 	go build -v -ldflags="-w -s -X github.com/0xERR0R/blocky/util.Version=${VERSION} -X github.com/0xERR0R/blocky/util.BuildTime=${BUILD_TIME}" -o $(BIN_OUT_DIR)/$(BINARY_NAME)$(BINARY_SUFFIX)
 
+build-static:  ## Build static binary
+	go generate ./...
+	go build -tags static -v -ldflags="-linkmode external -extldflags -static -X github.com/0xERR0R/blocky/util.Version=${VERSION} -X github.com/0xERR0R/blocky/util.BuildTime=${BUILD_TIME}" -o $(BIN_OUT_DIR)/$(BINARY_NAME)$(BINARY_SUFFIX)
```
0xERR0R, line 34: Never worked with static binaries. I assume, the binary is bigger (file size), but includes everything to be really portable, correct? This must be also changed in https://github.com/0xERR0R/blocky/blob/8180858efc7dc39b760806b23d847ad01f3d4d81/.goreleaser.yml#L27

0xERR0R (changes requested):
Hey, thanks for your work, looks very good. Please check my questions/remarks on this PR.

0xERR0R:
> Static build still needs some testing regarding DOT & DOH as I had no appropriate test setup on hand.

I can test it. My plan (near future 😄 )is to introduce some e2e tests with testcontainers. This could be a test scenario (create blocky container with configured certificates and enabled DoT and DoH)

kwitsch:
> > Static build still needs some testing regarding DOT & DOH as I had no appropriate test setup on hand.
> 
> I can test it. My plan (near future 😄 )is to introduce some e2e tests with testcontainers. This could be a test scenario (create blocky container with configured certificates and enabled DoT and DoH)

Yeah was also looking into something similar to better combat test flakeyness but wanted to do this first as it reduces container content.

cmd/healthcheck.go
kwitsch: No as this is the default value. Thought the description should be some value different from that.

Dockerfile
kwitsch: That's correct if the container user id isn't mapped to an external one.
Most modern systems use >1000 so I thought I change that accordingly.
Will revert it.

main_static.go
kwitsch: Currently go will lock on the container file system too and include existing certificates if they are in the correct folder.
This can be changed but i thought it is usefull if own certificates need to be added.
There are som infos https://github.com/breml/rootcerts in the readme.
As the revocation of rootcertificates are somwhat rare the release cyclus can stay the same.

main_static.go
kwitsch: It is good practice to have some init process with reaper functionality.
If i remember correctly i participated in the discussion around the introduction of tini. 🤔😅

I also currently use go-reaper as some kind of "safety net" in allmost all my go containers(if they are based on scratch images) without complications so far.

main_static.go
0xERR0R: > Currently go will lock on the container file system too and include existing certificates if they are in the correct folder. 

Ok, that means, if someone uses the binary, it is still possible to import own certificates. In docker it will be also possible if user mounts certificate direction to the right place?

Dockerfile
kwitsch: As the environment will be directly read through blocky it is good practice to prefix environment variables.
There is a fix for backward compatibility in root.go.

Makefile
kwitsch: "the binary is bigger (file size), but includes everything to be really portable" - correct.

Its not neccesarry to build statically outside of a scratch environment as most setups will already include things like tzdata and rootcerts. Most likely this will only waste spache for >90% of the users using binaries outside of containers. 😅

main_static.go
kwitsch: > > Currently go will lock on the container file system too and include existing certificates if they are in the correct folder.
> 
> Ok, that means, if someone uses the binary, it is still possible to import own certificates. In docker it will be also possible if user mounts certificate direction to the right place?

If I got the description correct this should be the case.
This also applies only to the root certificates, are there usecases where custom root(not intermediate) certificates are used?

kwitsch: removed static certs in favor of cert file(same as alpine)

Dockerfile
kwitsch: changed it back to 100

cmd/healthcheck.go
kwitsch: added OK & NOT OK as healthcheck output

Dockerfile
kwitsch: Added fallback to old var if new var isen't used in [initconfig](https://github.com/kwitsch/blocky-dev/blob/fb-dockeroptimization/cmd/root.go#L69).

kwitsch:
@0xERR0R 
Did some changes resolving some of your remarks and added unit tests.
Additionally I updated the versions of Github actions and added a buildcache tag to speed up image creation.

kwitsch:
@0xERR0R it could be necessary to initially create the "ghcr.io/0xerr0r/blocky:buildcache" tag for the changes on the github actions to work(this tag will be overridden after the first successful docker build action run)

0xERR0R (approved):

PR #639, opened 2022-08-27
Change self-signed cert to ECDSA

FileGo:
When I authored #532, I decided to use RSA certificates for increased compatibility. This takes quite a while, particularly on slower platforms, such as ARM based SoCs. The slow speed also affects development in the `server` package, especially if IDE is configured to run tests on every save - these take about 8-9 seconds on my fairly modern development machine.

Any DNS client supporting DoT or DoH will be relatively modern by design (RFCs dated 2016 and 2018, respectively), so they should support ECDSA certificates, which have been supported in OpenSSL for more than a decade.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2,12 +2,14 @@ package server
 
 import (
 	"bytes"
+	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rand"
-	"crypto/rsa"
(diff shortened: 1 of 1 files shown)


0xERR0R:
Thanks for your work. Looks good, it takes <1 sec on my machine to generate the certificate :+1:

PR #645, opened 2022-09-04
Special-Use Domain Names (#635)

### LINKED ISSUE #635: Special-Use Domain Names
At present, when special use domain names are queried (localhost, invalid, etc.), blocky will just forward them to one of the upstream servers:

```
[2022-08-24 13:10:51]  INFO queryLog: query resolved answer=A (127.0.0.1) client_ip=::1 client_names=::1 duration_ms=0 question=A (localhost.) response_code=NOERROR response_reason=RESOLVED (tcp+udp:1.1.1.1)
[2022-08-24 13:11:02]  INFO queryLog: query resolved answer= client_ip=::1 client_names=::1 duration_ms=61 question=A (invalid.) response_code=NXDOMAIN response_reason=RESOLVED (tcp+udp:1.1.1.1)
[2022-08-24 13:11:09]  INFO queryLog: query resolved answer=PTR (localhost.) client_ip=::1 client_names=::1 duration_ms=0 question=PTR (1.0.0.127.in-addr.arpa.) response_code=NOERROR response_reason=RESOLVED (tcp+udp:1.1.1.1)
```

According to [RFC 6761](https://datatracker.ietf.org/doc/html/rfc6761), caching DNS servers should not query authoratitive servers for these domain names.

kwitsch:
Implemented a SpecialUseDomainNames - Resolver as defined in [RFC6761](https://datatracker.ietf.org/doc/html/rfc6761) - 4.  Caching DNS Servers

@FileGo could you review the changes?

Closes #635


resolver/sudn_resolver.go
```diff
@@ -0,0 +1,141 @@
...
+		default:
+			return r.negativeResponse()
+		}
+	}
+
+	return r.next.Resolve(request)
+}
+
+func (r *SudnResolver) Configuration() []string {
+	return []string{"Special-Use Domain Names (RFC 6761)"}
```
FileGo, line 71: This could probably be covered in a test, as for example here: https://github.com/0xERR0R/blocky/blob/48dbc5bc13d1d72f1a42df85c058c0559d1f5ba5/resolver/custom_dns_resolver_test.go#L195

kwitsch: Yea it would actually be a pretty silly test as it is a fixed string and the methode is only included to qualify as ChainedResolver.
I will add it anyway 😅

resolver/sudn_resolver.go
```diff
@@ -0,0 +1,141 @@
...
+	return &model.Response{
+		Res:    response,
+		RType:  model.ResponseTypeSPECIAL,
+		Reason: "Special-Use Domain Name",
+	}, nil
+}
+
+func (r *SudnResolver) loopbackResponseA() (*model.Response, error) {
+	response := new(dns.Msg)
+	response.Rcode = dns.RcodeSuccess
```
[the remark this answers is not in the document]
kwitsch: Thats interessting, dident't know that, im pretty sure we got some other resolvers also doing it the wrong way...i'll look into it.

server/server.go
```diff
@@ -407,6 +407,7 @@ func createQueryResolver(
 		blockingResolver,
 		resolver.NewCachingResolver(cfg.Caching, redisClient),
 		resolver.NewRewriterResolver(cfg.Conditional.RewriteConfig, conditionalUpstreamResolver),
+		resolver.NewSudnResolver(),
```
0xERR0R, line 410: We can move this resolver to the begin of the chain (like fqdnOnly). There is no need to cache response from this resolver or to apply blocking rules

resolver/sudn_resolver.go
```diff
@@ -0,0 +1,145 @@
...
+		"19.172.in-addr.arpa.",
+		"24.172.in-addr.arpa.",
+		"31.172.in-addr.arpa.",
+		"20.172.in-addr.arpa.",
+		"25.172.in-addr.arpa.",
+		"168.192.in-addr.arpa.",
+	}
+}
+
+type SudnResolver struct {
```
0xERR0R, line 41: Was a bit confused about naming. 😕  
Found later as comment: Sudn = Special-Use Domain Names. Maybe we should rename the struct to "SpecialUseDomainNamesResolver" to make it more concise?

0xERR0R (commented):
Good work, thanks! LGTM, I have just 2 questions/notes. Please check it. 👍

server/server.go
kwitsch: We can't move it before custom an conditional as these can possibly return valid responses for such requests.
This resolver is intended to stop those requests from reaching unconditional upstream resolvers.

Moving it to the start would break conditional ARPA responses.

resolver/sudn_resolver.go
kwitsch: Yeah thought this would be a very long name especially as file name.
Only renaming the struct would be more readable without obscenely long filenames.

What do you think?

0xERR0R: > Yeah thought this would be a very long name especially as file name. Only renaming the struct would be more readable without obscenely long filenames.
> 
> What do you think?

Sure, renaming the struct only would be nice. 👍

resolver/sudn_resolver.go
```diff
@@ -0,0 +1,145 @@
...
+
+	rr := new(dns.A)
+	rr.Hdr = dns.RR_Header{
+		Name:   sudnLocalhost,
+		Rrtype: dns.TypeA,
+		Class:  dns.ClassINET,
+		Ttl:    0,
+	}
+
+	rr.A = net.ParseIP("127.0.0.1")
```
ThinkChaos, line 94:
```suggestion
	rr.A = net.ParseIP("127.0.0.1")
```

resolver/sudn_resolver.go
```diff
@@ -0,0 +1,145 @@
...
+
+	rr := new(dns.AAAA)
+	rr.Hdr = dns.RR_Header{
+		Name:   sudnLocalhost,
+		Rrtype: dns.TypeA,
+		Class:  dns.ClassINET,
+		Ttl:    0,
+	}
+
+	rr.AAAA = net.ParseIP("::1")
```
ThinkChaos, line 113: `net` has a `var` for this:

```suggestion
	rr.AAAA = net.IPv6loopback
```

resolver/sudn_resolver.go
ThinkChaos: I think this would be nice as a top-level `var` to avoid parsing the IP every time.

resolver/sudn_resolver.go
```diff
@@ -0,0 +1,145 @@
...
+		"23.172.in-addr.arpa.",
+		"29.172.in-addr.arpa.",
+		"19.172.in-addr.arpa.",
+		"24.172.in-addr.arpa.",
+		"31.172.in-addr.arpa.",
+		"20.172.in-addr.arpa.",
+		"25.172.in-addr.arpa.",
+		"168.192.in-addr.arpa.",
+	}
+}
```
ThinkChaos, line 39: Maybe this can be a `const` instead of a function?

resolver/sudn_resolver.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (r *SpecialUseDomainNamesResolver) Resolve(request *model.Request) (*model.Response, error) {
+	if r.isSpecial(request, sudnArpaSlice()...) ||
+		r.isSpecial(request, sudnInvalid) ||
+		r.isSpecial(request, sudnTest) {
+		return r.negativeResponse(request)
+	} else if r.isSpecial(request, sudnLocalhost) {
+		qtype := request.Req.Question[0].Qtype
+		fmt.Println("QType:", qtype)
```
ThinkChaos, line 56: Looks like there's some debug code left here.

```suggestion
```

ThinkChaos (commented):
Added a couple minor comment as there's a couple places where we can avoid doing some parsing/allocating for each query by using `const`s and `var`s.
Besides that looks like a great addition!

resolver/sudn_resolver.go
kwitsch: Thought it would be more readable using a slice instead of a bunch const strings.
As slices can't be const best practice is to create a function returning a fixed slice (as far as i know).
Please correct me if I'm wrong. 😅

resolver/sudn_resolver.go
kwitsch: Sounds reasonable, changed it.

resolver/sudn_resolver.go
ThinkChaos: Ah didn't remember that of the top of my head :laughing: 

Fine to ignore then!

ThinkChaos (approved):

ThinkChaos:
Slightly related, but maybe worth doing separately, RFC 6762 says:

> Any DNS query for a name ending with ".local." MUST be sent to the
> mDNS IPv4 link-local multicast address 224.0.0.251 (or its IPv6
> equivalent FF02::FB).

So we could also filter those even though they're not special use, it will also prevent leaks from misconfigured clients.

kwitsch:
> Slightly related, but maybe worth doing separately, [RFC 6762](https://datatracker.ietf.org/doc/html/rfc6762) says:
> 
> > Any DNS query for a name ending with ".local." MUST be sent to the
> > mDNS IPv4 link-local multicast address 224.0.0.251 (or its IPv6
> > equivalent FF02::FB).
> 
> So we could also filter those even though they're not special use, it will also prevent leaks from misconfigured clients.

Skimmed over it and and as far as I got it mDNS is somewhat special.
As caching and forwarding has special rules when applied in a mDNS context it would therefore be more reasonable to implement it in a separate resolver in my opinion.

kwitsch:
Just noticed I messed up the 6762 implementation.
Will fix it tomorrow.

kwitsch:
Queries that should be handeled by mDNS(RFC6762) are stopped from leaking.

ThinkChaos (commented):
I have just 2 tiny nits left:

0xERR0R (approved):

PR #648, opened 2022-09-05
add grafana template for Postgres

liboz:
This is branched off of https://github.com/0xERR0R/blocky/blob/development/docs/blocky-query-grafana.json and contains only the necessary changes to make it work in Postgres. Includes a small doc update


0xERR0R:
Thank you for your work. This is really useful :+1:

PR #651, opened 2022-09-12
GitHub workflow rework

kwitsch:
- release workflow doesn't run on forks anymore
- stale workflow doesn't run on forks anymore
- added .git folder to .dockerignore and substituted it with build-args(VERSION & BUILD_TIME)
- removed grype in favor of trivy(grype seams to have problems with scratch images)
- development-docker workflow can be enabled on forks by setting a secret
- ~~automatically initialize buildcache tag on forks~~
- docker layer caching optimizations
- crosscompile arm binaries with go + zig
- test run docker images
- added Architecture to version command
- ci-build as matrix


kwitsch:
Found a bug & got another speedup idea.
Not yet ready to merge.

0xERR0R:
> Found a bug & got another speedup idea. Not yet ready to merge.

I marked this as draft

kwitsch:
Got the build time of development-docker from 20 minutes down to ~4.
But have to do some more testing of the resulting images.

0xERR0R:
I used goreleaser in my other project to create multiarch docker image. The speed was good, goreleaser creates multi arch binaries and just puts each binary in a scratch image. But it creates ugly docker tags.

kwitsch:
Yeah I know also tested it but wasn't really satisfied with the results.
Had to write an extensive config to get better results (post build scripts etc). 😒

I'm working on a solution to do mostly the same with docker,go & zig.

kwitsch:
Sorry missed to squash development commits 😓

kwitsch:
@0xERR0R race seams flaky as it never fails on my fork O.o

0xERR0R:
Skimmed over your changes, I think I got the basic idea and it looks good for me. It looks like it doesn't change the result, right? Have you noticed any positive changes?

kwitsch:
Most notable change is the build time reduction.
Couldn't notice any other difference.

kwitsch:
Just reworked one of my [repos](https://github.com/kwitsch/OmadaSiteDns) from go-releaser to the same build process.
It now has roughly the same build time but nicer docker tags and works like before.

0xERR0R:
@kwitsch: there are some conflicts now. Is there some open points from your side or can it be merged?

kwitsch:
> @kwitsch: there are some conflicts now. Is there some open points from your side or can it be merged?

@0xERR0R : no open points left. Should I resolve the merge conflicts first?

0xERR0R (approved):

util/buildinfo.go
```diff
@@ -6,4 +6,6 @@ var (
 	Version = "undefined"
 	// BuildTime build time of the binary
 	BuildTime = "undefined"
+	// Architecture current CPU architecture
+	Architecture = "amd64"
```
ThinkChaos, line 10: Having `amd64` as the default is surprising IMO. Can we provide the value via the `Makefile` instead like for the other vars?  
`go env GOARCH` should give the correct output.

Dockerfile
```diff
@@ -1,35 +1,60 @@
...
+RUN --mount=type=bind,target=. \
+    --mount=type=cache,target=/root/.cache/go-build \ 
+    --mount=type=cache,target=/go/pkg \
+    go build \
+    -tags static \
+    -v \
+    -ldflags="-X github.com/0xERR0R/blocky/util.Version=${VERSION} -X github.com/0xERR0R/blocky/util.BuildTime=${BUILD_TIME} -X github.com/0xERR0R/blocky/util.Architecture=${TARGETARCH}${TARGETVARIANT}" \
+    -o /bin/blocky && \
+    chown 100 /bin/blocky && \
+    setcap 'cap_net_bind_service=+ep' /bin/blocky
```
ThinkChaos, line 57: Is there a reason this was moved out of the `Makefile`?  
My first reaction is that it would make sense to keep the `util.*` handling in the same spot.

Makefile
```diff
@@ -29,10 +29,6 @@ build:  ## Build binary
 	go generate ./...
 	go build -v -ldflags="-w -s -X github.com/0xERR0R/blocky/util.Version=${VERSION} -X github.com/0xERR0R/blocky/util.BuildTime=${BUILD_TIME}" -o $(BIN_OUT_DIR)/$(BINARY_NAME)$(BINARY_SUFFIX)
 
-build-static:  ## Build static binary
-	go generate ./...
-	go build -tags static -v -ldflags="-linkmode external -extldflags -static -X github.com/0xERR0R/blocky/util.Version=${VERSION} -X github.com/0xERR0R/blocky/util.BuildTime=${BUILD_TIME}" -o $(BIN_OUT_DIR)/$(BINARY_NAME)$(BINARY_SUFFIX)
```
ThinkChaos, line 34: As said in my other comments, I think it'd be nicer to consolidate everything related to building blocky itself in the `Makefile`.  
Here's a minimal one that should be enough to cover all uses and avoid repeating the `-X` flags in different places:
```Makefile
VERSION := $(shell git describe --always --tags)
BUILD_TIME := $(shell date '+%Y%m%d-%H%M%S')
ARCHITECTURE := $(shell go env GOARCH)

DOCKER_IMAGE_NAME := spx01/blocky
BINARY_NAME := blocky
BIN_OUT_DIR := bin

GO_BUILD_FLAGS := -v

GO_BUILD_LD_FLAGS := \
	-w \
	-s \
	-X github.com/0xERR0R/blocky/util.Version=${VERSION} \
	-X github.com/0xERR0R/blocky/util.BuildTime=${BUILD_TIME} \
	-X github.com/0xERR0R/blocky/util.Architecture=${ARCHITECTURE}

GO_BUILD_OUTPUT := $(BIN_OUT_DIR)/$(BINARY_NAME)$(BINARY_SUFFIX)

build: ## Build binary
	go generate ./...
	go build \
		${GO_BUILD_FLAGS} \
		-ldflags="${GO_BUILD_LD_FLAGS}" \
		-o ${GO_BUILD_OUTPUT}

build-static: GO_BUILD_FLAGS += -tags static -linkmode external -extldflags -static
build-static: build ## Build static binary
```

From the `Dockerfile`: `make build-static GO_BUILD_OUTPUT=/bin/blocky`

Also a couple of tiny pre-existing issues with the `Makefile` we can probably fix at the same time:

 - `build-static` is missing from `.PHONY`
 - the shebang is wrong and should be removed

Makefile
```diff
@@ -48,8 +44,13 @@ run: build ## Build and run binary
...
 
-docker-build:  ## Build docker image
-	docker buildx build -o type=docker --network=host -t ${DOCKER_IMAGE_NAME} .
+docker-build:  ## Build docker image 
+	go generate ./...
+	docker buildx \
+	build -o type=docker \
+	--build-arg VERSION=${VERSION} \
+	--build-arg BUILD_TIME=${BUILD_TIME} \
+	--network=host -t ${DOCKER_IMAGE_NAME} .
```
ThinkChaos, line 53: Ident is a bit weird here IMO since flags are aligned with the command.

I'd suggest this, but do as you wish:
```suggestion
	docker buildx build
		-o type=docker \
		--build-arg VERSION=${VERSION} \
		--build-arg BUILD_TIME=${BUILD_TIME} \
		--network=host -t ${DOCKER_IMAGE_NAME} .
```

ThinkChaos:
I'm not familiar with GitHub Actions so can't comment on that. I did suggest some stuff for the `Makefile`. If you think it's out of scope, no worries, I can make my own PR :stuck_out_tongue:

util/buildinfo.go
kwitsch: Yes that's correct.
I will change it. 👍

Dockerfile
kwitsch: Build time reduction.
build-static was only used inside the Docker file and didn't work for cross compilation.
Currently all necessary tools for it are defined in it. This ensures that the image can be build for multiple target systems without rerunning the environment setup again.
Moving these steps to the make file would create overhead without a benefit in my opinion.

Makefile
kwitsch: build-static was never intended for user usage. I only added it to run inside the Docker image creation.
The suggested changes would most likely create broken binaries during cross compilation.
To get propper images I relay on zig inside the Dockerfile.

Dockerfile
ThinkChaos: The main benefit is making the `-X` linker flags DRY but I get your point about user exposure.  
What do you think about having the `Makefile` support passing extra build flags? That way the `Dockerfile` could call `make build EXTRA_GO_BUILD_ARGS="-static [...]"` and we don't duplicate `-X` anywhere.

kwitsch: I get your point and will try to change it.

kwitsch:
> I'm not familiar with GitHub Actions so can't comment on that. I did suggest some stuff for the `Makefile`. If you think it's out of scope, no worries, I can make my own PR 😛

Tried to implement your suggestions.
May you review it as I'm not very familiar with Makefiles? 😅

kwitsch:
> ok i retested it and the development version currently get the wron version. Need to fix it bevore the merge.

Got confused with the versions on my fork as the current version determination systems seams only work on the main repository.
I believe this is a bit out of the scope of this PR.
I'll create a new one when i figured out a proper fix for this issue.

Dockerfile
```diff
@@ -1,35 +1,56 @@
...
+
+#add make & libcap
+RUN apk add --no-cache make libcap
+
+# build binary 
+RUN --mount=type=bind,target=. \
+    --mount=type=cache,target=/root/.cache/go-build \ 
+    --mount=type=cache,target=/go/pkg \
+    export GOARM=${TARGETVARIANT##*v} && \
+    make build
```
ThinkChaos, line 53: Looking at the commits I think you already tried this, but I'm sure it's supposed to work:
```suggestion
    make build GOARM="${TARGETVARIANT##*v}"
```

Other possibility:
```suggestion
    GOARM="${TARGETVARIANT##*v}" make build
```

Makefile
```diff
@@ -1,20 +1,30 @@
...
+BIN_OUT_DIR?=bin
+
+GO_BUILD_FLAGS?=-v
+
+GO_BUILD_LD_FLAGS:=\
+	-w \
+	-s \
+	-X github.com/0xERR0R/blocky/util.Version=${VERSION} \
+	-X github.com/0xERR0R/blocky/util.BuildTime=${BUILD_TIME} \
+	-X github.com/0xERR0R/blocky/util.Architecture=${GOARCH}${GOARM}
```
ThinkChaos, line 18: `GOARCH` and `GOARM` aren't defined anywhere besides the `Dockerfile`, so this won't work for plain `make`.  
Can you add them below the `BUILD_TIME` definition?

Makefile
```diff
@@ -1,20 +1,30 @@
...
+.PHONY: all clean build swagger test lint run fmt docker-build help
+.DEFAULT_GOAL:=help
 
-.PHONY: all clean build swagger test lint run help
-.DEFAULT_GOAL := help
-
-VERSION := $(shell git describe --always --tags)
-BUILD_TIME=$(shell date '+%Y%m%d-%H%M%S')
+VERSION?=$(shell git describe --always --tags)
+BUILD_TIME?=$(shell date '+%Y%m%d-%H%M%S')
```
ThinkChaos, line 5: To go with the above comment:
```suggestion
BUILD_TIME?=$(shell date '+%Y%m%d-%H%M%S')
GOARCH?=$(shell go env GOARCH)
GOARM?=$(shell go env GOARM)
```

Dockerfile
```diff
@@ -1,35 +1,56 @@
...
+
+# build environment
+FROM --platform=$BUILDPLATFORM golang:1-alpine AS build
+
+# required arguments(buildx will set target)
+ARG VERSION
+ARG BUILD_TIME
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETVARIANT
```
ThinkChaos, line 18: Didn't know about these, so makes sense to split and link the docs IMHO:
```suggestion
# required arguments
ARG VERSION
ARG BUILD_TIME

# auto provided by Docker
# https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
ARG TARGETOS
ARG TARGETARCH
ARG TARGETVARIANT
```

Dockerfile
```diff
@@ -1,35 +1,56 @@
...
-
-ENV GO111MODULE=on \
-    CGO_ENABLED=0
-    
-WORKDIR /src
+# get newest certificates
+FROM --platform=$BUILDPLATFORM alpine:3.16 AS ca-certs
+RUN apk add --no-cache ca-certificates
+RUN --mount=type=cache,target=/etc/ssl/certs \
+    update-ca-certificates 2>/dev/null || true
```
ThinkChaos, line 5: Can you please add a comment for why we need ` || true`?

ThinkChaos (changes requested):
I think it's looking way better already. Still have a couple comments though :sweat_smile:

Makefile
kwitsch: Your assumption is wrong as GOARCH & GOARM is set to your system architecture by default.
Its only overwritten in the Dckerfile for crosscompilation purpose.

Makefile
kwitsch: Sorry for my previous comment.
I get the problem....a few minutes to late 😓

Makefile
ThinkChaos: I might have missed something, but to the best of my knowledge, `make` inherits the shell's env, and on my system, those vars are not set. So this would end up with an empty `util.Architecture`.  
Maybe you have a different setup and they get set somehow? But I don't think that's common so not reliable.

Dockerfile
```diff
@@ -1,35 +1,56 @@
...
-ARG opts
-RUN env ${opts} make build-static && \
-    chown 100 /src/bin/blocky && \
-    setcap 'cap_net_bind_service=+ep' /src/bin/blocky
+# setup go+zig
+COPY --from=zig-env /usr/local/bin/zig /usr/local/bin/zig
+ENV PATH="/usr/local/bin/zig:${PATH}" \
+    CC="zigcc" \
+    CXX="zigcpp" \
+    CGO_ENABLED=0 \
```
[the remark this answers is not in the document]
kwitsch: Yea i played around with those for a while.
I only got cross compilation working with zig as there are some components that still use C(? I guess?).

Dockerfile
kwitsch: Sorry assumed it was common Docker knowledge.
Will add more comments.

Makefile
```diff
@@ -26,12 +36,20 @@ serve_docs: ## serves online docs
...
+endif
+	go build $(GO_BUILD_FLAGS) -ldflags="$(GO_BUILD_LD_FLAGS)" -o $(GO_BUILD_OUTPUT)
+ifdef BIN_USER
+	$(info setting owner of $(GO_BUILD_OUTPUT) to $(BIN_USER))
+	chown $(BIN_USER) $(GO_BUILD_OUTPUT)
+endif
+ifdef BIN_AUTOCAB
+	$(info setting cap_net_bind_service to $(GO_BUILD_OUTPUT))
+	setcap 'cap_net_bind_service=+ep' $(GO_BUILD_OUTPUT)
+endif
```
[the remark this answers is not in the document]
kwitsch: Thought this might come in handy for other usecases outside of docker build 🤔

Dockerfile
kwitsch: Added clarifying comment

Dockerfile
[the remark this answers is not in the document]
kwitsch: If i remember correctly we use some parts of the net package that isn't fully compatible "-tags netgo".
Maybe there is a more elegant way to combat this.
My only stable cross compilations all included zig though. 😅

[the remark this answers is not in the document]
ThinkChaos: Ok, not worth spending more time on it then. Can you please add a comment so we remember this?

Makefile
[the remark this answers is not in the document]
ThinkChaos: Fine by me to leave it then!

ThinkChaos (approved):
Besides the last comment for the zig/Cgo thing, everything looks good!

kwitsch:
> Besides the last comment for the zig/Cgo thing, everything looks good!

Looked into it again:
- net & os pages use cgo even if it is disabled
- using pure go core packages can be achieved with tags(netgo & osgo)
- the pure go packages have some limitations regarding the resolver part of netgo

Using zig the limitations don't apply and cross compilation is possible.
In my opinion this has the lowest impact on the compiled binary.

Other possible solutions:
- don't use cross compilation (build time X4)
- use pure go core packages (may need some workarounds)
- use gcc for cross compilation (way more complicated than using zig)

ThinkChaos:
Keeping zig sounds, good thanks for looking into it!  
Just think a small comment to say it's still used even with `CGO_DISABLED` would be nice :)

kwitsch:
Added comments to Dockerfile for clarification why zig is used and to better differentiate the build stages.

ThinkChaos (approved):

PR #653, opened 2022-09-15
Add ability to specify certificate common name.

amigan:
This uses a similar syntax to Unbound to specify the CN for verification in upstream host definitions.


config/config.go
```diff
@@ -100,10 +100,11 @@ var netDefaultPort = map[NetProtocol]uint16{
...
 type Upstream struct {
-	Net  NetProtocol
-	Host string
-	Port uint16
-	Path string
+	Net        NetProtocol
+	Host       string
+	Port       uint16
+	Path       string
+	CommonName *string
```
ThinkChaos, line 107: I think using a plain string (without pointer) and empty string to represent none would be simpler.
Just make sure to add a comment to mention it's optional and can be empty.

config/config.go
```diff
@@ -364,13 +373,27 @@ func ParseUpstream(upstream string) (Upstream, error) {
...
+	hashIdx := strings.Index(in, "#")
+
+	if hashIdx >= 0 {
+		cn = strings.TrimPrefix(in[hashIdx:], "#")
+		upstream = in[:hashIdx]
+	} else {
+		upstream = in
+	}
+
+	return
```
ThinkChaos, line 394: `strings.Cut` can simplify this:
```suggestion
	upstream, commonName, _ := strings.Cut(in, "#")
	return upstream, commonName
```

config/config.go
```diff
@@ -322,12 +323,20 @@ func (s *QTypeSet) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 
 	var port uint16
 
+	var commonName *string
+
+	cn, upstream := extractCN(upstream)
+
+	if cn != "" {
+		commonName = &cn
+	}
```
ThinkChaos, line 338: If you go with both of my above solutions, this can become:
```suggestion
	upstream, commonName := extractCN(upstream)
```

docs/configuration.md
```diff
@@ -51,13 +51,14 @@ following network protocols (net part of the resolver URL):
...
 
 | Parameter | Type                             | Mandatory | Default value                                     |
 |-----------|----------------------------------|-----------|---------------------------------------------------|
-| net       | enum (tcp+udp, tcp-tls or https) | no        | tcp+udp                                           |
-| host      | IP or hostname                   | yes       |                                                   |
-| port      | int (1 - 65535)                  | no        | 53 for udp/tcp, 853 for tcp-tls and 443 for https |
+| net        | enum (tcp+udp, tcp-tls or https) | no        | tcp+udp                                           |
+| host       | IP or hostname                   | yes       |                                                   |
+| port       | int (1 - 65535)                  | no        | 53 for udp/tcp, 853 for tcp-tls and 443 for https |
+| commonName | string                           | no        | the host value                                    |
```
ThinkChaos, line 61: I think the common name's description should make it explicit it's used for cert checking.  
Maybe "the expected certificate common name (host) value"

ThinkChaos (commented):
Great idea! And I like that the syntax follows precedent.

config/config.go
amigan: Yeah, I waffled back and forth about this. I've made the change.

config/config.go
amigan: Good call. I did reverse the order of return values from your suggestion for consistency with the other `extract*()` routines.

docs/configuration.md
amigan: The last column is default value? I've added prose explaining the semantics of the field.

ThinkChaos: Ah yeah my bad!

config/config.go
```diff
@@ -364,13 +367,20 @@ func ParseUpstream(upstream string) (Upstream, error) {
...
+		Port:       port,
+		Path:       path,
+		CommonName: commonName,
 	}, nil
 }
 
+func extractCN(in string) (cn string, upstream string) {
+	upstream, cn, _ = strings.Cut(in, "#")
+
+	return upstream, cn
```
ThinkChaos, line 381: Small typo: forgot to inverse the returned vars.
```suggestion
	return cn, upstream
```

config/config.go
```diff
@@ -364,13 +367,20 @@ func ParseUpstream(upstream string) (Upstream, error) {
...
-		Path: path,
+		Net:        n,
+		Host:       host,
+		Port:       port,
+		Path:       path,
+		CommonName: commonName,
 	}, nil
 }
 
+func extractCN(in string) (cn string, upstream string) {
```
ThinkChaos, line 378: Maybe we can rename the function to `extractCommonName` just to be extra clear for future readers.

ThinkChaos (commented):
Two tiny changes, besides that looks good to me!

ThinkChaos (approved):

PR #656, opened 2022-09-17
Ignore non YAML and non regular files when loading configs from a directory

### LINKED ISSUE #652: Limit scanning of config files only if they end with `.yml` or `.yaml`
While creating a helm chart for Blocky, realized that mounting files from configmap, creates some more files/dirs. Which from a bit of searching show ~~that is for rollback purposes.~~ is for [applying configmap changes](https://livebook.manning.com/concept/kubernetes/mount-configmap-volume)

The problem is that Block trues to scan that files/dirs and fails.

Content directory looks like this:
```
drwxrwsrwx    3 root     568           4096 Sep 12 19:05 .
drwxrwxrwx    3 root     root          4096 Sep 12 19:06 ..
drwxr-sr-x    2 root     568           4096 Sep 12 19:05 ..2022_09_12_19_05_24.266754652
lrwxrwxrwx    1 root     568             31 Sep 12 19:05 ..data -> ..2022_09_12_19_05_24.266754652
lrwxrwxrwx    1 root     568             18 Sep 12 19:05 config.yaml -> ..data/config.yaml
lrwxrwxrwx    1 root     568             28 Sep 12 19:05 redis-prometheus.yaml -> ..data/redis-prometheus.yaml
```

Blocky complains with:

```
[2022-09-12 16:50:55] FATAL unable to load configuration: can't read config files: read /app/config/..2022_09_12_16_47_26.500650961: is a directory
```

It would be nice, to either scan files only ending in `.yaml` or `.yml` Or 
- Check if it's a file 
  - if it is
    - check that has a valid yaml
      - if that's true then parse it. 
Other wise skip but not fail.

Second might be better for the future. But first one should do the job for now!

Edit: Using docker image
repository: spx01/blocky
tag: development@sha256:c55e676e89cee31edeee687d70f7ed957b727d61b5611e213809f7a0399fe4ef

ThinkChaos:
Small changes to ignore non YAML and non regular files in a config dir.
Alternative solution would be to support globing in `--config` but I'm not sure it's worth the trouble as this shouldn't break any sensible setup.

Fixes #652.


kwitsch:
Thanks for your contribution. 🙂👍

PR #684, opened 2022-10-17
feat(api): Adding parameter to use actual remote address or specific IP

Adi3000:
I'm trying to implement new features : for certains ip I would like to enjoy OpenDNS as a DNS fallback over my rule, but I wanted to switch it off and go back to another DNS (like google or cloudflare) like a disabled group (for parental control at home)
To do so, I need to add some testing endpoint.

It is my first Go contribution in my life (even at work) so, do not hesitate to comment (mainly the way I've done inheritence on `dns.ResponseWriter`)

Also, as it is the [Hacktoberfest](https://hacktoberfest.com/) has begun, could you please add the `hacktoberfest-accepted` label to this PR ?

In advance thank

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -38,6 +38,10 @@ type QueryRequest struct {
 	Query string
 	// request type (A, AAAA, ...)
 	Type string
+	// optional : Use remote address : default false
+	UseRemoteAddress bool
+	// optional : Remote address override
+	RemoteAddress string
 }
 
 // QueryResult is a data structure for the DNS result
--- a/server/server_endpoints.go
+++ b/server/server_endpoints.go
@@ -33,6 +33,21 @@ const (
 	corsMaxAge        = 5 * time.Minute
 )
 
+type ApiResponseWriter struct {
+	ip string
+}
+
+func (r *ApiResponseWriter) RemoteAddr() net.Addr {
+	return &net.TCPAddr{IP: net.ParseIP(r.ip)}
+}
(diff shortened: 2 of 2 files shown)


kwitsch:
According to your comment it seems like you want to acquire switchable upstream resolvers for client groups.
Am I right?
Why did you start with api endpoints?

Sorry if I got your intention wrong. 😕
Could you elaborate further?

Adi3000:
For now I have a client group which is using some blocklist addresses (like youtube and tiktok for my kids) and also managed by OpenDns. As we all use the same computer I wanted to disable openDns filtering by switching to classic cloudfare or google DNS (in another way, disable all parental control filtering).
To do so, I wanted to check how and which dns is used for some clientName groups. The API endpoint looks to be the best place to do so and debug the code. It just an easy step to understand the code :)

If you want to check what is in WIP please check this branch. I do not have the knowledge to use the inheritance properly, or the base code is not fully compatible with my needs to avoid some duplicates lines but it might work (didnt tested yet, just finish yesturday) https://github.com/Adi3000/blocky/pull/1/files

Adi3000:
Also I was not clear, my bad, this feature is just one step to debug, nothing to do with dns switch... I should not have mentionned that here...

kwitsch:
Yeah I slowly get your idea. 😅

You like to disable an upstream group so that the specified clients fall back to using the default group until it is enabled again?

Consider blocking the default group from getting disabled as this may break some things.

kwitsch:
I converted your PR to a draft since it is still work in progress.

PR #699, opened 2022-10-29
Support Docker env variables based configuration (#375)

### LINKED ISSUE #375: Support Docker label based configuration
Traefik supports specifying configuration within the docker-compose file.

```yml
pilot:
  token: test

# compose.yml
traefik:
  image: traefik
  labels:
    traefik.pilot.token: test
```

This would be great for blocky and would eliminate the need for maintaining a separate config file.

kwitsch:
Koanf config rework

!UNFINISHED!
Need to rework remaining UnmarshalYAML to hook funcs

Closes #375


config/config_koanf_hooks.go
```diff
@@ -0,0 +1,171 @@
...
+					if qt == q.String() {
+						qtypes = append(qtypes, q)
+
+						break
+					}
+
+					if qi == queryTypeMax {
+						return nil, fmt.Errorf("unknown DNS query type: %s", qt)
+					}
+				}
```
ThinkChaos, line 62: Maybe I misunderstood the code, but I think this should be doable with `dns.StringToType` ([docs](https://pkg.go.dev/github.com/miekg/dns#StringToType)) instead of checking all types.  
It's easy to miss in the `dns` docs though so maybe that's why you did it like this.

config/config_koanf_hooks.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func queryTypeHookFunc() mapstructure.DecodeHookFuncType {
+	return func(
+		f reflect.Type,
+		t reflect.Type,
+		data interface{}) (interface{}, error) {
+		if f.Kind() == reflect.Slice &&
+			t == reflect.TypeOf(QTypeSet{}) {
+			s := reflect.ValueOf(data)
```
ThinkChaos, line 44: I think you can do this to get an actual slice: `s := reflect.ValueOf(data).Interface().([]any)`  
That would make the iteration work with the classic `range` which is nicer IMO.

config/config_koanf.go
```diff
@@ -0,0 +1,97 @@
...
+	"strings"
+
+	"github.com/knadh/koanf"
+	"github.com/knadh/koanf/parsers/yaml"
+	"github.com/knadh/koanf/providers/env"
+	"github.com/knadh/koanf/providers/file"
+	"github.com/mitchellh/mapstructure"
+)
+
+const prefix = "BLOCKY_"
```
ThinkChaos, line 16: Maybe we should consolidate all config related constants? Having all of that in this package would make sense to me, but it's fine to leave as is too.  
I'm thinking of the ones in `root.go`: `defaultConfigPath`, `configFileEnvVar`, `configFileEnvVarOld`.

ThinkChaos (changes requested):
Really cool!

Added a bunch of comments even though I understand this is not final. Hopefully it's good feedback :)

config/config.go
```diff
@@ -615,45 +547,50 @@ func LoadConfig(path string, mandatory bool) (*Config, error) {
...
-
-		return nil, fmt.Errorf("can't read config file(s): %w", err)
-	}
-
-	var data []byte
-
-	if fs.IsDir() {
-		data, err = readFromDir(path, data)
+	var k = koanf.New("_")
+	if path == envKey {
```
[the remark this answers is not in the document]
kwitsch: It's a relict of the conversion since I didn't touch the file loading at first. 😬

If we load from multiple sources the environment variables should override the files in my opinion. Do you agree?

config/config_koanf_hooks.go
```diff
@@ -0,0 +1,171 @@
...
+			// duration is defined as number without unit
+			// use minutes to ensure back compatibility
+			minutes := data.(int)
+
+			result := Duration(time.Duration(minutes) * time.Minute)
+
+			return result, nil
+		}
+
+		return data, nil
```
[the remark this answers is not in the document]
kwitsch: Yeah looks way cleaner.
The int part was added last minute because reading from yaml returns base types but from environment everything is string. As I did environment first...

config/config_koanf_hooks.go
kwitsch: Yes you guessed correct... I missed it🤦‍♂️

config/config_koanf.go
kwitsch: Yeah I think you're right.
All environment related stuff in the same go file with comments would be best.
Currently it's all config related so the config package would be reasonable.🤔

config/config.go
```diff
@@ -548,83 +545,34 @@ func LoadConfig(path string, mandatory bool) (*Config, error) {
...
 
 	k := koanf.New("_")
-	if path == envKey {
-		log.Log().Info("Loading configuration from environment.")
 
-		err := loadEnvironment(k)
-		if err != nil {
-			return nil, fmt.Errorf("can't read environment config: %w", err)
+	fs, err := os.Stat(path)
+	if fs.IsDir() {
```
ThinkChaos: This is using `fs` without checking `err` first.

cmd/root.go
```diff
@@ -70,11 +68,11 @@ func init() {
...
-		val, present := os.LookupEnv(configFileEnvVar)
+		val, present := os.LookupEnv(config.ConfigFilePath)
 		if present {
 			configPath = val
 		} else {
-			val, present = os.LookupEnv(configFileEnvVarOld)
+			val, present = os.LookupEnv(config.ConfigFilePathOld)
 			if present {
 				configPath = val
 			}
```
ThinkChaos, line 78: What do you think of also moving the contents of the `if` to `config/environment.go`?

config/environment.go
```diff
@@ -0,0 +1,10 @@
+package config
+
+const (
+	// Prefix of all environment configurations
+	EnvConfigPrefix = "BLOCKY_"
+	// Environment variable with the path of the config file or folder
+	ConfigFilePath = "BLOCKY_CONFIG_FILE"
+	// Legacy environment variable with the path of the config file or folder
+	ConfigFilePathOld = "CONFIG_FILE"
```
ThinkChaos, line 9: I think these make it sound like they're hardcoded paths. Maybe add a `EnvVar` prefix or suffix to the names?

Maybe also take the oppurtinity to remove "file" from the name (`FilePath` -> `Path`) as the env var values can be a directory.

ThinkChaos (commented):
Really like the way this is shaping up!

A couple of small things I noticed :)

cmd/root.go
kwitsch: Sounds reasonable, i moved the environment files to a separate package as there may config unrelated variables. Will also move the loading logic to this package.

config/config.go
kwitsch: Yeah I know one of the methods I'm currently decluttering. I'll promise to handle it in the future. 😅

ThinkChaos: Yeah no problem just thought I'd comment in case it can help debugging :stuck_out_tongue:

kwitsch:
was sick, will work on it again soon

0xERR0R:
@kwitsch Can you please change the destination branch to main (no hurry), I'd like to delete the development branch

kwitsch:
I'll close it for now and open a new one if I got time to work further on it.
Current state isn't very far and needs excessive merging anyways.

PR #706, opened 2022-11-02
fix: change network names (#703)

### LINKED ISSUE #703: blocky doesn't start if "connectIPVersion" is set
Latest development version (v0.18-230-g3aedd1c) on Docker for windows with following config:

```yaml
upstream:
  default:
    - https://dns.quad9.net/dns-query

queryLog:
  type: console
  
connectIPVersion: v4

port: 53
logLevel: debug
```

Blocky doesn't start:

```
[2022-10-31 08:15:40]  INFO bootstrap: bootstrapDns is not configured, will use system resolver
[2022-10-31 08:15:40]  WARN parallel resolver: upstream group default: unknown network ipv4
Error: can't start server: 1 error occurred:
        * parallel resolver:  no external DNS resolvers configured as default upstream resolvers. Please configure at least one under 'default' configuration name
```

0xERR0R:
Changed network names, according to https://pkg.go.dev/net#Resolver.LookupIP

closes #703

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -48,9 +48,9 @@ func (ipv IPVersion) Net() string {
 	case IPVersionDual:
 		return "ip"
 	case IPVersionV4:
-		return "ipv4"
+		return "ip4"
 	case IPVersionV6:
-		return "ipv6"
+		return "ip6"
 	}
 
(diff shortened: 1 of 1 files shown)


ThinkChaos (approved):
:facepalm:

PR #711, opened 2022-11-03
fix: format ipv6 URL for DoH bootstrap request according to RFC3513

0xERR0R:
It should avoid "magic" timeouts (#701) on Ipv6 bootstrap queries (if DoH is used).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -39,6 +39,7 @@ linters:
     - nilerr
     - nilnil
     - nlreturn
+    - nosprintfhostport
     - prealloc
     - predeclared
     - revive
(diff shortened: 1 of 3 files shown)


ThinkChaos (changes requested):
Good catch!

Can you also enable the [nosprintfhostport](https://github.com/stbenjam/no-sprintf-host-port) linter so it doesn't happen again?

Also I definitely need to get my dev VM setup with IPv6 to catch this stuff.

ThinkChaos (approved):

.golangci.yml
```diff
@@ -39,6 +39,7 @@ linters:
     - nilerr
     - nilnil
     - nlreturn
+    - nosprintfhostport
```
0xERR0R, line 42: This is cool, I didn't know that 👍

PR #731, opened 2022-11-08
feat: caching for empty DNS responses (#700)

### LINKED ISSUE #700: Caching for empty DNS responses
Just noticed that when a client requests A and AAAA records and only the A record exists the AAAA record is tried to get resolved every time by one of the upstream resolvers and the A record is taken from cache. This makes domains without a AAAA record basically as slow as non cached domains.

I understand, that because non existing records don't have a TTL they have to be resolved at some time, but doesn't this kind of defeat caching of DNS entries when a client requests A and AAAA and the A record is still cached but the AAAA is being tried to get resolved all the time? 

Could non existing record types somehow in the cache be tied to existing record types of the same domain, so that they are not tried to be resolved all the time? Or give them a some default TTL?

0xERR0R:
closes #700


kwitsch (approved):
Looks good 👍

ThinkChaos (approved):

PR #741, opened 2022-11-10
fix: add timezone data to docker container

### LINKED ISSUE #740: Timezone ignored in docker container
Timezone (set as env variable TZ) is ignored in development tag:

Current localtime (Europe/Berlin): 11:24

Test with development -> shows UTC:

```sh
docker run --rm -e TZ=Europe/Berlin spx01/blocky:development
[2022-11-10 10:24:29]  INFO _/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/
```

Test with v0.19 -> shows local time

```sh
docker run --rm -e TZ=Europe/Berlin spx01/blocky:v0.19
[2022-11-10 11:25:37]  INFO _/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/_/
```

0xERR0R:
closes #740

### DIFF
--- a/main_static.go
+++ b/main_static.go
@@ -4,6 +4,9 @@
 package main
 
 import (
+	"fmt"
+	"os"
+	"time"
 	_ "time/tzdata"
 
 	reaper "github.com/ramr/go-reaper"
@@ -12,4 +15,13 @@ import (
 //nolint:gochecknoinits
 func init() {
 	go reaper.Reap()
+
+	if tz := os.Getenv("TZ"); tz != "" {
+		var err error
+		time.Local, err = time.LoadLocation(tz)
+
+		if err != nil {
+			fmt.Printf("error loading location '%s': %v\n", tz, err)
+		}
+	}
 }


kwitsch:
This change won't work as it just duplicates the zone database.
The zonedata is already included in the static build.

A possible fix would be an init function in main_static.go which reads the TZ environment variable and sets the localtime with the help of the embedded zonedatabase.

kwitsch (changes requested):
This change won't work as it just duplicates the zone database.
The zonedata is already included in the static build.

A possible fix would be an init function in main_static.go which reads the TZ environment variable and sets the localtime with the help of the embedded zonedatabase.

kwitsch:
Currently unable to test but this should fix the problem if added to the init function of main_static.go:

```
  if tz := os.Getenv("TZ"); tz != "" {
      var err error
      time.Local, err = time.LoadLocation(tz)
      if err != nil {
         log.Printf("error loading location '%s': %v\n", tz, err)
      }
   }
```

kwitsch:
Interesting wouldn't expect that 😅

If you like the APK approach better we should at least remove the native time/tzdata package to reduce image size.
Alternatively there is an approach in the source link which uses the zone info zip file of go.

0xERR0R:
What do you mean with "remove the native time/tzdata package"? There is no tzdata files in the scratch image. I'm not sure if its the best solution to install the tzdata package, but it is common practice.

kwitsch:
The [time/tzdata](https://github.com/0xERR0R/blocky/blob/development/main_static.go#L7) contains the tzdata information as a native golang package.

Sorry for the missunderstanding 😅

Blocky should still work if you remove it because you currently copy the files from the apk installation.

0xERR0R:
ok, got it. I removed the tzdata from image. It makes sense to use it from golang and save the image size. 👍

kwitsch (approved):
Unable to test it at the moment but looks correct code wise. 👍

PR #745, opened 2022-11-12
Fb 319

### LINKED ISSUE #319: queryLog: log instance hostname or identifier to distinguish log entries in multi-instance installation

kwitsch:
- restructured log configuration options
- added hostname configuration option
- uses /etc/hostname if provided(if not os.Hostname() is used)
- reworked documentation

Closes #319


kwitsch:
Just reread the ticket and realized that I missed the point 🤦‍♂️

PR #750, opened 2022-11-14
Fix docs for connectIPVersion param

tmuellerleile:
Should be `enum()` instead of `bool`.


0xERR0R:
Good catch, thank you! :+1:

PR #756, opened 2022-11-18
queryLog: log instance hostname or identifier to distinguish log entries in multi-instance installation(#319)

### LINKED ISSUE #319: queryLog: log instance hostname or identifier to distinguish log entries in multi-instance installation

kwitsch:
Adds hostname field to querylog entry.

If `/etc/hostname` is present the content of this file will be used as hostname.
Otherwise os.Hostname is used.

Closes #319

**Table migration is untested but should work gorm.db.AutoMigrate**

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -79,6 +79,10 @@ services:
     environment:
       - TZ=Europe/Berlin # Optional to synchronize the log timestamp with host
     volumes:
+      # Optional to synchronize the log timestamp with host
+      - /etc/localtime:/etc/localtime:ro
+      # Optional to synchronize the hostname with host
+      - /etc/hostname:/etc/hostname:ro
       # config file
       - ./config.yml:/app/config.yml
 ```
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -72,6 +72,8 @@ services:
     image: spx01/blocky
     container_name: blocky
     restart: unless-stopped
+    # Optional the instance hostname for logging purpose
+    hostname: blocky-hostname
     ports:
       - "53:53/tcp"
       - "53:53/udp"
@@ -81,8 +83,6 @@ services:
     volumes:
       # Optional to synchronize the log timestamp with host
       - /etc/localtime:/etc/localtime:ro
-      # Optional to synchronize the hostname with host
-      - /etc/hostname:/etc/hostname:ro
       # config file
       - ./config.yml:/app/config.yml
 ```
--- a/querylog/file_writer_test.go
+++ b/querylog/file_writer_test.go
@@ -186,7 +186,6 @@ var _ = Describe("FileWriter", func() {
 						DurationMs: 20,
 					})
 				})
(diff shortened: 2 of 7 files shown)


docs/installation.md
```diff
@@ -79,6 +79,10 @@ services:
     environment:
       - TZ=Europe/Berlin # Optional to synchronize the log timestamp with host
     volumes:
+      # Optional to synchronize the log timestamp with host
+      - /etc/localtime:/etc/localtime:ro
+      # Optional to synchronize the hostname with host
+      - /etc/hostname:/etc/hostname:ro
```
0xERR0R, line 85: It will work, but I think it is better to use build in docker feature to pass the hostname to the container (`docker run --hostname xxx`), see https://docs.docker.com/compose/compose-file/compose-file-v3/#domainname-hostname-ipc-mac_address-privileged-read_only-shm_size-stdin_open-tty-user-working_dir for compose

util/hostname.go
```diff
@@ -0,0 +1,40 @@
...
+	return hostname, hostnameErr
+}
+
+// Only return the hostname(may be empty if there was an error)
+func HostnameString() string {
+	return hostname
+}
+
+func getHostname(location string) {
+	hostname, hostnameErr = os.Hostname()
```
0xERR0R, line 32: Do we need the fallback to "/etc/hostname"? Why not just to use the os.Hostname (KISS)?

querylog/file_writer_test.go
```diff
@@ -186,7 +186,6 @@ var _ = Describe("FileWriter", func() {
 						DurationMs: 20,
 					})
 				})
-				fmt.Println(tmpDir.Path)
```
0xERR0R, line 189: :+1:

docs/installation.md
kwitsch: Thought of it more like a solution for replicated instances in a swarm environment.

Setting it with the hostname feature would set it to os.Hostname but wouldn't differentiate between instances if more than one is deployed.

The file based solution would always sync the docker node hostname to the container and wouldn't require additional configuration if the deployment replication is modified.

0xERR0R: os.Hostname is always unique in each container. If you mount "/etc/hostname" and all replicas run on the same host (e.g. docker compose stack), you'll get the same name for each instance.

The goal is to have always a different name per instance.

If you think, there is a use case to sync host name with the host name in container, we can leave this implementation. But we should change the comment/documentation: user should mount "etc/hostname" only if there is only one blocky instance or one instance per host.

util/hostname.go
kwitsch: I like to keep it since it provides an simple solution to the previous described problem of consistent hostnames in a multi instance setup.

0xERR0R (approved):

PR #766, opened 2022-11-24
feat: querylog: define which information should be logged (#765)

### LINKED ISSUE #765: querylog: define which information should be logged

0xERR0R:
closes #765


resolver/query_logging_resolver.go
```diff
@@ -85,6 +90,24 @@ func NewQueryLoggingResolver(cfg config.QueryLogConfig) ChainedResolver {
...
 }
 
+func resolveQueryLogFields(cfg config.QueryLogConfig) []config.QueryLogField {
+	var fields []config.QueryLogField
+
+	if len(cfg.Fields) == 0 {
+		// no fields defined, use all fields as fallback
+		for _, v := range config.QueryLogFieldNames() {
+			qlt, err := config.ParseQueryLogField(v)
+			util.LogOnError("unknown query log field", err)
```
ThinkChaos, line 100: Would be nice to be explicit IMO:
```suggestion
			util.LogOnError("ignoring unknown query log field", err)
```

resolver/query_logging_resolver.go
```diff
@@ -125,6 +144,44 @@ func (r *QueryLoggingResolver) Resolve(request *model.Request) (*model.Response,
...
+		entry.QuestionName = request.Req.Question[0].Name
+		entry.QuestionType = dns.TypeToString[request.Req.Question[0].Qtype]
+	}
+
+	if slices.Contains(r.fields, config.QueryLogFieldDuration) {
+		entry.DurationMs = durationMs
+	}
+
+	return &entry
+}
```
ThinkChaos, line 183: Since we're doing this for every query, and with multiple fields, I think it would make sense to refactor this to not iterate over `r.fields` a bunch of times via `slices.Contains`.

A couple possibilities come to mind:

1. use a set `map[config.QueryLogField]struct{}`/`map[config.QueryLogField]bool`:   
   `if _, ok := r.fields[config.QueryLogFieldClientIP]; ok { ... }`

2. iterate over `r.fields` and have a `switch` that sets the correct info on `entry`

0xERR0R: Yes, that makes sense 👍 will rework this function

ThinkChaos (approved):
Looks good to me :)

kwitsch (approved):

querylog/file_writer.go
```diff
@@ -102,18 +101,17 @@ func (d *FileWriter) CleanUp() {
...
 	return []string{
 		logEntry.Start.Format("2006-01-02 15:04:05"),
-		request.ClientIP.String(),
-		strings.Join(request.ClientNames, "; "),
+		logEntry.ClientIP,
+		strings.Join(logEntry.ClientNames, "; "),
 		fmt.Sprintf("%d", logEntry.DurationMs),
-		response.Reason,
-		util.QuestionToString(request.Req.Question),
-		util.AnswerToString(response.Res.Answer),
```
0xERR0R, line 115: Obfuscates question and answer if log privacy is enabled

resolver/query_logging_resolver.go
```diff
@@ -125,6 +143,42 @@ func (r *QueryLoggingResolver) Resolve(request *model.Request) (*model.Response,
...
+		case config.QueryLogFieldResponseReason:
+			entry.ResponseReason = response.Reason
+			entry.ResponseType = response.RType.String()
+			entry.ResponseCode = dns.RcodeToString[response.Res.Rcode]
+
+		case config.QueryLogFieldResponseAnswer:
+			entry.Answer = util.AnswerToString(response.Res.Answer)
+
+		case config.QueryLogFieldQuestion:
+			entry.QuestionName = request.Req.Question[0].Name
```
0xERR0R, line 171: should use util.QuestionToString -> this will obfuscate question if log privacy is enabled

PR #775, opened 2022-11-30
fix: `startVerifyUpstream` not disabling all start checks

### LINKED ISSUE #749: parameter 'startVerifyUpstream = false' checks upstream on start
When following configuration is used:

```yaml
upstream:
  default:
    - some.wrong.host
startVerifyUpstream: false
```

Then blocky fails to start with following message: `upstream group default: lookup some.wrong.host on 127.0.0.11:53: no such host`

I would expect, the configuration "startVerifyUpstream: false" will disable all startup checks. This is useful if server gets restarted and DNS resolution is not ready if blocky is started.

ThinkChaos:
Fixes #749

### DIFF
--- a/e2e/blocking_test.go
+++ b/e2e/blocking_test.go
@@ -72,7 +72,7 @@ var _ = Describe("External lists and query blocking", func() {
 				})
 
 				It("should fail to start", func() {
-					Expect(blocky.IsRunning()).Should(BeFalse())
+					Eventually(blocky.IsRunning).Should(BeFalse())
 
 					Expect(getContainerLogs(blocky)).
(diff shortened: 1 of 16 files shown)


ThinkChaos:
Looks like e2e tests are randomly failing.  
Worked fine in my fork, but for the MR even with a retry it didn't pass and each try had a different test fail.

0xERR0R:
> Looks like e2e tests are randomly failing. Worked fine in my fork, but for the MR even with a retry it didn't pass and each try had a different test fail.

I think, we'll need some time to get them stable on GH environment. Please just ignore them, I'll take a look

0xERR0R (approved):
LGTM 👍   

I also added 2 additional e2e tests.

PR #779, opened 2022-12-02
refactor(resolver): only build mocks during tests

ThinkChaos:
This makes iterating with breaking refactors easier as you can also break the mocks without breaking normal compilation.

Also unexport code only used in the resolver package. Only `MockUDPUpstreamServer` is kept as non test and exported as it is used by the server package.


kwitsch (approved):
👍

PR #781, opened 2022-12-03
style: fix lint

ThinkChaos:

PR #782, opened 2022-12-03
feat(bootstrap): support multiple upstreams

### LINKED ISSUE #760: Use multiple upstreams for bootstrapDns.
Correct me if I'm wrong, but it seems now I can define only one upstream for bootstrapDns. In worst case scenario when my upstream is down, blocky won't download filters when started.

Can we have bootstrapDns.upstream as a list, maybe?

ThinkChaos:
If more than one upstream is configured, they are raced via a `ParallelBestResolver`.

Other improvements:
- feat(bootstrap): support IP only encrypted DNS
  Also make `tcp+udp` upstreams use any IPs provided.
- feat: always prefetch upstream IPs to avoid stalling user queries
  Otherwise, a request to blocky could end up waiting for 2 DNS requests:
    1. lookup the DNS server IP
    2. forward the user request to the server looked-up in 1
- feat: stack log prefixes to differentiate between log emitters
  The goal is to be able to tell apart logs from difference sources, such as `bootstrap.parallel_best_resolver` and `parallel_best_resolver`.

Also includes #781 so the lint passes here.

Fixes #760.

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -3,12 +3,15 @@ package log
 //go:generate go run github.com/abice/go-enum -f=$GOFILE --marshal --names
 
 import (
+	"fmt"
 	"io"
 	"strings"
 
 	"github.com/sirupsen/logrus"
 	prefixed "github.com/x-cray/logrus-prefixed-formatter"
 )
+
+const prefixField = "prefix"
 
 // Logger is the global logging instance
 //
@@ -60,7 +63,16 @@ func Log() *logrus.Logger {
 
 // PrefixedLog return the global logger with prefix
 func PrefixedLog(prefix string) *logrus.Entry {
-	return logger.WithField("prefix", prefix)
+	return logger.WithField(prefixField, prefix)
+}
+
+// WithPrefix adds the given prefix to the logger.
+func WithPrefix(logger *logrus.Entry, prefix string) *logrus.Entry {
+	if existingPrefix, ok := logger.Data[prefixField]; ok {
+		prefix = fmt.Sprintf("%s.%s", existingPrefix, prefix)
+	}
+
+	return logger.WithField(prefixField, prefix)
 }
 
 // EscapeInput removes line breaks from input
--- a/resolver/bootstrap.go
+++ b/resolver/bootstrap.go
@@ -63,11 +63,23 @@ func NewBootstrap(cfg *config.Config) (b *Bootstrap, err error) {
 		return nil, fmt.Errorf("could not create bootstrap ParallelBestResolver: %w", err)
 	}
 
+	// Always enable prefetching to avoid stalling user requests
+	// Otherwise, a request to blocky could end up waiting for 2 DNS requests:
+	//   1. lookup the DNS server IP
+	//   2. forward the user request to the server looked-up in 1
+	cachingCfg := cfg.Caching
+	cachingCfg.EnablePrefetch()
+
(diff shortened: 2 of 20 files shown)


log/logger.go
```diff
@@ -60,7 +63,16 @@ func Log() *logrus.Logger {
...
+	return logger.WithField(prefixField, prefix)
+}
+
+// WithPrefix adds the given prefix to the logger.
+func WithPrefix(logger *logrus.Entry, prefix string) *logrus.Entry {
+	if existingPrefix, ok := logger.Data[prefixField]; ok {
+		prefix = fmt.Sprintf("%s.%s", existingPrefix, prefix)
+	}
+
+	return logger.WithField(prefixField, prefix)
```
ThinkChaos, line 75: I haven't looked into it, but this might break the dashboards or whatever else depends on the prefix being exactly the name of the resolver.

resolver/bootstrap_test.go
```diff
@@ -319,4 +398,33 @@ var _ = Describe("Bootstrap", Label("bootstrap"), func() {
...
+
+		It("uses both", func() {
+			_, err := sut.resolve("example.com.", []dns.Type{dns.Type(dns.TypeA)})
+
+			Expect(err).To(Succeed())
+
+			Expect(mockUpstream1.GetCallCount()).To(Equal(1))
+			Expect(mockUpstream2.GetCallCount()).To(Equal(1))
+		})
+	})
```
ThinkChaos, line 429: This test was working locally, but on the CI it's flaky. I'll rewrite it differently.

ThinkChaos (commented):
I recommend reviewing commit by commit. Everything should be relatively simple that way.

0xERR0R (commented):
LGTM 👍 

I added one test case to check the conversion of the new YAML structure (list of bootstrap entries).

Could you please change the documentation (configuration.md and config.yaml in docs folder)?

I think it would also make sense to write some e2e tests to test the bootstrap resolution, but this is not a part of this PR/issue. I'll take a look, how it can be achieved.

resolver/bootstrap.go
```diff
@@ -29,56 +29,58 @@ type Bootstrap struct {
...
+	// Otherwise, a request to blocky could end up waiting for 2 DNS requests:
+	//   1. lookup the DNS server IP
+	//   2. forward the user request to the server looked-up in 1
+	cachingCfg := cfg.Caching
+	cachingCfg.EnablePrefetch()
+
+	if cachingCfg.MinCachingTime == 0 {
+		// Set a min time in case the user didn't to avoid prefetching too often
+		cachingCfg.MinCachingTime = config.Duration(time.Hour)
+	}
```
ThinkChaos, line 76: Testing this some more, I noticed that `caching.minTime` has no default.  
Since with this MR, the caching resolver is always enabled with prefetching as part of the `Bootstrap`, this can cause a lot of churn for not much gains.
I noticed because the domain `s3.amazonaws.com` used by the list `https://s3.amazonaws.com/lists.disconnect.me/simple_ad.txt` I have in my test config has a TTL of 3, and was being prefetched every 10 seconds.

I added a default of 1h here, only if the user hasn't specified anything (or they put 0). Let me know if you think I should revert this and we should just trust the TTLs.

ThinkChaos:
Added the docs, rebased and cleaned up the commits. This should be all good now if you approve!

PR #783, opened 2022-12-03
style(resolver): standardize configuration disabled implementations and tests

ThinkChaos:
Also includes #781 so the CI passes here.


0xERR0R (approved):
Looks very clear :+1:

PR #798, opened 2022-12-13
Test refactoring

0xERR0R:
Some small improvements:

- make tests more readable:
  - replace for example `dns.Type(dns.TypeA)` with `A` constant
  - add special assert functions to check TTL, Response, Return type
  - Squash multiple assertions on Response to one "Expect" call
  - remove "hidden" and "nested" assertions and variable definition in "beforeEach"
  - DNS response assertions can now be applied direct to model.Response, this make error check unnecessary (since gomega does it)

- remove TTL check from ´BeDNSRecord´ and use dedicated `HaveTTL` which allows to use also "<=" comparison and make TTL assertion optional (without "hack" with 0 TTL before)
- some small improvement on "Eventually" calls

After removing of sleep in a test, a race condition in cache is occurred. This was also fixed by a separate commit.

### DIFF
--- a/e2e/metrics_test.go
+++ b/e2e/metrics_test.go
@@ -9,7 +9,6 @@ import (
 
 	. "github.com/0xERR0R/blocky/helpertest"
 	"github.com/0xERR0R/blocky/util"
-	"github.com/miekg/dns"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/testcontainers/testcontainers-go"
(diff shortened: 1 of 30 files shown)


e2e/metrics_test.go
```diff
@@ -75,38 +74,57 @@ var _ = Describe("Metrics functional tests", func() {
...
-					Eventually(fetchBlockyMetrics).WithArguments(metricsURL).Should(ContainElement("blocky_cache_hit_count 0"))
-					Eventually(fetchBlockyMetrics).WithArguments(metricsURL).Should(ContainElement("blocky_cache_miss_count 1"))
+					Expect(doDNSRequest(blocky, msg)).
+						Should(SatisfyAll(
+							BeDNSRecord("google.de.", A, "1.2.3.4"),
+							HaveTTL(BeNumerically("==", 123)),
+						))
+
+					Eventually(fetchBlockyMetrics).WithArguments(metricsURL).Should(
+						SatisfyAll(
```
kwitsch, line 96: I think using one indentation style for `Should` `SatisfyAll` would benefit the readability of the tests.
Currently there are:
```Go
.Should(
  SatisfyAll(
```
```Go
.Should(SatisfyAll(
```
```Go
.
Should(
  SatisfyAll(
```

0xERR0R: I agree with this in principle. We probably have other examples in the code that should be reformatted along the same lines. I would like to "fix" it in some automated way. I have tried the tool "[golines](https://github.com/segmentio/golines)". It breaks long lines, but in this particular case the line is too short. Do you have any idea how to automate this?

ThinkChaos: I haven't looked at the docs to try and figure out if it can do this, but [gofumpt](https://github.com/mvdan/gofumpt) is "a stricter gofmt", and might have a rule that applies.

kwitsch: Skimmed over it and it seems it won't solve this particular problem but looks like a great tool for readability enhancement.

kwitsch: Tested gofump and unfortunately it breaks the "only one cuddle assignment allowed before if statement" linter rule.
I tinker around with it for a bit.

0xERR0R: I added linebreaks on "SatisfyAll" manually

0xERR0R:
rebased on latest development branch

kwitsch (approved):
I really like how `SatisfyAll` makes the unit tests way more readable.

`dns.Type` constants 👍
`defer` ➡ `DeferCleanup` 👍

PR #805, opened 2022-12-17
Added gofumpt

kwitsch:
Added [gofumpt](https://github.com/mvdan/gofumpt) to `make fmt` command.

Issue #709, opened 2022-11-03
Reworking of external build tools management


0xERR0R:
With #617 I introduced the "tools.go" pattern (https://www.jvt.me/posts/2022/06/15/go-tools-dependency-management/). With this approach, we can "pin" the versions of external tools (like golangci-lint etc) in just one file (no need to update multiple files (Makefile, GitHub actions etc) to update one tool.

This approach has some downsides if the external tool needs a different version of the dependency. Dependabot tries to downgrade the golangci-lint in each PR.

I found this promising tool (https://github.com/bwplotka/bingo) and a blog article here (https://www.bwplotka.dev/2020/bingo/). I think it should help since it separates the go.mod for external binaries from the main go.mod

ThinkChaos:
Seems nice. I like it doesn't require installing the tool if you don't want to update/add to the tools, and that it has a Makefile integration.  
I wonder how it interacts with dependabot.

0xERR0R:
I'm not sure it it is the best solution. Maybe we should use the tools.go approach but with a separate go.mod (like https://github.com/prometheus-operator/prometheus-operator/blob/main/scripts/tools.go). I'll try it out.

I think, we need dependabot only for direct blocky dependencies. If it can also handle all external tools -> it is nice, but is not necessary.

0xERR0R:
I tried "bingo" -> doesn't work for 1.19 go.mod files (https://github.com/bwplotka/bingo/issues/125)

0xERR0R:
fixed via #1733

PR #816, opened 2023-01-04
chore: define version of golangci-lint in makefile

0xERR0R:
If golangci-lint version is defined in tools.go (via go.mod), golang downgrades this version to 1.42 (instead of using the latest version v.50.1). The definition of version was moved from tools.to to Makefile in #709, but it was moved again in tools.go in #805. The weird thing: It works on main build, but it doesn't work on dependabot PRs (if some dependencies are being updated)


kwitsch (approved):
Unable to test it at the moment but looks correct logic wise.

PR #837, opened 2023-01-20
Proposal: use revgen instead of 'go generate'

kwitsch:
Use [revgen(v0.1.0)](https://github.com/inigolabs/revgen) instead of `go generate`.
Version 0.2.0 is not used since it lacks propper documentation.

The Makefile uses revgen in the build process therfore only changed files are newly generated.
If a new file uses `//go:generate` it has to be added to the revgen config by running `go run github.com/inigolabs/revgen update` and configuring the `.revgen.yaml` afterwards.

This should speed up `make build` since the `//go:generate` parts are only executed if the files changed.


ThinkChaos:
My first thought is it would be nicer to use make's ability to not run rules that aren't needed instead of using a new external tool.  

Here's a minimal example:
```Makefile
config/config_enum.go: config/config.go
	go generate $<
```

With that `go generate config/config.go` will only run when `config.go` is newer than `config_enum.go`:

```sh
$ make config/config_enum.go
go generate config/config.go
go-enum started. file: config.go
go-enum finished. file: config.go
$ make config/config_enum.go
make: 'config/config_enum.go' is up to date.
$ touch config/config.go
$ make config/config_enum.go
go generate config/config.go
go-enum started. file: config.go
go-enum finished. file: config.go
```

kwitsch:
CI won't run code generation as this should be done before a push.

Found the tool an thought of it as useful but on second thought it only adds complexity with only 4 affected files.

Will close it.

ThinkChaos:
Just noticed that `go-enum` has Makefile rules in the README: https://github.com/abice/go-enum#using-makefile
Could be nice to use those.

PR #857, opened 2023-01-31
Workflow changes

kwitsch:
Goal of this PR is to combat flakyness & streamline workflows.

Content:
- changed timings of Eventually to combat flakyness
- merged Build and e2e workflow(Makefile Workflow)
- added go caching in workflow(Makefile Workflow)
- skip docker or go checks by matrix fields(Makefile Workflow)
- merged lint to Makefile Workflow


kwitsch:
The flakiness of e2e unit tests could be reduced by reducing the Eventually polling times from default 10ms to 2ms.

There seems to be a problem with with the upstart of test containers that occured randomly(?).

During my test i could reproduce the errors more often if polling times where higher (100-500ms).

0xERR0R (approved):
looks good 👍

0xERR0R:
> The flakiness of e2e unit tests could be reduced by reducing the Eventually polling times from default 10ms to 2ms.
> 
> There seems to be a problem with with the upstart of test containers that occured randomly(?).
> 
> During my test i could reproduce the errors more often if polling times where higher (100-500ms).

Reducing the polling time from 10ms to 2ms will increase the load and to be honest, I don't understand why it reduces the test flakyness 🤷‍♂️ . IMHO the poling time should have no effect, only the duration is relevant.

Issue #867, opened 2023-02-08
[Feature Request] Option to get hosts file from http url


adb76:
Hello,

as far a I understood the configuration options correctly, currently it is only possible to get a hosts configuration from a file on the file system:

```
hostsFile:
  filePath: /etc/hosts
  hostsTTL: 24h
  refreshPeriod: 6h
```

It would be great to have an option to get it additionally or optionally also from an http url (analog to the ad blacklists), e.g.:

```
hostsFile:
  hosts:
    - /etc/hosts
    - https://some.url/hosts
  hostsTTL: 24h
  refreshPeriod: 6h
```

What is the idea behind it? Well, I am using an DNS service to avoid geoip blocking, which offers an URL to an up to date hosts list, which of course can change over time. So the idea is to avoid manual download of that list - or to avoid to forget it to do so ;-) 

Kind regards.

ThinkChaos:
Started working on this by unifying the `HostsFileResolver` and `ListCache` download and parsing code.  

From a quick look at go-getter, I'm not sure it'd be a quick replacement since it doesn't seem to have a way to set the HTTP transport (we need it so the resolving uses the bootstrap DNS). Also we'd need to also use the bootstrap for DNS with the other protocols, so that would need a rework anyways.

I'll do a MR with just re-using our existing code, and we can always decide later for go-getter, or something else, to support more protocols if we want.

kwitsch:
@ThinkChaos should this issue be closed since your branch is already merged?

ThinkChaos:
First MR only allows using a link, I'm still working on the multiple sources code.  
Currently setting up a module with sync algorithms and primitives to use. It's mostly done, have it coded and integrated in blocky, working on the CI ATM. Doing a setup to have coverage in PRs with just GH actions, hoping we can reuse that here too since codecov has some quirks.

TL;DR: WIP, coming soon™

kwitsch:
Ok 👍

Was just asking since a successful merge is linked but the issue is still open. 😅

PR #884, opened 2023-02-18
feat: allow hosts file resolver to use a HTTP(S) link or inline block

ThinkChaos:
This is a first step towards #867.  
In example config from the issue, @adb76 also uses multiple hosts files, which this does not implement. I'll do that as a follow up so this PR doesn't get too huge.

I ended up doing a bit more than "just re-using our existing code" as I said in the issue:
- Had a bit of fun with the list parsing implementation. It's streaming now, so we don't need to have the full file in memory in addition to the storage where we keep the parsed data. We do still have 2 copies of the parsed data while we refresh everything.  
  There aren't tests for those new types/functions yet, I first wanted some feedback.
  All commits pass all existing tests though.
- I changed the `HostsFileResolver` so it uses a map internally so lookups are faster (actually multiple to also speedup reserve lookups)

EDIT: force-pushed to  remove a `FDescribe` I accidentally committed.

### DIFF
--- a/lists/list_cache.go
+++ b/lists/list_cache.go
@@ -2,7 +2,7 @@ package lists
 
 //go:generate go run github.com/abice/go-enum -f=$GOFILE --marshal --names
 import (
-	"bufio"
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -12,20 +12,20 @@ import (
 	"sync"
 	"time"
 
-	"github.com/0xERR0R/blocky/cache/stringcache"
+	"github.com/hako/durafmt"
+	"github.com/hashicorp/go-multierror"
 	"github.com/sirupsen/logrus"
 
-	"github.com/hako/durafmt"
-
-	"github.com/hashicorp/go-multierror"
-
+	"github.com/0xERR0R/blocky/cache/stringcache"
 	"github.com/0xERR0R/blocky/evt"
+	"github.com/0xERR0R/blocky/lists/parsers"
 	"github.com/0xERR0R/blocky/log"
 )
 
 const (
 	defaultProcessingConcurrency = 4
 	chanCap                      = 1000
+	maxErrorsPerFile             = 5
 )
 
 // ListCacheType represents the type of cached list ENUM(
@@ -147,6 +147,8 @@ func logger() *logrus.Entry {
 }
 
 // downloads and reads files with domain names and creates cache for them
+//
+//nolint:funlen // will refactor in a later commit
 func (b *ListCache) createCacheForGroup(links []string) (stringcache.StringCache, error) {
 	var err error
 
@@ -162,11 +164,14 @@ func (b *ListCache) createCacheForGroup(links []string) (stringcache.StringCache
 
 	processingLinkJobs := len(links)
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
 	// loop over links (http/local) or inline definitions
 	// start a new goroutine for each link, but limit to max. number (see processingConcurrency)
-	for _, link := range links {
-		go func(link string) {
-			// thy to write in this channel -> this will block if max amount of goroutines are being executed
+	for idx, link := range links {
+		go func(idx int, link string) {
+			// try to write in this channel -> this will block if max amount of goroutines are being executed
 			guard <- struct{}{}
 
 			defer func() {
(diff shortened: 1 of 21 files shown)


lists/list_cache.go
```diff
@@ -147,6 +147,8 @@ func logger() *logrus.Entry {
 }
 
 // downloads and reads files with domain names and creates cache for them
+//
+//nolint:funlen // will refactor in a later commit
```
ThinkChaos, line 151: I'll refactor this in the follow-up.  
I already have something locally, but think it'll be better in the PR that allows the `HostsFileResolver` to use multiple sources since that is going to look a lot like this function.

lists/list_cache.go
```diff
@@ -200,7 +211,12 @@ Loop:
 		}
 	}
 
-	return factory.Create(), err
+	cache := factory.Create()
+	if cache.ElementCount() == 0 && err != nil {
+		cache = nil // don't replace existing cache
```
ThinkChaos, line 216: This is a change from the previous behavior (I also changed a test). I think makes more sense to keep the current cache if the parsing completely failed.  
I can easily revert it if you prefer the current behavior as it's isolated to 8caf7c528408f0424b0e467d28ff3de41442ad9c.

lists/list_cache.go
```diff
@@ -231,35 +247,54 @@ func (b *ListCache) refresh(init bool) error {
...
+		logger().WithFields(logrus.Fields{
+			"group":       group,
+			"total_count": cacheForGroup.ElementCount(),
+		}).Info("group import finished")
 	}
 
 	return err
 }
 
+func (b *ListCache) groupElementCount(group string, isInit bool) int {
```
0xERR0R, line 282: In "Configuration()" we do already count elements per group. Maybe we can reuse it?

lists/list_cache_test.go
```diff
@@ -131,7 +131,7 @@ var _ = Describe("ListCache", func() {
 			})
 		})
 		When("non transient err occurs on download", func() {
-			It("should delete existing elements from group cache", func() {
+			It("should keep existing elements from group cache", func() {
```
0xERR0R, line 134: see comment above

0xERR0R (commented):
Skimmed over your code. I leaved some comments, some minor points. The parsers package is difficult to understand, unit tests would be really helpful here.

lists/list_cache_test.go
ThinkChaos: I replied on the other comment too.

This test has an error occur on second parse, so it's not the case you mentioned (no entries and no error).  
I'll add a test for that case.

ThinkChaos: I replied on the other comment too.

This test has an error occur on second parse, so it's not the case you mentioned (no entries and no error).  
I'll add a test for that case.

ThinkChaos: Added a test with https://github.com/0xERR0R/blocky/pull/884/commits/08b17e2cd99f845bbb5231e9a293598f619e30d6

ThinkChaos:
Wrote tests to get 100% coverage on parsers and `util/arpa.go`.  
Will write a couple more to get missing coverage in the resolvers.

ThinkChaos:
Got test coverage to a point I'm happy with.  
For some reason, Codecov didn't update with the latest push and only shows results from [7d91165](https://github.com/0xERR0R/blocky/pull/884/commits/7d91165b1da16cf8bcc56e5300020d812bbf20a8).

I'll squash the commits and see if that unblocks it.

ThinkChaos:
Seems like CodeCov isn't working properly with rebase+force push.
You can see the coverage by manually going to the [last commit's tree coverage page](https://app.codecov.io/gh/0xERR0R/blocky/commit/9bca5d0f34ee6fef6bd463e45f59de6f720b3522/tree).

I think everything is ready on my side :)
Tests cover everything new 100%, except for `MockCallSequence` which is only used in tests. The hosts file resolver is also at 100% now.

0xERR0R:
CodeCov results are something "strange". I wanted to check it out, but as usual, no time. :)

I skimmed over your commit, it looks good. Is it now a stand what can be checked in? Or do you want to continue to develop? I think, threre was a TODO in your code (something with golint).

If you want to merge it, please squash your commits to one together.

0xERR0R (approved):

ThinkChaos:
Yeah the `nolint` will be solved by the follow-up.  
I'll try the "squash and merge" feature.

Thanks!

PR #902, opened 2023-02-27
feat: add upstream string to resolve error

### LINKED ISSUE #872: [Feature] Using actual endpoint string in the config to display warning/error, instead of all IP address
It's harder to confirm which endpoint is failing when they are all IP address, you'd need to find the reverseIP to located the issue

```
[2023-02-10 15:27:56]  WARN parallel resolver: test resolve of upstream server failed: can't resolve request via upstream server 92.249.39.1:853: x509: certificate has expired or is not yet valid: current time 2023-02-10T15:27:56Z is after 2022-08-23T14:34:49Z                                                                                                                                                                                                                        
[2023-02-10 15:27:56]  WARN parallel resolver: test resolve of upstream server failed: can't resolve request via upstream server https://92.249.39.1:443/dns-query: http return content type should be 'application/dns-message', but was ''
[2023-02-10 15:28:05]  WARN parallel resolver: test resolve of upstream server failed: can't resolve request via upstream server 185.228.168.9:853: EOF
```

ThinkChaos:
Fixes #872

### DIFF
--- a/resolver/upstream_resolver.go
+++ b/resolver/upstream_resolver.go
@@ -260,7 +260,7 @@ func (r *UpstreamResolver) Resolve(request *model.Request) (response *model.Resp
 				return nil
 			}
 
-			return fmt.Errorf("can't resolve request via upstream server %s: %w", upstreamURL, err)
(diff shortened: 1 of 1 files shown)


0xERR0R (approved):
👍

PR #920, opened 2023-03-10
refactor: `config.Duration`, and config use and printing

ThinkChaos:
I started these changes while working the hosts file resolver, and wanted to make a PR before I do the same to other resolvers, if there's interest.

The `config.Duration` change is to try and make using the type easier: with an embed all original methods are available.  
The tradeoffs are:
- defaults need to be parsed from a JSON string, which has to be escaped: ```struct { x Duration `default:"\"1h\""` }```
- converting to `time.Duration` is a bit less explicit: `x.Cast()` vs `time.Duration(x)`. That could be `x.Duration`, but I thought that was awkward to read (ex: `cfg.ConnectionCooldown.Duration`).

That first change allowed me to not copy the values out of the `config.HostsFileConfig` into the `HostsFileResolver` since using the `config.Duration` directly is easy.  
I think this is nice because it makes adding config options faster. And finding all references to an option/renaming it is instant. Depending on how we end up doing config reloading, it could help too: update a single field and trigger a refresh.

The config printing I moved to the config struct itself, instead of the resolver. And changed it to print via a logger instead of returning an array.  
I think that's valuable so we can make what's shown more or less verbose depending on the log level.

Let me know if what's of interest, if any, and I can adapt the PR.
The first change can be merged as is. The other two, I'd apply the idea to more resolvers.  
I also didn't take the time to fix the tests for the third change.

### DIFF
--- a/config/duration.go
+++ b/config/duration.go
@@ -23,6 +23,10 @@ func (c Duration) IsZero() bool {
 	return c.Duration == 0
 }
 
+func (c Duration) SecondsU32() uint32 {
+	return uint32(c.Seconds())
+}
+
 func (c Duration) String() string {
 	return durafmt.Parse(c.Cast()).String()
 }
--- a/config/duration.go
+++ b/config/duration.go
@@ -15,20 +15,20 @@ func NewDuration[T constraints.Integer | time.Duration](value T) Duration {
 	return Duration{Duration: time.Duration(value)}
 }
 
-func (c Duration) Cast() time.Duration {
+func (c Duration) ToDuration() time.Duration {
 	return c.Duration
 }
 
 func (c Duration) IsZero() bool {
-	return c.Duration == 0
+	return c.ToDuration() == 0
 }
 
 func (c Duration) SecondsU32() uint32 {
-	return uint32(c.Seconds())
+	return uint32(c.ToDuration().Seconds())
 }
 
 func (c Duration) String() string {
-	return durafmt.Parse(c.Cast()).String()
+	return durafmt.Parse(c.ToDuration()).String()
 }
 
 func (c *Duration) UnmarshalText(data []byte) error {
--- a/config/duration.go
+++ b/config/duration.go
@@ -1,0 +1,50 @@
+package config
+
+import (
+	"strconv"
+	"time"
+
+	"github.com/0xERR0R/blocky/log"
+	"github.com/hako/durafmt"
+	"golang.org/x/exp/constraints"
+)
+
+type Duration struct{ time.Duration }
+
+func NewDuration[T constraints.Integer | time.Duration](value T) Duration {
+	return Duration{Duration: time.Duration(value)}
+}
+
+func (c Duration) Cast() time.Duration {
+	return c.Duration
+}
+
+func (c Duration) IsZero() bool {
+	return c.Duration == 0
+}
+
+func (c Duration) String() string {
+	return durafmt.Parse(c.Cast()).String()
+}
+
+func (c *Duration) UnmarshalText(data []byte) error {
+	input := string(data)
+
+	if minutes, err := strconv.Atoi(input); err == nil {
+		// number without unit: use minutes to ensure back compatibility
+		*c = NewDuration(time.Duration(minutes) * time.Minute)
+
+		log.Log().Warnf("Setting a duration without a unit is deprecated. Please use '%s min' instead.", input)
+
+		return nil
+	}
+
+	duration, err := time.ParseDuration(input)
+	if err == nil {
+		*c = NewDuration(duration)
+
+		return nil
+	}
+
+	return err
+}
--- a/config/duration.go
+++ b/config/duration.go
@@ -6,37 +6,37 @@ import (
 
 	"github.com/0xERR0R/blocky/log"
 	"github.com/hako/durafmt"
-	"golang.org/x/exp/constraints"
 )
 
-type Duration struct{ time.Duration }
-
-func NewDuration[T constraints.Integer | time.Duration](value T) Duration {
-	return Duration{Duration: time.Duration(value)}
-}
+type Duration time.Duration
 
 func (c Duration) ToDuration() time.Duration {
-	return c.Duration
+	return time.Duration(c)
 }
 
 func (c Duration) IsZero() bool {
 	return c.ToDuration() == 0
 }
 
+func (c Duration) Seconds() float64 {
+	return c.ToDuration().Seconds()
+}
+
 func (c Duration) SecondsU32() uint32 {
-	return uint32(c.ToDuration().Seconds())
+	return uint32(c.Seconds())
 }
 
 func (c Duration) String() string {
 	return durafmt.Parse(c.ToDuration()).String()
 }
 
+// UnmarshalText implements `encoding.TextUnmarshaler`.
 func (c *Duration) UnmarshalText(data []byte) error {
 	input := string(data)
 
 	if minutes, err := strconv.Atoi(input); err == nil {
 		// number without unit: use minutes to ensure back compatibility
-		*c = NewDuration(time.Duration(minutes) * time.Minute)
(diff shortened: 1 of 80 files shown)


config/duration.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"strconv"
+	"time"
+
+	"github.com/0xERR0R/blocky/log"
+	"github.com/hako/durafmt"
+	"golang.org/x/exp/constraints"
+)
+
+type Duration struct{ time.Duration }
```
0xERR0R, line 12: I like the idea to move Duration to a new file and to provide some convenience methods/functions to make the code more readable 👍 

One think, I don't understand is, why did you change the Duration type to the embedded struct? I played a bit on your branch and changed the duration.go to following:
```go
package config

import (
	"strconv"
	"time"

	"github.com/0xERR0R/blocky/log"
	"github.com/hako/durafmt"
	"golang.org/x/exp/constraints"
)

type Duration time.Duration

func NewDuration[T constraints.Integer | time.Duration](value T) Duration {
	return Duration(time.Duration(value))
}

func (c Duration) Cast() time.Duration {
	return time.Duration(c)
}

func (c Duration) IsZero() bool {
	return c.Cast() == 0
}

func (c Duration) Seconds() float64 {
	return c.Cast().Seconds()
}

func (c Duration) SecondsU32() uint32 {
	return uint32(c.Cast().Seconds())
}

func (c Duration) String() string {
	return durafmt.Parse(c.Cast()).String()
}

func (c *Duration) UnmarshalText(data []byte) error {
	input := string(data)

	if minutes, err := strconv.Atoi(input); err == nil {
		// number without unit: use minutes to ensure back compatibility
		*c = NewDuration(time.Duration(minutes) * time.Minute)

		log.Log().Warnf("Setting a duration without a unit is deprecated. Please use '%s min' instead.", input)

		return nil
	}

	duration, err := time.ParseDuration(input)
	if err == nil {
		*c = NewDuration(duration)

		return nil
	}

	return err
}

```

so basically I reverted the type definition (Duration is alias for time.Duration), leaved your methods and added the "Seconds" function. With this change, we can use your improvements, but without changing the default tag to JSON. Or did I miss something?

config/duration.go
```diff
@@ -0,0 +1,54 @@
...
+
+func NewDuration[T constraints.Integer | time.Duration](value T) Duration {
+	return Duration{Duration: time.Duration(value)}
+}
+
+func (c Duration) Cast() time.Duration {
```
0xERR0R, line 18: 👍  `cfg.someProperty.Cast()` is definitively more readable as `time.Duration(cfg.someProperty)`

Suggestion: maybe we should rename "Cast" to "asDuration" or "toDuration"?

resolver/bootstrap.go
```diff
@@ -116,10 +116,10 @@ func (b *Bootstrap) resolveUpstream(r Resolver, host string) ([]net.IP, error) {
 		ctx := context.Background()
 
 		timeout := cfg.UpstreamTimeout
-		if timeout != 0 {
+		if timeout.IsZero() {
```
0xERR0R, line 119: 👍

server/server.go
```diff
@@ -443,8 +443,17 @@ func (s *Server) printConfiguration() {
 	for res != nil {
 		logger().Infof("-> resolver: '%s'", resolver.Name(res))
 
-		for _, c := range res.Configuration() {
-			logger().Infof("     %s", c)
+		if resCfg, ok := res.(resolver.ConfigGetter); ok {
```
0xERR0R, line 446: I think the idea here is to replace "Configuration" with new "LogValues" method in all resolvers?

server/server.go
```diff
@@ -443,8 +443,17 @@ func (s *Server) printConfiguration() {
...
+			func() {
+				undo := log.PrefixMessages("     ", logger().Logger)
+				defer undo()
```
0xERR0R, line 449: Is it possible to create a new logger instance with prefix formatter? I mean a new instance only for configuration logging. I think, it is a bit "dangerous" to modify the existing instance and to rollback this change after usage. Later someone can use this logic somewhere else and forget to call the undo function.

resolver/custom_dns_resolver.go
```diff
@@ -38,7 +37,7 @@ func NewCustomDNSResolver(cfg config.CustomDNSConfig) ChainedResolver {
 		}
 	}
 
-	ttl := uint32(time.Duration(cfg.CustomTTL).Seconds())
+	ttl := uint32(cfg.CustomTTL.Seconds())
```
0xERR0R, line 40: can use new "SecondsU32" func

log/logger.go
```diff
@@ -118,3 +120,42 @@ func ConfigureLogger(cfg *Config) {
 func Silence() {
 	logger.Out = io.Discard
 }
+
+// PrefixMessages modifies a logger entry which will add `prefix` to all messages.
+//
+// The returned function must be called to remove the prefix.
+func PrefixMessages(prefix string, logger *logrus.Logger) func() {
```
0xERR0R, line 127: see my comment in server.go

config/duration.go
kwitsch: I had the same question while reading it.

config/duration.go
kwitsch: Agreed `ToDuration` would be more readable as the name implies the result.

server/server.go
ThinkChaos: Yeah exactly, just did it to a single resolver as a proof of concept for now.  
The `ConfigGetter` interface would not exist, and we'd have the `Cfg` method directly in `Resolver` instead of the current `Configuration`.  
I could also rename `Cfg` to `Configuration` since that name would be free.

config/duration.go
ThinkChaos: Yeah that's a good middle ground!

I didn't do it like that to avoid needing to wrap every method of `time.Duration`, but we don't use that many so this seems better :)

server/server.go
ThinkChaos: Yeah I can do that too, this was just a bit easier with the existing `log` code!

ThinkChaos:
Thanks for taking a look so fast, I'll make the changes you proposed and do more resolvers!

server/server.go
ThinkChaos: So I tried changing this, but using a new logger loses stuff like fields that were added.
I changed the API to be impossible to misuse though, so hopefully that solves the issue with another approach!

config/duration.go
ThinkChaos: Went with `ToDuration` :)

config/duration.go
ThinkChaos: Used the code you posted as a starting point for the continuation. Thanks!

ThinkChaos (commented):
Updated all resolvers, their configs and all tests!

server/server.go
ThinkChaos: I removed `ConfigGetter` and just went with `config.Configurable` (which is the new name for `ValueLogger`).
With that rename, I added `IsEnabled() bool` to the interface.

`Cfg()` was replaced by `LogConfig(*logrus.Entry)`. That allows having a single interface that covers both the config types, and the resolvers.

ThinkChaos:
Pushed some more commits, main change there is adding `Type() string` to `Resolver`. It's used as the log prefix, and when printing the config at startup.

I also tweaked the startup config so it's nicer (that's obviously subjective, so let me know if there's anything you'd like to improve).  
Here's a comparison, left is before, right after:
image

kwitsch:
> I also tweaked the startup config so it's nicer (that's obviously subjective, so let me know if there's anything you'd like to improve).  

I'm in favor of the new structure. 👍

kwitsch (approved):
Enhances readability & streamlines functionality 👍

0xERR0R (approved):
Great Job! Looks very clean :rocket: :+1:

Issue #541, opened 2022-05-25
[Feature Request] Add API endpoint to flush the DNS Cache


skycubeuk:
As a user I would like to be able to flush the dns cache programmatically

When deploying CI/CD pipelines dns records and TTLs can update multiple times in a short period. Being able to clear the DNS cache via an API call would be helpful.

ashmckenzie:
I'm working on a PR for this 🙂

PR #940, opened 2023-03-18
Adding a switch on/off for active disable OpenDNS by ip groups

Adi3000:
Adding two endpoint : 
```
GET /api/dns/disable?groups=<client-group-id>&duration=120m
GET /api/dns/enable
GET /api/dns/status
```
to switch on or off specific DNS applied for a client group id
A way to activate desactivate parental control or adguard temporary

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -38,6 +38,10 @@ type QueryRequest struct {
 	Query string `json:"query"`
 	// request type (A, AAAA, ...)
 	Type string `json:"type"`
+	// optional : Use remote address : default false
+	UseRemoteAddress bool
+	// optional : Remote address override
+	RemoteAddress string
 }
 
 // QueryResult is a data structure for the DNS result
--- a/api/api.go
+++ b/api/api.go
@@ -22,6 +22,15 @@ const (
 	// PathBlockingDisablePath defines the REST endpoint for blocking disable
 	PathBlockingDisablePath = "/api/blocking/disable"
 
+	// PathClientDNSResolverStatusPath defines the REST endpoint for client dns resolver status
+	PathClientDNSResolverStatusPath = "/api/dns/status"
+
+	// PathClientDNSResolverEnablePath defines the REST endpoint for client dns resolver enable
+	PathClientDNSResolverEnablePath = "/api/dns/enable"
+
+	// PathClientDNSResolverDisablePath defines the REST endpoint for client dns resolver disable
+	PathClientDNSResolverDisablePath = "/api/dns/disable"
+
 	// PathListsRefresh defines the REST endpoint for blocking refresh
 	PathListsRefresh = "/api/lists/refresh"
 
@@ -39,9 +48,11 @@ type QueryRequest struct {
 	// request type (A, AAAA, ...)
 	Type string `json:"type"`
 	// optional : Use remote address : default false
-	UseRemoteAddress bool
+	UseRemoteAddress bool `json:"useRemoteAddress"`
 	// optional : Remote address override
-	RemoteAddress string
+	RemoteAddress string `json:"remoteAddress"`
+	// optional : refresh cache : default false
+	RefreshCache bool `json:"refreshCache"`
 }
 
 // QueryResult is a data structure for the DNS result
--- a/api/api_endpoints.go
+++ b/api/api_endpoints.go
@@ -24,6 +24,13 @@ type BlockingControl interface {
 	BlockingStatus() BlockingStatus
 }
 
+// ClientDNSResolverControl interface to control the blocking status
+type ClientDNSResolverControl interface {
+	EnableClientDNSResolver()
+	DisableClientDNSResolver(duration time.Duration, disableGroups []string) error
+	ClientDNSResolverStatus() BlockingStatus
+}
+
 // ListRefresher interface to control the list refresh
 type ListRefresher interface {
 	RefreshLists()
@@ -32,6 +39,11 @@ type ListRefresher interface {
 // BlockingEndpoint endpoint for the blocking status control
 type BlockingEndpoint struct {
 	control BlockingControl
+}
+
+// BlockingEndpoint endpoint for the blocking status control
+type ClientDNSResolverEndpoint struct {
+	control ClientDNSResolverControl
 }
 
 // ListRefreshEndpoint endpoint for list refresh
@@ -43,6 +55,10 @@ type ListRefreshEndpoint struct {
 func RegisterEndpoint(router chi.Router, t interface{}) {
 	if a, ok := t.(BlockingControl); ok {
 		registerBlockingEndpoints(router, a)
+	}
+
+	if a, ok := t.(ClientDNSResolverControl); ok {
+		registerClientDNSResolverEndpoints(router, a)
 	}
 
 	if a, ok := t.(ListRefresher); ok {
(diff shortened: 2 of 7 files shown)


0xERR0R:
Thank you for you contribution. I skimmed over you changes and I don't quite understand what you want to achieve. Do you want to implement some kind of "parential control" and you want to be able to complete disable the DNS resolution for a particular group of clients (e.g. kids devices)? So basically, you want to call a REST endpoint and all DNS queries of your kids should be answered with "blocked" response, right?

Adi3000:
OpenDNS is a DNS which provide some parental control (malicious site, streaming, pornographic content...)
As a parent, I offer some controlled freedom to my children. But when it come to me to use this computer, I should be able to fetch some of those site for a short lap time. So I switch from OpenDNS to Cloudflare, or Google DNS.

Here a sample of my configuration : 
```yaml
upstream:
  default:
  - 8.8.8.8
  - 8.8.4.4
  - 1.1.1.1
  - 1.0.0.1

  adi-home:
  - 208.67.222.222 #OpenDns 1
  - 208.67.220.220 #OpenDns 2
...
clientLookup:
  clients:
    adi-home:
    - 123.231.123.12 # fake ip
```
So, when I use `GET /api/dns/disable?groups=adi-home&duration=30m` , it switch from OpenDns to Cloudflare during 30 minutes.

When it is resolved by DNS status is still `RESOLVED`
```
query resolved answer= client_ip=123.231.123.12 client_names=adi-home duration_ms=9 hostname=e98b21836ff2 question_name=naugthy-malicious-unfriendly-site.com. question_type=HTTPS response_code=NOERROR response_reason=RESOLVED (tcp+udp:208.67.222.222) response_type=RESOLVED
```
but it is either blocked by OpenDNS by default, or allowed and resolved by Cloudflare or Google when client group is "dns disabled"

I may have fork from the initial behaviour. But with your base code and some tricks around the actual configuration I made it doable. Also I am working on this since 3 or 4 month (not all time, but when I find time) so this is my initial approch if this kind of feature may fit some of your need.
I also know that it is some big deviation and some big misleading of meanful feature (disable a DNS is a non sense) but, your project was one of the more likely to do so (by configuring group of IP, DoH or other rule to affect a specific DNS or batch of rules). 
Please feel free to ask me to go further, or stale my PR, even reject it if it doesn't fit your roadmap.
Also, beware, it is my first Golang project, there may be some ugly part...

0xERR0R:
ok, you want that in a default case, some clients in your Lan use a blocking upstream DNS and you want to disable this behavior temporarily (no blocking for them). I think, the implementation will work, but it looks for me like it is an edge case and you are trying to archive something on a more difficult way:

Did you try to define a blocking list for your kids (ads, adult pages etc.) in blocky (we have some links for blacklists as idea [here](https://0xerr0r.github.io/blocky/v0.20/additional_information/#blacklists)) and use blocky as a content blocker for your clients? In this case you would have under controll which content should be blocked and you can disable the blocking via existing API?

Adi3000:
In my case, I have already an account at OpenDNS. There is some feature I like on it that I can't provide by myself mostly including the banner which says that site is blocked and why it's blocked.

I can use indeed one of the list online (I already do for ads blocking), and whitelist some of them manually (allowing them on some website for support), BUT, the frame coming from OpenDNS.
image

The good thing on it is that I do not need to maintain categorization or adding/removing list.

In this configuration, I only have to maintain the ad blockers lists and my proper parental control (I added some blocking rules on Tiktok and Youtube, to be used only with my supervision)

Here is a summup of my need : 
* Effortless unappropriate children content blocking (no need to check if there is some new better list somewhere else, OpenDNS fullfil this need, but if you think it can be done here, I'll try)
* Explicitly block some of site to make my children understand that they go to deep in the web, I have my proper 403 web page for Youtube and Tiktok, they understand, and blocking ads is transparent in a site web (it only left some blank part on pages)
* Using opensource list (ad blocker, GAFAM blocker, pgl.yoyo.org ...) to control my browser
* Adding specific rules for some devices (DoH with client group parameter is nice for that)
* Usable in multiple house (my whole large family is using your project to avoid at least tiktok, and feels more secure now they are limited on some unappropriate content) : using client group and enable disable blocking list for some client group should be great. For now I copy paste blocking list for each group, but disabled status and timer is set on all disabled rules (not really important, it was my step brother advice).

The mix with OpenDNS and its feature is the main issue where I think I'll loose some of my comfort (it has some feature with highlight / report and notification when some unappropriate website has been asked)

ThinkChaos:
If the main feature you want of this is a webpage with the reason for the block, IMO it could be done directly in blocky with a new block type (you'd also need to ensure that the expected traffic arrives on blocky's interface).  
That seems like a more interesting feature for this project to me since it doesn't introduce a new interaction with something external.  
Of course it's up to @0xERR0R to decide if they think it's a good idea and fit for the project!

0xERR0R:
> If the main feature you want of this is a webpage with the reason for the block, IMO it could be done directly in blocky with a new block type (you'd also need to ensure that the expected traffic arrives on blocky's interface). That seems like a more interesting feature for this project to me since it doesn't introduce a new interaction with something external. Of course it's up to @0xERR0R to decide if they think it's a good idea and fit for the project!

This is a great idea to provide a blocking page with reason text why the domain was blocked. It would be an extension to current blockType "IP". But... It doesn't work out of the box:

Today, most of the internet sites a using HTTPS. And if you browser requests https://ads.domain and blocky will "redirect" the page to a custom HTTP server, the user will see warning in the browser and not this site. You can either create own CA and import the root cert on each device (and create in the blockpage server a certificate on-the-fly for each domain) or wait for proper support of EDNS responses in the browser:

- https://datatracker.ietf.org/doc/html/rfc8914
- https://datatracker.ietf.org/doc/html/draft-reddy-dnsop-error-page-07

AFAIK, cloudflare's blockpage works only if you import cloudflare's root cert (https://developers.cloudflare.com/cloudflare-one/policies/filtering/configuring-block-page/)

Please correct me if my information is outdated and there is a better way to implement this feature

Adi3000:
Yeah this didn't work well on https redirection. But at least works on some http domains. I don't know how but, Opendns manage to display it's block page in many cases

You're righe about ssl : debugging curl shows that the first thing that will be done inn https transaction is to download and control certificate before control any other header, query parameter.

One of my tries was to script a wildcard self signed CA based on blacklist imported. But then I had to repush it on every devices indeed. And I'm not at work, I do not manage computer with remote/ssh control. Within chrome, it allows me to have a less restrained blocked pop-up witch allows at least to to click on "Continue anyway"

Anyway, for now my blocking page with blocky is defined with the `blockType` instruction.

Adi3000:
Nevertheless, @0xERR0R, should I try to create some issues and drop this PR for now? Then we'll may discuss of how we can answer it a proper way?

I'll not end up mad if we do so. I have forked a nearly working solution on my side. And my goal is only to get more confident in GoLang with a purpose :)

0xERR0R:
> Nevertheless, @0xERR0R, should I try to create some issues and drop this PR for now? Then we'll may discuss of how we can answer it a proper way?
> 
> I'll not end up mad if we do so. I have forked a nearly working solution on my side. And my goal is only to get more confident in GoLang with a purpose :)

Yes, I think we should split your requirements to different tasks and discuss them separately.

For one part of your request (reset cache) we do already have an issue: https://github.com/0xERR0R/blocky/issues/541

Adi3000:
Okay, for this part (#541) , I'll make another PR with this feature (not today, but I hope very soon). For the rest, I'll create the issue, thank for this chat, see you in another issue ;)

ThinkChaos:
For future reference:
AFAICT OpenDNS does have its own Root CA to make block pages work with TLS.  
See this support page explaining how to install it: https://support.opendns.com/hc/en-us/articles/227987007-Block-Page-Errors-Installing-the-Cisco-Umbrella-Root-CA

That also means they most likely don't have any special magic, and any pages showing the block without that CA were HTTP pages.

0xERR0R:
> For future reference: AFAICT OpenDNS does have its own Root CA to make block pages work with TLS. See this support page explaining how to install it: https://support.opendns.com/hc/en-us/articles/227987007-Block-Page-Errors-Installing-the-Cisco-Umbrella-Root-CA
> 
> That also means they most likely don't have any special magic, and any pages showing the block without that CA were HTTP pages.

Yes, no magic. But the Root cert needs to be installed on each device. This is very annoying and is associated with unpleasant side effects, such as security warnings on Android. :(

ThinkChaos:
Yeah agree it's not a solution for most users, just meant for @Adi3000 it seemed to already be enough to just support HTTP :)

PR #963, opened 2023-03-28
fix: support IDNA in lists

### LINKED ISSUE #961: Support international domain names in lists (IDNA)
See https://github.com/0xERR0R/blocky/discussions/960

Example list: `https://raw.githubusercontent.com/rahilpathan/pihole-whitelist/main/1.LowWL.txt`  
Fix idea: use the idna module's [ToASCII](https://pkg.go.dev/golang.org/x/net/idna#Profile.ToASCII) to convert those domains to their punycode representation which is what blocky should receive from clients.

ThinkChaos:
Not sure this is actually supported by most adblocker, but we might as well be permissive.

Fixes #961.

### DIFF
--- a/lists/parsers/hosts.go
+++ b/lists/parsers/hosts.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/asaskevich/govalidator"
 	"github.com/hashicorp/go-multierror"
+	"golang.org/x/net/idna"
 )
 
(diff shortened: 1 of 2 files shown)


0xERR0R (approved):
LGTM 👍

kwitsch (approved):

PR #974, opened 2023-04-03
chore(build): add format with gofumpt to build

0xERR0R:
make build and lint should automatically format go files with gofumpt


kwitsch (approved):
usefull change 👍

Issue #966, opened 2023-03-29
Improve list parsing error handling


0xERR0R:
Currently, if a single list file (on inline definition) contains more than 5 errors ([list_cache.go#maxErrorsPerFile](https://github.com/0xERR0R/blocky/blob/main/lists/list_cache.go#L26)), the parser will stop the import process. It means, the file is partially in cache (all entries parsed until the 5th error is occurred).

I think the idea was to stop processing if the input file contains only "garbage" (for example user references HTML page instead of a plain text file).

If user imports big files (> 1M entries), 5 errors as threshold can be reached very quickly. In this case, only a part of the file is in the cache and user may not notice it.

I think we can improve it, some ideas here:

- Threshold value should be dependent on the file line count (maybe as percentage? 5% of row count, but min. 5?
- Log all errors for a single list together as summary like: List XXX import finished in XX ms, XXX rows imported, XXX rows ignored, erros: (XXX list of errors). Currently we are logging each error as a single log entry
- New Prometheus metrics for ignored/malformed import list rows. User can define alerts/monitor it

MaKeG0:
Hello,

I've encountered a similar issue due to the current parsing error limitation in Blocky. I'm running Blocky in a container within my router, following the setup guide provided here: https://xaizone.eu/post/setup-blocky-on-mikrotik-routeros/

The setup works well, and I've successfully integrated blocklists from the following sources:
- https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
- https://raw.githubusercontent.com/blocklistproject/Lists/master/ads.txt
- https://raw.githubusercontent.com/blocklistproject/Lists/master/tracking.txt
- https://malware-filter.gitlab.io/malware-filter/urlhaus-filter-domains.txt
- https://www.snort.org/downloads/ip-block-list

However, I've run into problems when trying to incorporate lists from these sources:
- https://easylist.to/easylist/easylist.txt
- https://pgl.yoyo.org/adservers/serverlist.php?hostformat=hosts;showintro=0

These lists were found on the uBlock project page: https://github.com/gorhill/uBlock

The issue seems to stem from the parsing error limit in Blocky. If the maximum number of allowed parsing errors could be increased, or even better, made configurable, it would likely resolve this issue. 

Additionally, it would be beneficial to have a default parsing mechanism that skips HTML tags. This would be particularly useful for lists generated by PHP pages, which often include HTML tags that contribute to parsing errors.

[...]

PR #986, opened 2023-04-10
Add maxErrorsPerFile blocking configuration

mandrav:
The default max errors per file of 5 is too small IMHO. This commit makes this number user-configurable.


kwitsch:
Looks good to me. 👍

I'll approve it after the linter errors are fixed.

ThinkChaos:
Thanks for the PR!

I pushed a couple commits to fix the lint error, add docs and a test, and allow using -1 as the value for "no limit".

FYI we also want to support percentage and switch the default to that, see #966.  
I'll try to get that done before the next release so we can that in the config from the start. That'll involve making the config a string, so maybe `-1` can be replaced with literally writing "no-limit".

kwitsch (approved):
- linter ok 👍
- documentation enhanced 👍
- unit test coverage 👍

The change to disable the error limit sounds reasonable since it's already an internal constant.

mandrav:
Awesome, thanks!

Issue #945, opened 2023-03-21
Redis rework


0xERR0R:
- [ ] Implementation of GroupedCache to enable read access with multiple groups. (#944)
- [ ] Redis implementation of GroupedCache (using [rueidis](https://github.com/rueian/rueidis))
- [ ] Redis implementation of ExpirationCache (DNS response caching) (using [rueidis](https://github.com/rueian/rueidis))
- [ ] Rework Prefetching (Redis & in-memory implementation)
- [ ] if Redis enabled: execution of Jobs (e.g. List refresh, periodic metrics with lists count, prefetching, ...?) only on one instance, (using Redis lock with https://github.com/rueian/rueidis/tree/main/rueidislock)

kwitsch:
I picked this one up again since rueidis is stable now and I hadn't worked with it in a while.

I'm stuck with the unit tests(again). 😞 
Most of the tests need to be rewritten since minredis don't support some features which rueidis neets. 

I can't get my head around how to solve the requirements of the test setups with rueidismock. 

I'm currently considering simply spinning up a test container with redis running alongside the tests...

0xERR0R:
If we can cover all use cases with e2e tests, why not? Unit test can be helpful to test some edge cases, but if it needs much efford to mock the infrastructure, it is better to throw them away 😄

kwitsch:
I'll open a PR draft without the tests soon.
Maybe I'm only missing something and it isn't that hard.


more eyes & more thoughts may help 😅

Serverfrog:
can this ticket also consider some implementation of #694 ?
As im currently running a kind of GitOps where blocky is also is configured, i don't really want to go my way out to somehow put a Secret i have into a Configmap.
For me the Redis Password is coming from a SecretProvider so its kinda hard to put a singular value from these into a Configmap, and just putting the value itself in there defies the usage of the SecretProvider for it

0xERR0R:
In your case, you have the password in k8s secret and you can mount it to env variable (for example) and it would be sufficient if blocky can read the password from the env variable?

Serverfrog:
For me personally at my current install it would be not so much a problem, but also consider https://github.com/0xERR0R/blocky/issues/694#issuecomment-1420920276
TL;DR Env Variables are kinda public, in usage of Docker (so in my instance not the case) every user on the machine can read them.

PR #994, opened 2023-04-14
redis rework base

kwitsch:
- rewrote old redis logic with rueidis
- removed almost all redis unit tests(they rely on minredis which isn't compatible with rueidis)

As most of the logic will change during implementation of #945 unit tests should be added afterwards.


ThinkChaos:
Looks like a lot of the changes are from auto formatting. Did you use `gofumpt`?  
Running `gofumpt -w ./**.go` doesn't make any changes for me.  
I think that would be nice as a separate MR/commit.

I did take a quick look at the Redis changes, and only thing that stuck out to me was storing the context in `Client`. I think it's best practice to avoid that and have a context argument for every method.

kwitsch:
> I did take a quick look at the Redis changes, and only thing that stuck out to me was storing the context in `Client`. I think it's best practice to avoid that and have a context argument for every method.

That was intended to be used later on. Had the idea to use a cancellation context which is stored inside the client and is used for all requests. Then we can implement a client close method that cancels the context so all requests are terminated gracefully.

Sorry I'm not that familiar with context best practices. 😓

Is it better to use context.Background() at every call instead of doing it one time and storing it?

kwitsch:
> Looks like a lot of the changes are from auto formatting. Did you use `gofumpt`?  
> Running `gofumpt -w ./**.go` doesn't make any changes for me.  

I used `make fmt` because `make lint` wouldn't run.
Wasn't that the correct fix?
Often did that in the past.
Do we need to modify the Makefile?

ThinkChaos:
> Wasn't that the correct fix?

Yeah that's what I'd do too.  
What I don't understand is when I run `make fmt` on main (ea95d36f37a80c59e72c6de15bf04b1651ca6c16) I don't get any of those changes.

> That was intended to be used later on. Had the idea to use a cancellation context which is stored inside the client and is used for all requests. Then we can implement a client close method that cancels the context so all requests are terminated gracefully.

In that case I think we can use [rueidis' Client.Close](https://pkg.go.dev/github.com/rueian/rueidis#Client.Close).  
But yeah the recommended use of a context is for the caller to pass it with each call/action so that each one can have an independent timeout/cancellation.  
See [Contexts and structs](https://go.dev/blog/context-and-structs) for more details.

kwitsch:
Thanks for the info.
I'll change it.

Regarding the fmt problem: I'll try to reproduce it to see what went wrong

kwitsch:
> Regarding the fmt problem: I'll try to reproduce it to see what went wrong

I currently can't reproduce it and also don't remember anything else I did at this commit. 🫤


I'll close this PR and implement the context changes on a new branch.

0xERR0R:
I though about Redis integration, I think it would make sense to create a new branch fb-redis-integration for long running redis implementation tasks. And put all related PRs on this branch. We would have in this case main without CI buld errors and we can use this branch to integrate and test redis. Afterwards we can merge it into master. @kwitsch what do you think?

kwitsch:
> I though about Redis integration, I think it would make sense to create a new branch fb-redis-integration for long running redis implementation tasks. And put all related PRs on this branch. We would have in this case main without CI buld errors and we can use this branch to integrate and test redis. Afterwards we can merge it into master. @kwitsch what do you think?

Sounds good.
@0xERR0R if you create the branch on the main repository I'll request future PRs to it.

0xERR0R:
Branch was created

Issue #465, opened 2022-03-21
CustomDNS should support other query types (CNAME, TXT, ...)


0xERR0R:

SuperSandro2000:
I added cname support in https://github.com/0xERR0R/blocky/pull/996

BenMcH:
While the CNAME PR appears to be DOA, you can emulate CNAMEs for internal IPs using yaml anchor tags like this:

```yaml
customDNS:
  customTTL: 1h
  mapping:
    home-server.lan: &server 192.168.0.1
    app-1.lan: *server
    app-2.lan: *server
```

This of course has its downsides and doesn't solve for external domain names, but for my use-cases this works well enough. 

If I have time in the coming weeks/months, I may look at picking up where @SuperSandro2000 left off and add @0xERR0R's suggestion for psuedo-funcitons in the config file to see if we can get that merged. I know I'd prefer my apps to be CNAMEs even if this is technically functional.

joryirving:
I saw that @BenMcH merged #1352 and I'm very thankful, but I wanted to know if there was a way to handle multiple zones in 0.24? I got one zone working wonderfully.

kwitsch:
Shouldn't be a problem since according to [RFC1035](https://datatracker.ietf.org/doc/html/rfc1035) the master file format supports it. The commonly used name 'zone file' is a bit misleading since it supports more than just one zone.

PR #996, opened 2023-04-16
customDns: add cname support

SuperSandro2000:
related #465


jkossis:
Is there an ETA on this by any chance?

0xERR0R:
Thanks for your contribution and sorry for delay, I had a lot to do at work and little time for private projects.

I tried your change and it works: for each rewrite entry a CNAME DNS response will be returned. I testet also with following config (from documentation):


 ```yaml
 customDNS:
   customTTL: 1h
   filterUnmappedTypes: true
   rewrite:
     home: lan
     replace-me.com: with-this.com
   mapping:
     printer.lan: 192.168.178.3
     otherdevice.lan: 192.168.178.15,2001:0db8:85a3:08d3:1319:8a2e:0370:7344
 ```

DNS questions for "printer.lan" or "printer.home" return A 192.168.178.3, because the "home" part will be rewritten to "lan". But I'll also receive CNAME record for "home". It is desired? I think, using of "rewrite" section for query rewrite and CNAME records is a bit confusing, just IMHO.

Maybe we should do it more explicitly, for example:
 ```yaml
 customDNS:
   ...
   mapping:
     printer.lan: A(192.168.178.3)
     some.domain: CNAME(example.com)
     other.domain: TXT(blabla)
 ```

With support for A/AAAA (default), CNAME, TXT, MX, ...?

SuperSandro2000:
I am not planning to work more on this currently.

PR #997, opened 2023-04-16
Add suoport to read dns servers from resolv.conf

SuperSandro2000:
I wanted blocky to read my nameserver entries from /etc/resolv.conf bzw. systemd-resolved.

### DIFF
--- a/config/resolvconf.go
+++ b/config/resolvconf.go
@@ -1,0 +1,48 @@
+package config
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/moby/moby/libnetwork/resolvconf"
+)
+
+type ResolvconfConfig struct {
+	Enable bool `yaml:"enable"`
+}
+
+func getDNSServerFromResolvConf() ([]Upstream, error) {
+	resolvconfFile, err := os.ReadFile(resolvconf.Path())
+	if err != nil {
+		return nil, fmt.Errorf("can't read resolv.conf file: %w", err)
+	}
+
+	resolvconfServers := resolvconf.GetNameservers(resolvconfFile, resolvconf.IP)
+	upstreams := make([]Upstream, 0, len(resolvconfServers))
+
+	for _, server := range resolvconfServers {
+		upstream, err := ParseUpstream(server)
+		if err != nil {
+			return nil, err
+		}
+
+		upstreams = append(upstreams, upstream)
+	}
+
+	return upstreams, nil
+}
+
+func (c *ParallelBestConfig) AddResolvconfToConfig() error {
+	upstreams, err := getDNSServerFromResolvConf()
+	if err != nil {
+		return err
+	}
+
+	if c.ExternalResolvers == nil {
+		c.ExternalResolvers = make(ParallelBestMapping)
+	}
+
+	c.ExternalResolvers["default"] = append(c.ExternalResolvers["default"], upstreams...)
+
+	return nil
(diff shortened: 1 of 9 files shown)


server/server.go
```diff
@@ -397,13 +397,15 @@ func createQueryResolver(
 	redisClient *redis.Client,
 ) (r resolver.Resolver, err error) {
 	blocking, blErr := resolver.NewBlockingResolver(cfg.Blocking, redisClient, bootstrap)
+	rErr := cfg.Upstream.AddResolvconfToConfig()
```
ThinkChaos, line 400: I don't see the implementation for this function anywhere. Did you forget to stage it?

ThinkChaos (commented):
Thanks for the MR!

Reminds me I should contribute my changes to the blocky NixOS module :smile:

server/server.go
SuperSandro2000: Upsi

config/resolvconf.go
```diff
@@ -0,0 +1,48 @@
...
+	upstreams, err := getDNSServerFromResolvConf()
+	if err != nil {
+		return err
+	}
+
+	if c.ExternalResolvers == nil {
+		c.ExternalResolvers = make(ParallelBestMapping)
+	}
+
+	c.ExternalResolvers["default"] = append(c.ExternalResolvers["default"], upstreams...)
```
ThinkChaos, line 45: There's two things I think we should change:
   - avoid limiting this to the default group
   - have the option in `ParallelBestConfig` directly, since that's where it applies (and to go with the above, where groups are defined)

I understand that it's just a `map[string][]Upstream` right now which makes that more difficult, and is probably why you made a global bool.

Here's a two ideas I could come up with:

1. Keep the `map`, but change `Upstream` to a new type that is "either an upstream or a special source"
```yaml
upstream:
  my-group:
    - !from resolv.conf # the syntax could be something else
```
  (proposed syntax uses a YAML tag which is supported since Go YAML v3)

2. Give the resolver its own section:
```yaml
parallelBest:
  upstreams:
    my-group:
      - example.com
  resolvconf: # list of groups
    - my-group
```

I like option 2 less TBH, but adding a `parallelBest` section at some point would be nice since that's the only resolver without its own section.  
But I also think we should split parallel best since it's actually 2 resolvers in a trench coat: "client groups" and "parallel". So maybe a section should wait until it's split to avoid too much config churn. (made #1001 for that split)

Do others have opinions on any of this?  
I'm not opposed to merging something simple like `resolvconf: ["default", "my-group"]` and building on it after. Using a YAML tag seems be interesting to me since I think those could be very useful for #476.

SuperSandro2000: > avoid limiting this to the default group

I didn't know that the group names actually matter, I thought they where just for organization. 

---
I don't really have strong opinions on the organization of the settings but I have limited time changing that potentially multiple times.

ThinkChaos: Sorry thought I had replied again.

Yeah the groups matter if you use [client groups](https://0xerr0r.github.io/blocky/v0.21/configuration/#client-groups).

I'd be fine with this MR with just an extra option for the group to use, a description in `docs/configuration.md` and an example in `docs/config.yml`.  
We can build on this resolver in the future.

0xERR0R:
@ThinkChaos: What is the current status of this PR?

PR #998, opened 2023-04-18
feat: support multiple hosts files

### LINKED ISSUE #867: [Feature Request] Option to get hosts file from http url

ThinkChaos:
Fixes #867


go.mod
```diff
@@ -37,6 +37,7 @@ require (
 
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.0
+	github.com/ThinkChaos/parcour v0.0.0-20230418015731-5c82efbe68f5
```
ThinkChaos, line 40: This is a module I wrote with the goal of making concurrency patterns reusable and hard to misuse.
The module is not really complete, so I just did `go get` with a specific commit.

It is fully documented, so hopefully that can help make things clear. I'll also post a comment to explain the basics in the hosts file resolver.

resolver/hosts_file_resolver.go
```diff
@@ -144,27 +151,76 @@ func (r *HostsFileResolver) resolve(req *dns.Msg, question dns.Question, domain
...
-	f, err := os.Open(r.cfg.Filepath)
+	//nolint:ineffassign,staticcheck // keep `ctx :=` so if we use ctx in the future, we use the correct one
+	consumersGrp, ctx := jobgroup.WithContext(ctx)
+	defer consumersGrp.Close()
+
+	producersGrp := jobgroup.WithMaxConcurrency(consumersGrp, r.cfg.Loading.Concurrency)
+	defer producersGrp.Close()
+
+	producers := parcour.NewProducersWithBuffer[*HostsFileEntry](producersGrp, consumersGrp, producersBuffCap)
+	defer producers.Close()
```
ThinkChaos, line 167: Basically a `JobGroup` is a scope for goroutines: the `defer grp.Close()` ensures no goroutine continues running after the function returns. Groups can be nested: here `producersGrp` is a child of `consumersGrp`.
These are ideas from [structured concurrency](https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/) (highly recommend that post, it's one of the most influential CS articles I've read).

Having a clear scope for goroutines also allows for clear failure propagation:
- when explicitly checking for goroutine errors with `Wait`:
  - if any goroutines returned errors, it returns them (just like `errgroup.Group`)
  - if any goroutines panicked, it propagates the panic to the current goroutine
- when we leave the function without explicitly waiting for goroutine, `Close` handles the failures:
  - if any goroutines returned errors, it propagates them to the parent `JobGroup` (or panics if there is no parent)
  - if any goroutines panicked, it propagates the panic to the current goroutine

`producers` is an abstraction based on `JobGroup`s to make the pattern we have here of producing host file entries from sources in parallel and consuming them from a single goroutine simple.

The fact that `producers` uses the `JobGroup`s we created means we know for sure it won't leave anything running since we're closing them before returning, and we can customize how those goroutines run.

lists/list_cache.go
```diff
@@ -117,152 +109,138 @@ func logger() *logrus.Entry {
...
+}
 
-	// guard channel is used to limit the number of concurrent executions of the function
-	guard := make(chan struct{}, b.processingConcurrency)
+func (b *ListCache) refresh() error {
+	unlimitedGrp, _ := jobgroup.WithContext(context.Background())
+	defer unlimitedGrp.Close()
 
-	processingLinkJobs := len(links)
+	producersGrp := jobgroup.WithMaxConcurrency(unlimitedGrp, b.processingConcurrency)
```
ThinkChaos, line 126: This limit applies to the whitelist and blacklist independently.  
With the new code it should be easy to make both share the same limit. That seems like more user friendly behavior to me since there's a single option in the config. Do you think I should make that change?

go.mod
0xERR0R: I like the idea to use some "high-level" library for concurrency to reduce some boiler plate code. How does your lib differ from https://github.com/sourcegraph/conc or similar libraries? I'm not against your implementation, it is only the question about support/maturity/features etc

ThinkChaos: I didn't find that module when I searched for existing implementations. I searched using the godoc website and GitHub, I guess I should've used a search engine too.  

A couple differences I noticed:
- conc's pools are all different types with slightly different APIs and aren't composable.  
  So no single interface can cover them all, meaning each function has to choose a flavor. Abstractions using a pool must choose for the users which type to use. See `Stream`: can't use it if you need to return errors.
  It also means that they have to implement each combination, which is not maintainable. See: `ContextPool`, `ResultContextPool`, `ResultPool`, etc. If you want to both use a context, and return an error, you can't without writing a new pool.
- conc's pools don't have a parent/child relationship, so compared to parcour, you loose the automatic error propagation to parents
- conc's pools are not well integrated with contexts
- conc's pools are not safe for concurrent use.  
  Given that it's based on `sync.WaitGroup`, just like parcour, it would be nice to be able to call `Go` and `Wait` concurrently
- parcour has `Close` and `Wait`, conc only `Wait`.
  Pro for conc: a single method is easier to understand.
  But I think it's also easier to misuse: you can't just `defer Close` and be sure you're not leaking tasks.

So I'm not convinced on conc's abstractions. The main things being they miss nesting and some of the failure propagation which are both a core part of structured concurrency. And the pools are not composable and create fractured and less reusable APIs.  
Hopefully I don't sound too harsh, I think the `iter` and `panics` packages have some interesting ideas, but globally I'm not sold.  
I'll look for other libraries to try and find something I like more.

config/config.go
```diff
@@ -263,6 +263,34 @@ func (c *toEnable) LogConfig(logger *logrus.Entry) {
...
+
+func (c *SourceLoadingConfig) LogConfig(logger *logrus.Entry) {
+	logger.Infof("concurrency = %d", c.Concurrency)
+	logger.Infof("maxErrorsPerSource = %d", c.MaxErrorsPerSource)
+}
+
+type DownloaderConfig struct {
+	Timeout  Duration `yaml:"timeout" default:"5s"`
+	Attempts uint     `yaml:"attempts" default:"3"`
+	Cooldown Duration `yaml:"cooldown" default:"500ms"`
```
0xERR0R, line 281: Should be documented in docs/configuration.md and docs/config.yml

go.mod
0xERR0R: If you think it makes sense to make a separate implementation for this and it provides additional features we need -> why not?

lists/list_cache.go
0xERR0R: I think, one limit for both is good enough

config/config.go
```diff
@@ -263,6 +263,34 @@ func (c *toEnable) LogConfig(logger *logrus.Entry) {
...
 
+type SourceLoadingConfig struct {
+	DownloaderConfig DownloaderConfig `yaml:",inline"`
+
+	Concurrency        uint `yaml:"concurrency" default:"4"`
+	MaxErrorsPerSource int  `yaml:"maxErrorsPerSource" default:"5"`
+}
+
+func (c *SourceLoadingConfig) LogConfig(logger *logrus.Entry) {
+	logger.Infof("concurrency = %d", c.Concurrency)
```
0xERR0R, line 274: Should it not also log all parameters from "DownloaderConfig"?

0xERR0R (changes requested):

resolver/hosts_file_resolver.go
```diff
@@ -4,47 +4,52 @@ import (
...
 
-		r.cfg.Filepath = "" // don't try parsing the file again
-	} else {
-		go r.periodicUpdate()
+	err := cfg.Loading.StartPeriodicRefresh(r.loadSources, func(err error) {
+		r.log().WithError(err).Errorf("could not load hosts files")
+	})
+	if err != nil {
+		return nil, err
 	}
```
ThinkChaos, line 50: The hosts file resolver will now behave like the blocking resolver and have a `startStrategy`.  
This is different from before: we were never trying again to parse the file if it failed at initialization.

I think this makes more sense now that we support remote files and multiple sources. Do you agree?

config/config.go
ThinkChaos: Was a bit afraid that'd be a lot of output.  
I added those, but also changed some to `Debugf`. Let me know if you think I should adjust the levels I chose.

config/config.go
ThinkChaos: Yeah I had made minor changes to the files locally to not forget, but was already thinking of supporting `RefreshPeriod` and `StartStrategy` in the hosts file resolver so wanted to wait until after that to update everything.  
Will get to it soon, still need to bring back the old options for back-compat.

lists/list_cache.go
ThinkChaos: Do you mean the current code is fine?  
TBH it feels like we're not really respecting the user's choice here to me. But it's already the case so if you want to keep it as is, it's fine by me.

resolver/hosts_file_resolver.go
0xERR0R: Yes, it makes definitively more sense. This should be documented properly to avoid confusions

ThinkChaos:
Will finish this up this week-end :)

ThinkChaos:
Didn't have time to do the docs or run the tests locally, I should have time tomorrow for that.

I refactored how we do config options migration cause it was quite verbose and copy-pasty. This will cause conflicts in the Redis rework branch, but I think it should be relatively easy to fix.  
The "_x_ and _y_ are both configured" configured and "please use _x_ instead" logs are automatically handled once you define the migrations with the new DSL.  
The pattern that branch introduces of returning an error from `RedisConfig.validateConfig` can be supported by making validation a second pass after migration.

Also moving all the options into a single `Deprecated` struct is nice cause it prevents old code from compiling, so you're guaranteed to catch all usage of old options.  
That's how I noticed the old HTTP port option was used in `cmd/root.go`.

ThinkChaos:
Just pushed the docs, and cleaned up the commits.

You can see the rendered docs [here](https://thinkchaos.github.io/blocky/feat-multiple-hosts-sources-ci/configuration/).

config/config.go
```diff
@@ -263,6 +285,86 @@ func (c *toEnable) LogConfig(logger *logrus.Entry) {
 	logger.Info("enabled")
 }
 
+type SourceLoadingConfig struct {
+	Concurrency        uint              `yaml:"concurrency" default:"4"`
+	MaxErrorsPerSource int               `yaml:"maxErrorsPerSource" default:"5"`
+	RefreshPeriod      Duration          `yaml:"refreshPeriod" default:"4h"`
+	StartStrategy      StartStrategyType `yaml:"startStrategy" default:"blocking"`
+	Downloads          DownloaderConfig  `yaml:"downloads"`
```
ThinkChaos, line 293: Some of the defaults were changed here.
I don't think we currently have a way to manually write changelog text but if there's an easy way to do it it could be nice to mention this.

resolver/hosts_file_resolver.go
ThinkChaos: This would also make more sense in the changelog IMO, not resolving this yet since if that's not easy to do I can add it in the docs.

0xERR0R:
Is there some open points or is it ready to be merged? Can you please rebase it, there is some conflict in go.mod

kwitsch:
I like the new implementation for config depreciations. ♥️


Sadly I have currently very little time so I won't be able to test it. It would help if there was a propper Go development suit for Android. 😕

0xERR0R (approved):

ThinkChaos:
I'd prefer to merge this without squashing if possible, but I think the repo settings prevent that.  
Would you want to change the settings to allow that, or should I just squash and merge?

0xERR0R:
> I'd prefer to merge this without squashing if possible, but I think the repo settings prevent that. Would you want to change the settings to allow that, or should I just squash and merge?

I think you must only rebase your branch on master. It is not required to squash commits.

ThinkChaos:
Thanks, I rebased and there was a `go.sum` conflict. So I guess that's why GitHub didn't want to do it.  
The WebUI didn't mention that, or I missed it. Anyways, merged it :)

PR #1007, opened 2023-04-30
Issues/952

### LINKED ISSUE #952: When blocky is used as the upstream of other dns forwarders
When blocky is used as the upstream of other dns forwarders, client_ip and client_names are not displayed as the real ip of the client, but are displayed as the internal network 127.0.0.1 ip

kwitsch:
Closes #952

Tested it with some blocky containers in a chain and it works well.

I'm currently at a loss how to write propper unit tests as most of the testing has to be done in the mock resolver that is configured as next(?). 🤔

Ideas & hints how to do it properly are appreciate. 😅

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -181,6 +181,7 @@ type Config struct {
 	FqdnOnly            FqdnOnlyConfig            `yaml:",inline"`
 	Filtering           FilteringConfig           `yaml:"filtering"`
 	Ede                 EdeConfig                 `yaml:"ede"`
+	Ecs                 EcsConfig                 `yaml:"ecs"`
 	// Deprecated
 	DisableIPv6 bool `yaml:"disableIPv6" default:"false"`
 	// Deprecated
--- a/config/config.go
+++ b/config/config.go
@@ -398,6 +398,8 @@ func validateConfig(cfg *Config) {
 		}
 	}
 
+	cfg.Ecs.validateConfig()
+
 	fixDeprecatedLog(cfg)
 
 	fixDeprecatedPorts(cfg)
--- a/config/ecs.go
+++ b/config/ecs.go
@@ -1,6 +1,14 @@
 package config
 
-import "github.com/sirupsen/logrus"
+import (
+	"github.com/0xERR0R/blocky/log"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	ipv4MaskMax = 32
+	ipv6MaskMax = 128
+)
 
 type EcsConfig struct {
 	UseEcsAsClient bool  `yaml:"useEcsAsClient" default:"false"`
@@ -17,4 +25,20 @@ func (c *EcsConfig) LogConfig(logger *logrus.Entry) {
 	logger.Infof("Use ECS as client = %t", c.UseEcsAsClient)
 	logger.Infof("IPv4 netmask      = %d", c.IPv4Mask)
 	logger.Infof("IPv6 netmask      = %d", c.IPv6Mask)
+}
+
+func (c *EcsConfig) validateConfig() {
+	if c.IPv4Mask > ipv4MaskMax {
+		log.Log().Errorf("the current value %d of ipv4Mask is above the maxvalue of %d",
+			c.IPv4Mask, ipv4MaskMax)
+
+		c.IPv4Mask = 0
+	}
+
+	if c.IPv6Mask > ipv6MaskMax {
+		log.Log().Errorf("the current value %d of ipv6Mask is above the maxvalue of %d",
+			c.IPv6Mask, ipv6MaskMax)
+
+		c.IPv6Mask = 0
+	}
 }
--- a/config/ecs.go
+++ b/config/ecs.go
@@ -1,0 +1,19 @@
+package config
+
+import "github.com/sirupsen/logrus"
+
+type EcsConfig struct {
+	UseEcsAsClient bool  `yaml:"useEcsAsClient" default:"false"`
+	IPv4Mask       uint8 `yaml:"ipv4Mask" default:"0"`
+	IPv6Mask       uint8 `yaml:"ipv6Mask" default:"0"`
+}
+
+func (c *EcsConfig) IsEnabled() bool {
+	return c.UseEcsAsClient || c.IPv4Mask > 0 || c.IPv6Mask > 0
+}
+
+func (c *EcsConfig) LogConfig(logger *logrus.Entry) {
+	logger.Infof("Use ECS as client = %t", c.UseEcsAsClient)
+	logger.Infof("IPv4 netmask      = %d", c.IPv4Mask)
+	logger.Infof("IPv6 netmask      = %d", c.IPv6Mask)
+}
--- a/config/ecs.go
+++ b/config/ecs.go
@@ -4,6 +4,7 @@ import "github.com/sirupsen/logrus"
 
 type EcsConfig struct {
 	UseEcsAsClient bool  `yaml:"useEcsAsClient" default:"false"`
+	ForwardEcs     bool  `yaml:"forwardEcs" default:"false"`
 	IPv4Mask       uint8 `yaml:"ipv4Mask" default:"0"`
 	IPv6Mask       uint8 `yaml:"ipv6Mask" default:"0"`
 }
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -31,8 +31,8 @@ configuration properties as [JSON](config.yml).
 
 All logging port are optional.
 
-| Parameter  | Type                   | Default value | Description                                                                                                                                                                                                                                       |
-|------------|------------------------|---------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Parameter   | Type                   | Default value | Description                                                                                                                                                                                                                                       |
+|-------------|------------------------|---------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | ports.dns   | [IP]:port[,[IP]:port]* | 53            | Port(s) and optional bind ip address(es) to serve DNS endpoint (TCP and UDP). If you wish to specify a specific IP, you can do so such as `192.168.0.1:53`. Example: `53`, `:53`, `127.0.0.1:53,[::1]:53`                                         |
 | ports.tls   | [IP]:port[,[IP]:port]* |               | Port(s) and optional bind ip address(es) to serve DoT DNS endpoint (DNS-over-TLS). If you wish to specify a specific IP, you can do so such as `192.168.0.1:853`. Example: `83`, `:853`, `127.0.0.1:853,[::1]:853`                                |
 | ports.http  | [IP]:port[,[IP]:port]* |               | Port(s) and optional bind ip address(es) to serve HTTP used for prometheus metrics, pprof, REST API, DoH... If you wish to specify a specific IP, you can do so such as `192.168.0.1:4000`. Example: `4000`, `:4000`, `127.0.0.1:4000,[::1]:4000` |
@@ -734,6 +734,25 @@ Configuration parameters:
     ```yaml
     ede:
       enable: true
+    ```
+
+## EDNS Client Subnet options
+
+Configuration parameters:
+
+| Parameter          | Type | Mandatory | Default value | Description                                                                   |
+|--------------------|------|-----------|---------------|-------------------------------------------------------------------------------|
+| ecs.useEcsAsClient | bool | no        | false         | Use ECS if it is present with a netmask is 32 for IPv4 or 128 for IPv6        |
+| ecs.forwardEcs     | bool | no        | false         | Forward ECS option to upstream                                                |
+| ecs.ipv4Mask       | int  | no        | 0             | Add ECS option for IPv4 requests if mask is greater than zero (max value 32)  |
+| ecs.ipv6Mask       | int  | no        | 0             | Add ECS option for IPv6 requests if mask is greater than zero (max value 128) |
+
+!!! example
+
+    ```yaml
+    ecs:
+      ipv4Mask: 32
+      ipv6Mask: 128
     ```
 
 ## SSL certificate configuration (DoH / TLS listener)
--- a/resolver/ecs_resolver.go
+++ b/resolver/ecs_resolver.go
@@ -30,23 +30,38 @@ func (r *EcsResolver) Resolve(request *model.Request) (*model.Response, error) {
 }
 
 func (r *EcsResolver) setClientIP(request *model.Request) bool {
-	if edns := request.Req.IsEdns0(); edns != nil {
-		for _, o := range edns.Option {
-			switch sub := o.(type) {
-			case *dns.EDNS0_SUBNET:
-				// v4 and unmasked
-				if (sub.Family == 1 && sub.SourceNetmask == 32) ||
-					// v6 and unmasked
-					(sub.Family == 2 && sub.SourceNetmask == 128) {
-					request.ClientIP = sub.Address
-				}
+	edns := request.Req.IsEdns0()
+	if edns == nil {
+		return false
+	}
 
-				return true
+	result := false
+
+	newOpts := []dns.EDNS0{}
+
+	for _, o := range edns.Option {
+		switch so := o.(type) {
+		case *dns.EDNS0_SUBNET:
+			// v4 and mask
+			if (so.Family == 1 && so.SourceNetmask == 32) ||
+				// v6 and unmasked
+				(so.Family == 2 && so.SourceNetmask == 128) {
+				request.ClientIP = so.Address
 			}
+
+			if r.cfg.ForwardEcs {
+				newOpts = append(newOpts, o)
+			}
+
+			result = true
+		default:
+			newOpts = append(newOpts, o)
 		}
 	}
 
-	return false
+	edns.Option = newOpts
+
+	return result
 }
 
 func (r *EcsResolver) appendSubnet(request *model.Request) {
@@ -68,6 +83,10 @@ func (r *EcsResolver) appendSubnet(request *model.Request) {
 }
 
 func (r *EcsResolver) appendOption(request *model.Request, opt dns.EDNS0) {
+	if edns := request.Req.IsEdns0(); edns != nil {
+		edns.Option = append(edns.Option, opt)
+	}
+
 	o := new(dns.OPT)
 	o.Hdr.Name = "."
 	o.Hdr.Rrtype = dns.TypeOPT
--- a/resolver/ecs_resolver.go
+++ b/resolver/ecs_resolver.go
@@ -1,0 +1,77 @@
+package resolver
+
+import (
+	"github.com/0xERR0R/blocky/config"
+	"github.com/0xERR0R/blocky/model"
+	"github.com/miekg/dns"
+)
+
+type EcsResolver struct {
+	configurable[*config.EcsConfig]
+	NextResolver
+	typed
+}
+
+func NewEcsResolver(cfg config.EcsConfig) ChainedResolver {
+	return &EcsResolver{
+		configurable: withConfig(&cfg),
+		typed:        withType("extended_client_subnet"),
+	}
+}
+
+func (r *EcsResolver) Resolve(request *model.Request) (*model.Response, error) {
+	if r.cfg.IsEnabled() {
+		if !r.setClientIP(request) {
+			r.appendSubnet(request)
+		}
+	}
+
+	return r.next.Resolve(request)
+}
+
+func (r *EcsResolver) setClientIP(request *model.Request) bool {
+	if edns := request.Req.IsEdns0(); edns != nil {
+		for _, o := range edns.Option {
+			switch sub := o.(type) {
+			case *dns.EDNS0_SUBNET:
+				// v4 and unmasked
+				if (sub.Family == 1 && sub.SourceNetmask == 32) ||
+					// v6 and unmasked
+					(sub.Family == 2 && sub.SourceNetmask == 128) {
+					request.ClientIP = sub.Address
+				}
+
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+func (r *EcsResolver) appendSubnet(request *model.Request) {
+	e := new(dns.EDNS0_SUBNET)
+	e.Code = dns.EDNS0SUBNET
+	e.SourceScope = 0
+
+	if ip := request.ClientIP.To4(); ip != nil && r.cfg.IPv4Mask > 0 {
+		e.Family = 1
+		e.SourceNetmask = r.cfg.IPv4Mask
+		e.Address = ip
+		r.appendOption(request, e)
+	} else if request.ClientIP.To16() != nil && r.cfg.IPv6Mask > 0 {
+		e.Family = 2
+		e.SourceNetmask = r.cfg.IPv6Mask
+		e.Address = ip
+		r.appendOption(request, e)
+	}
+}
+
+func (r *EcsResolver) appendOption(request *model.Request, opt dns.EDNS0) {
+	o := new(dns.OPT)
+	o.Hdr.Name = "."
+	o.Hdr.Rrtype = dns.TypeOPT
+	o.Option = append(o.Option, opt)
+
+	request.Req.Extra = append(request.Req.Extra, o)
+}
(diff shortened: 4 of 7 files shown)


kwitsch:
> You can take a look at the rewriter, it has some tests that hook into the inner and next resolvers. For example: https://github.com/0xERR0R/blocky/blob/v0.21/resolver/rewriter_resolver_test.go#L144-L150

Thanks, that's axactly what I was looking for. ❤️

bjw-s:
Having support for ECS would seriously benefit my use case for Blocky! Just wanted to drop a big thank you for working on this!

bjw-s:
Still very much needed

buroa:
Would also love this in my network!

fisherbln:
> Would also love this in my network!

same!

kwitsch:
@bjw-s @buroa @fisherbln  Sorry guys i got almost no time at least until end of next month. I'll work on it again as soon as time allows it.

buroa:
@kwitsch Any update on this?

kwitsch:
> @kwitsch Any update on this?

Still no time to work on it.

I'm currently only online with my mobile phone which isn't suitable to work on blocky.

Sorry to let you down. 😔

kwitsch:
Sorry I replied to it yesterday but it seems it's missing? 🤨

> @kwitsch Any update on this?

Sorry to let you down but I still had no time to work on blocky again. 😔

I'm currently only able to be online with my mobile phone which is not suitable for working on blocky. 🫣

0xERR0R:
@kwitsch  I skimmed over your PR and I think I can continue to work on it. 

I have some questions/discussion points:
- As far I understand, ECS makes only sense, if public IP is used. If blocky runs on a local network, it receives only local client addresses. Does it make sense to forward a local address (like 192.168.x) to a remote upstream? Would it better to check in EcsResolver if client IP is a local network and only if not, apply it to request? How covers it the use case of original issue (#952)?
- What is about caching? IMHO it not correct to store a server reply for one Client Submask and return the same answer to another one... Maybe it is better not to cache answers for ECS request?
- Just idea: maybe it would make sense to provide a config option to set a constant ECS IP/subnet? Use case: local network, someone is behind CGNAT and has only NAT address as IPv4 (100.64.x.x). User can set custom subnet of his ISP and ECS aware upstream resolvers will serve better results.

kwitsch:
> - As far I understand, ECS makes only sense, if public IP is used. If blocky runs on a local network, it receives only local client addresses. Does it make sense to forward a local address (like 192.168.x) to a remote upstream? Would it better to check in EcsResolver if client IP is a local network and only if not, apply it to request? How covers it the use case of original issue (#952)?

I viewed it the other way around: especially in local network setups this feature is handy 😅

Example for an IPv4 network:
Multiple servers with blocky on port 53 and netmask 32 resolve against another blocky instance in a secured network which isn't accessible to the original requestor. This way the secured instance gets the whole IP of the original requestor and uses it for client name resolution instead of the blocky instance working as gateway/proxy. If the secured instance resolves against a local server the ECS info can be forwarded to use it there.

> - What is about caching? IMHO it not correct to store a server reply for one Client Submask and return the same answer to another one... Maybe it is better not to cache answers for ECS request?

Agreed 👍

> - Just idea: maybe it would make sense to provide a config option to set a constant ECS IP/subnet? Use case: local network, someone is behind CGNAT and has only NAT address as IPv4 (100.64.x.x). User can set custom subnet of his ISP and ECS aware upstream resolvers will serve better results.

I think for Nat purpose a second instance for resolution would be the better choice (as described in the first example).

kwitsch:
@0xERR0R I can provide an example project (docker compose & blocky config) tomorrow if it helps clarifying how I intended this feature.

kwitsch:
Yes i'll be working on it again around the weekend.
Need to clean up some build up stuff (updates, API changes, etc) first.

kwitsch:
Depends on #1220 since i needed the PR for development it is already merged into this branch.

kwitsch:
Long long time ago... 😅

I got some time lately, so here it is. 😉

resolver/ecs_resolver.go
```diff
@@ -0,0 +1,103 @@
...
+	hasEcs := util.HasEdns0Option(request.Req, dns.EDNS0SUBNET)
+	if hasEcs && !r.cfg.ForwardEcs {
+		util.RemoveEdns0Option(request.Req, dns.EDNS0SUBNET)
+	}
+
+	return !hasEcs && r.cfg.ForwardEcs
+}
+
+// appendSubnet appends the subnet information to the request as EDNS0 option
+// if the client IP is IPv4 or IPv6 and the corresponding mask is set in the configuration
```
ThinkChaos, line 73: Minor rephrasing to hopefully help with clarity:
```suggestion
// appendSubnet adds the subnet information to the request as an EDNS0 option
// and sets the mask as configured depending on the request's IP version.
```

resolver/ecs_resolver.go
```diff
@@ -0,0 +1,103 @@
...
+	return r.next.Resolve(request)
+}
+
+// setClientIP sets the client IP from the EDNS0 option to the request if the
+// client IP is IPv4 or IPv6 and the corresponding mask is set in the configuration
+func (r *EcsResolver) setClientIP(request *model.Request) {
+	if eso := util.GetEdns0Option(request.Req, dns.EDNS0SUBNET); eso != nil {
+		so := eso.(*dns.EDNS0_SUBNET)
+		if (so.Family == ecsIpv4Family && so.SourceNetmask == ecsIpv4Mask) ||
+			(so.Family == ecsIpv6Family && so.SourceNetmask == ecsIpv6Mask) {
```
ThinkChaos, line 55: This whole check is repeated from `shouldRequestNotUseCache`. Maybe make a function?

I really wish `dns` would provide these kinds of checks and constants. And I've thought of adding a `blocky/dnx` or something just to have a  better place than `util` to collect these. Maybe someday :smile:

resolver/ede_resolver.go
```diff
@@ -42,12 +49,9 @@ func (r *EdeResolver) addExtraReasoning(res *model.Response) {
...
-	opt.Option = append(opt.Option, &dns.EDNS0_EDE{
-		InfoCode:  infocode,
-		ExtraText: res.Reason,
-	})
-	res.Res.Extra = append(res.Res.Extra, opt)
+	edeOption := new(dns.EDNS0_EDE)
+	edeOption.InfoCode = infocode
+	edeOption.ExtraText = res.Reason
+
+	util.SetEdns0Option(res.Res, edeOption)
```
ThinkChaos, line 56: :+1:

util/edns0.go
```diff
@@ -0,0 +1,131 @@
...
+
+	opt := GetEdns0Record(msg)
+	for _, o := range opt.Option {
+		if o.Option() == code {
+			return o
+		}
+	}
+
+	return nil
+}
```
ThinkChaos, line 85: Maybe we can merge these into just `GetEdns0Option(msg *dns.Msg, code uint16) (dns.EDNS0, bool)` like maps?  
To avoid doing the same thing twice in a row with `if Has(...) { _ = Get(...) }`.

util/edns0.go
```diff
@@ -0,0 +1,131 @@
...
+
+// SetEdns0Option adds the given option to the OPT record in the Extra section of the
+// given message. If the option already exists, it will be replaced.
+func SetEdns0Option(msg *dns.Msg, opt dns.EDNS0) {
+	if msg == nil {
+		return
+	}
+
+	optRecord := GetEdns0Record(msg)
+	newOpts := make([]dns.EDNS0, 0, len(optRecord.Option))
```
ThinkChaos, line 121: I think it's small enough to be worth risking allocating an unused slot:
```suggestion
	newOpts := make([]dns.EDNS0, 0, len(optRecord.Option) + 1)
```

util/edns0.go
```diff
@@ -0,0 +1,131 @@
...
+	opt := GetEdns0Record(msg)
+	newOpts := make([]dns.EDNS0, 0, len(opt.Option))
+
+	for _, o := range opt.Option {
+		if o.Option() != code {
+			newOpts = append(newOpts, o)
+		}
+	}
+
+	opt.Option = newOpts
```
ThinkChaos, line 106: We can modify in place and short circuit:
```suggestion
	for i, o := range opt.Option {
		if o.Option() == code {
			opt.Option = slices.Delete(opt.Option, i, i+1)

			break
		}
	}
```

util/edns0.go
```diff
@@ -0,0 +1,131 @@
...
+		extra := make([]dns.RR, 0, len(msg.Extra))
+
+		for _, rr := range msg.Extra {
+			if rr.Header().Rrtype != dns.TypeOPT {
+				extra = append(extra, rr)
+			}
+		}
+
+		msg.Extra = extra
+	}
```
ThinkChaos, line 25: Invert the `if`, modify in place, and short-circuit:
```suggestion
	if len(msg.Extra) == 0 {
		return
	}

	for i, rr := range msg.Extra {
		if rr.Header().Rrtype == dns.TypeOPT {
			msg.Extra = slices.Delete(msg.Extra, i, i+1)

			break
		}
	}
```

resolver/ecs_resolver.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/0xERR0R/blocky/util"
+	"github.com/miekg/dns"
+)
+
+const (
+	ecsSourceScope = uint8(0)
+	ecsIpv4Family  = uint16(1)
+	ecsIpv4Mask    = uint8(32)
+	ecsIpv6Family  = uint16(2)
+	ecsIpv6Mask    = uint8(128)
```
ThinkChaos, line 15: Maybe we can rename these to look more enum like?
I didn't adjust any other code, probably easier to do with an IDE :stuck_out_tongue: 

Also changing the type of the masks make it possible to compare with the config values easily, and make sure we won't compare to the wrong config value.

```suggestion
// https://www.rfc-editor.org/rfc/rfc7871.html#section-6
const (
	ecsSourceScope = uint8(0)

	ecsMaskIPv4 = config.ECSv4Mask(net.IPv4len * 8)
	ecsMaskIPv6 = config.ECSv6Mask(net.IPv6len * 8)
)

// https://www.iana.org/assignments/address-family-numbers/address-family-numbers.xhtml
const (
	ecsFamilyIPv4 = uint16(iota + 1)
	ecsFamilyIPv6
)
```

ThinkChaos (commented):
Lost of comments, sorry! But everything is minor and nit-picky. I tried to put suggestions as much as I could. Might need some reformatting though.
Feel free to ignore stuff!

kwitsch:
> Lost of comments, sorry! But everything is minor and nit-picky. I tried to put suggestions as much as I could. Might need some reformatting though.
> Feel free to ignore stuff!

Comments are always welcome. 😊👍

resolver/caching_resolver.go
```diff
@@ -232,22 +232,22 @@ func setTTLInCachedResponse(resp *dns.Msg, ttl time.Duration) {
...
-
-		for _, rr := range msg.Extra {
-			if rr.Header().Rrtype != dns.TypeOPT {
-				extra = append(extra, rr)
-			}
+// shouldRequestNotUseCache returns true if the request should be cached
+func shouldRequestNotUseCache(request *model.Request) bool {
+	// don't cache responses with EDNS Client Subnet option with masks that include more than one client
+	if eso := util.GetEdns0Option(request.Req, dns.EDNS0SUBNET); eso != nil {
+		so := eso.(*dns.EDNS0_SUBNET)
```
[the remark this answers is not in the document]
kwitsch: I tried that but wasn't sure how.  
In my opinion it should be possible with gernerics but I'm not very familar using them in Go.  
Ideally the type could be mapped to the option code so only the request and type has to be provided.

util/edns0.go
kwitsch: Has has beeen removed in favor of nil check on get.

liang-hiwin:
> > I would only disable caching if the ecs mask contains more than one IP(v4 <32 & v6 <128). Otherwise this feature wouldn't be usefull for most of the issues it was intended to solve.
> 
> Definitely. I have a lot of clients in my network and would love to capture the IP; but I also do a ton of caching... so if it's favor of caching vs ecs, then I would pick ecs. Hopefully we can have both of these features enabled at the same time :)

I also support it. It would be best if the parsed value corresponding to ecs can be cached.

kwitsch:
> I also support it. It would be best if the parsed value corresponding to ecs can be cached.

I agree that this would be the best solution.  
But according to my knowledge this would require extensive work on the caching resolver.  
This seems to be out of scope of this PR.

The current implementation disables caching if the request contains ECS information with a subnet of more than one IP.  
Seperating the response caches according to the request information could be added later on.

ThinkChaos:
Created #1256 for those kinds of requests (not necessarily ECS specific)

liang-hiwin:
> > I also support it. It would be best if the parsed value corresponding to ecs can be cached.
> 
> I agree that this would be the best solution. But according to my knowledge this would require extensive work on the caching resolver. This seems to be out of scope of this PR.
> 
> The current implementation disables caching if the request contains ECS information with a subnet of more than one IP. Seperating the response caches according to the request information could be added later on.



I think they should be cached separately instead of disabling caching. Because this is the most perfect.
for example:
**1. When the client subnet IP is: 1.1.1.1, the 1.com domain name resolution result cache: 1.2.3.4
2. When the client subnet IP is: 2.2.2.2, the 1.com domain name resolution result cache: 2.3.4.5**

.....

liang-hiwin:
Although it takes a lot of time to write code, I believe that even if I write 500 lines a day, I can eventually achieve it one day. Also thank you for your contribution

kwitsch:
> I think they should be cached separately instead of disabling caching. Because this is the most perfect.

As already said: out of scope of this PR

kwitsch:
> Although it takes a lot of time to write code, I believe that even if I write 500 lines a day, 

Lines/day is a silly measurement nobody should use 🫣

> I can eventually achieve it one day.

You are always welcome to make your own contribution. 😊

resolver/ecs_resolver.go
```diff
@@ -0,0 +1,103 @@
...
+	e.SourceScope = ecsSourceScope
+
+	if ip := request.ClientIP.To4(); ip != nil {
+		if mask := getMask(uint8(r.cfg.IPv4Mask), ecsIpv4Mask); mask > 0 {
+			e.Family = ecsIpv4Family
+			e.SourceNetmask = mask
+			e.Address = ip
+			util.SetEdns0Option(request.Req, e)
+		}
+	} else if request.ClientIP.To16() != nil {
```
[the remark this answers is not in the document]
kwitsch: Checking both is more fault tollerant in my opinion.
If the value of ClientIP is incorrect checking it with `else if` prevents broken ecs records.

resolver/ecs_resolver.go
```diff
@@ -0,0 +1,103 @@
...
+// if the client IP is IPv4 or IPv6 and the corresponding mask is set in the configuration
+func (r *EcsResolver) appendSubnet(request *model.Request) {
+	e := new(dns.EDNS0_SUBNET)
+	e.Code = dns.EDNS0SUBNET
+	e.SourceScope = ecsSourceScope
+
+	if ip := request.ClientIP.To4(); ip != nil {
+		if mask := getMask(uint8(r.cfg.IPv4Mask), ecsIpv4Mask); mask > 0 {
+			e.Family = ecsIpv4Family
+			e.SourceNetmask = mask
```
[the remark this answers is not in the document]
kwitsch: Refactored the function.

util/edns0_test.go
```diff
@@ -90,7 +90,6 @@ var _ = Describe("EDNS0 utils", func() {
 			opt.Hdr.Name = "."
 			opt.Hdr.Rrtype = dns.TypeOPT
 			eso := new(dns.EDNS0_SUBNET)
-			eso.Code = dns.EDNS0SUBNET
```
ThinkChaos: Is this intentional?

kwitsch: Yes since `Code` is completely irrelevant if `.Option()` is used. 😉

kwitsch:
@ThinkChaos Caching should be only disabled if max time is negative according to the [documentation](https://0xerr0r.github.io/blocky/v0.22/configuration/#caching).  
Is my last commit really a fix or did I miss something? 🤔

ThinkChaos:
I think 0 doesn't matter: why would we cache something to delete it immediately?  
So I'd rather update the docs than the code.

util/edns0.go
```diff
@@ -0,0 +1,130 @@
...
+func GetEdns0Option[T EDNS0Option](msg *dns.Msg) T {
+	code := getCode[T](msg)
+	if code == 0 {
+		return nil
+	}
+
+	opt := GetEdns0Record(msg)
+	for _, o := range opt.Option {
+		if o.Option() == code {
+			return o.(T)
```
ThinkChaos, line 61: I guess it's not supposed to happens because `dns` wouldn't create a broken struct, but having a nice error message can't hurt:
```suggestion
			t, ok := o.(T)
			if !ok {
				panic(fmt.Errorf("dns option with code %d is not of type %T", code, t))
			}
```

model/models.go
```diff
@@ -19,6 +19,7 @@ import (
 // FILTERED // the query was filtered by query type
 // NOTFQDN // the query was filtered as it is not fqdn conform
 // SPECIAL // the query was resolved by the special use domain name resolver
+// OTHER // won't be used for EDE option
```
ThinkChaos, line 22: Other should be technically be used but we ignore it because of broken clients. So having this for "won't be used" seems like it might make it harder for us to return EDE other in the future (if/when clients are fixed).  
The only spot it's used is in a test. So maybe we change the test to use a value that's not in the enum?  
Like `model.ResponseType(math.MaxInt)`

config/ecs.go
```diff
@@ -0,0 +1,81 @@
...
+
+	*x = ECSv6Mask(res)
+
+	return nil
+}
+
+// EcsConfig is the configuration of the ECS resolver
+type EcsConfig struct {
+	UseEcsAsClient bool      `yaml:"useEcsAsClient" default:"false"`
+	ForwardEcs     bool      `yaml:"forwardEcs" default:"false"`
```
ThinkChaos, line 48: Maybe rename this to `forwardUpstream` or just `forward` since it's already under `ecs:`?

util/edns0.go
```diff
@@ -0,0 +1,130 @@
...
+// It is used to get the option code as uint16 for generics.
+// If the given message is nil or doesn't contain an OPT record, 0 will be returned.
+func getCode[T EDNS0Option](msg *dns.Msg) uint16 {
+	if msg == nil || msg.IsEdns0() == nil {
+		return 0
+	}
+
+	ref := *new(T)
+
+	return ref.Option()
```
ThinkChaos, line 129: Avoid allocating:
```suggestion
	var t T

	return t.Option()
```

resolver/ede_resolver_test.go
```diff
@@ -81,8 +82,17 @@ var _ = Describe("EdeResolver", func() {
...
+			opt := res.Res.IsEdns0()
+			if opt != nil {
+				for _, o := range opt.Option {
+					if o.Option() == dns.EDNS0EDE {
+						return *o.(*dns.EDNS0_EDE)
+					}
+				}
+			}
+
+			return dns.EDNS0_EDE{}
```
ThinkChaos, line 95: Can this use `util.GetEdns0Option`?

```suggestion
		extractEdeOption := func(res *Response) dns.EDNS0_EDE {
			return utils.GetEdns0Option[dns.EDNS0_EDE](res.Res)
```

util/edns0.go
```diff
@@ -0,0 +1,130 @@
...
+		return res
+	}
+
+	res := new(dns.OPT)
+	res.Hdr.Name = "."
+	res.Hdr.Rrtype = dns.TypeOPT
+	msg.Extra = append(msg.Extra, res)
+
+	return res
+}
```
ThinkChaos, line 48: I think a `Get` function shouldn't create the option if missing. Can we move that part to `SetEdns0Option` and here always return `msg.IsEdns0()`?  
Maybe we can do without the function if it's just that though.

resolver/ecs_resolver.go
```diff
@@ -0,0 +1,104 @@
...
+	ecsFamilyIPv6
+)
+
+// ECSMask is an interface for all ECS subnet masks as type constraint for generics
+type ECSMask interface {
+	config.ECSv4Mask | config.ECSv6Mask
+}
+
+// A EcsResolver is responsible for adding the subnet information as EDNS0 option
+type EcsResolver struct {
```
ThinkChaos, line 33: Had to go in the RFC to figure out what `ECS` means. Let's add it to the comment:
```suggestion
// EcsResolver is responsible for adding the EDNS Client Subnet information as EDNS0 option.
type EcsResolver struct {
```

And if possible rename to have `Ecs` in all caps since it's an abbreviation. The `New` function too in that case.

docs/configuration.md
```diff
@@ -714,6 +714,25 @@ Configuration parameters:
       enable: true
     ```
 
+## EDNS Client Subnet options
+
+Configuration parameters:
```
ThinkChaos, line 719: Trying to make the abbreviation obvious in the docs too:
```suggestion
EDNS Client Subnet (ECS) configuration parameters:
```

config/config.go
```diff
@@ -211,6 +211,7 @@ type Config struct {
 	FqdnOnly            FqdnOnlyConfig            `yaml:"fqdnOnly"`
 	Filtering           FilteringConfig           `yaml:"filtering"`
 	Ede                 EdeConfig                 `yaml:"ede"`
+	Ecs                 EcsConfig                 `yaml:"ecs"`
```
ThinkChaos, line 214: Rename to `ECS`  since it's an abbreviation?

ThinkChaos (commented):
Just a couple details I noticed going through everything again.

config/config.go
kwitsch: Could be done but wouldn't that also apply to `Ede` & `Fqdn`?  
I think renaming all of them to unifying the name pattern would be out of scope and maybe part of another PR.(?)

util/edns0.go
kwitsch: Maybe renaming it would be an option.

Since there should only be one EDNS0 OPT record I thought it would be handy to have one function which always returns this record.😅. 
Always checking for it's existence and creating it if it's missing would possibly create code redundancy....

config/ecs.go
kwitsch: Agreed `UseAsClient` & `Forward` would reduce naming redundancy.

resolver/ede_resolver_test.go
kwitsch: Correct 👍

Missed it during refactoring 🤦‍♂️

util/edns0.go
ThinkChaos: Yeah but if we have it in `SetEdns0Option`, we still wouldn't need that pattern IIUC?

config/config.go
ThinkChaos: Yeah agreed just commented since this is new code. And we have SUDN that follows the convention already.  
I'll make an issue and we can do it separately.

kwitsch: Ok I'll rename it to `ECS` and we'll rename the others in a separate PR. 👍

util/edns0.go
kwitsch: You're right.

Since we have utility functions to directly manipulate the options this function is technically obsolete.

I also considered moving it to another package since util is a bit generic but wasn't sure about the name.  
Since `dns` would often interfer with other libraries... maybe `dnsutil`? 🤔

config/config.go
ThinkChaos: We can do this in another too, doesn't matter to me :)

util/edns0.go
ThinkChaos: I think leaving it here for now is fine. We have a bunch of other stuff that should go in such a package so I think it'd be best to do that separately.

kwitsch:
@ThinkChaos I think I resolved all of your comments. :)

ThinkChaos (approved):

ThinkChaos:
Yes looks all good to me!

buroa:
@kwitsch Big BIG thanks to this!!

kwitsch:
> @kwitsch Big BIG thanks to this!!

I'm glad I could finally finish it & that it's appreciated. 😊  
Sorry again for taking this log...I had some RL struggles. 😖

PR #1008, opened 2023-04-30
Rueidis base

kwitsch:
Redis rework #945
- Rewrite current logic to use rueidis

Currently no unit tests as they have to be rewritten and the logic changes with forther implementation of the ticket.


kwitsch:
@ThinkChaos I moved the context as you requested in my previous PR.

config/config.go
```diff
@@ -181,24 +181,15 @@ type Config struct {
...
-	TLSPorts ListenConfig `yaml:"tlsPort"`
+	DisableIPv6         bool                      `yaml:"disableIPv6" default:"false"` // Deprecated: use Filtering with TypeAAAA
+	LogLevel            log.Level                 `yaml:"logLevel" default:"info"`     // Deprecated: use Log.Level
+	LogFormat           log.FormatType            `yaml:"logFormat" default:"text"`    // Deprecated: use Log.Format
+	LogPrivacy          bool                      `yaml:"logPrivacy" default:"false"`  // Deprecated: use Log.Privacy
+	LogTimestamp        bool                      `yaml:"logTimestamp" default:"true"` // Deprecated: use Log.Timestamp
+	DNSPorts            ListenConfig              `yaml:"port" default:"53"`           // Deprecated: use Ports.DNS
+	HTTPPorts           ListenConfig              `yaml:"httpPort"`                    // Deprecated: use Ports.HTTP
+	HTTPSPorts          ListenConfig              `yaml:"httpsPort"`                   // Deprecated: use Ports.HTTPS
+	TLSPorts            ListenConfig              `yaml:"tlsPort"`                     // Deprecated: use Ports.TLS
```
ThinkChaos, line 192: Nice to add the `use`!

config/redis.go
```diff
@@ -0,0 +1,50 @@
...
+	Address            string   `yaml:"address"`                  // Deprecated: use Addresses
+	SentinelAddresses  []string `yaml:"sentinelAddresses"`        // Deprecated: use Addresses
+}
+
+func fixDeprecatedRedis(cfg *Config) {
+	if len(cfg.Redis.Addresses) == 0 && len(cfg.Redis.SentinelAddresses) > 0 {
+		log.Log().Warnln("'redis.sentinelAddresses' is deprecated. Please use 'redis.addresses' instead.")
+
+		cfg.Redis.Addresses = cfg.Redis.SentinelAddresses
+	} else if len(cfg.Redis.Addresses) == 0 && len(cfg.Redis.Address) > 0 {
```
ThinkChaos, line 34: Should the `else` be removed here so we use both `SentinelAddresses` and `Address` if they're both set?  
Might be obvious we don't want that but I don't know Redis enough :smile: 

The `len(cfg.Redis.Addresses) == 0` condition would need to be evaluated before the first if to still work.

redis/redis.go
```diff
@@ -146,59 +146,114 @@ func (c *Client) PublishEnabled(state *EnabledMessage) {
...
+			c.client.Do(ctx,
+				c.client.B().Publish().
+					Channel(SyncChannelName).
+					Message(rueidis.BinaryString(binMsg)).
+					Build())
 		}
 	}
 }
 
 // GetRedisCache reads the redis cache and publish it to the channel
```
ThinkChaos, line 158: Small grammar fix since you're already changing the function:
```suggestion
// GetRedisCache reads the redis cache and publishes it to the channel
```

redis/redis.go
```diff
@@ -146,59 +146,114 @@ func (c *Client) PublishEnabled(state *EnabledMessage) {
...
+			}
+
+			if cursor == 0 { // no more keys
+				break
 			}
 		}
 	}()
 }
 
+func generateClientOptions(cfg *config.RedisConfig) rueidis.ClientOption {
```
ThinkChaos, line 200: Could this be a `RedisConfig` method?

redis/redis.go
```diff
@@ -146,59 +146,114 @@ func (c *Client) PublishEnabled(state *EnabledMessage) {
...
-				}
-			} else {
+		var cursor uint64
+
+		searchKey := prefixKey("*")
+
+		for {
+			sres, err := c.client.Do(ctx, c.client.B().Scan().Cursor(cursor).Match(searchKey).Count(1).Build()).AsScanEntry()
+			if err != nil {
 				c.l.Error("GetRedisCache ", err)
```
ThinkChaos, line 170: Would be nice to expand the error message, not 100% sure what I wrote is appropriate, so please improve if you wish:
```suggestion
				c.l.Errorf("could not start a searching through Redis cache: %s", err)
```

redis/redis.go
```diff
@@ -146,59 +146,114 @@ func (c *Client) PublishEnabled(state *EnabledMessage) {
...
+
+				if resp.Error() == nil {
+					str, err := resp.ToString()
+					if err == nil {
+						response, err := c.getResponse(ctx, str)
+						if err == nil {
+							c.CacheChannel <- response
+						}
+					}
+				}
```
ThinkChaos, line 190: Are any of these errors worth logging?

e2e/redis_test.go
```diff
@@ -104,73 +106,6 @@ var _ = Describe("Redis configuration tests", func() {
 			})
 		})
 	})
-
-	Describe("Cache loading on startup", func() {
```
ThinkChaos, line 108: Since these are end to end tests, I think we could keep them as is no?

resolver/blocking_resolver_test.go
```diff
@@ -1130,78 +1127,4 @@ var _ = Describe("BlockingResolver", Label("blockingResolver"), func() {
 			})
 		})
 	})
-
-	Describe("Redis is configured", func() {
```
ThinkChaos, line 1134: Would be nice to keep this, and other tests that validate redis use. You know better if it's feasible, but just wanted to create a thread and see if there's something we can do about it.

config/config.go
```diff
@@ -400,6 +375,10 @@ func validateConfig(cfg *Config) {
 	fixDeprecatedLog(cfg)
 
 	fixDeprecatedPorts(cfg)
+
+	fixDeprecatedRedis(cfg)
+
+	return validateRedisConfig(cfg)
```
ThinkChaos, line 381: These would be nicer as `RedisConfig` methods IMO.

Maybe we can also merge them into just `validate`?  
Since anyways we're already "fixing" inside `validateConfig`, I don't think it won't be out of place.

And if we merge those functions we can also return an error if both a new and old options have values, instead of ignoring the old value as the current `fixDeprecatedRedis` does.

ThinkChaos (commented):
I'm not the best person to review Redis stuff, but the context usage seems good!

config/redis.go
kwitsch: Nope rueidis uses only one address pool which only contains the sentinel addresses if sentinel is used.
Otherwise the pool contains all redis instances(multiple entries if a redis cluster is used).
A mix of sentinel and redis addresses is not allowed (as far as I know).

redis/redis.go
kwitsch: Error description is correct 👍

redis/redis.go
kwitsch: An error at this location would indicate a broken record in the database.
I guess we could log it but it's unlikely that it would procure as almost all keys are stored with a ttl.
And since the cache is usually quite big and not really maintainable for most users I deemed it as unnessecary.

Debug level would be ok I guess(?)

e2e/redis_test.go
kwitsch: I removed it since it kept failing on GitHub when a nearly identically setup worked on my machine. 🤷‍♂️

Thought I investigate it further when I add the now removed unit tests again.

resolver/blocking_resolver_test.go
kwitsch: Not possible without a complete rewrite of the test itself since it is based on miniredis which is not compatible to rueidis.

The logic regarding the redis usage will change significantly so it would be more reasonable to develop new unit tests for the new logic.

ThinkChaos: Ok, you know better than me :)

e2e/redis_test.go
ThinkChaos: Sounds good.

redis/redis.go
kwitsch: moved it

config/config.go
kwitsch: implemented it as `RedisConfig.validateConfig`

0xERR0R (commented):
Looks good, good job! Please check my comments? I think it would be better to merge this PR on "fb-redis-integration" branch and not main, what do you think?

redis/redis.go
```diff
@@ -146,59 +148,91 @@ func (c *Client) PublishEnabled(state *EnabledMessage) {
...
+					Message(rueidis.BinaryString(binMsg)).
+					Build())
 		}
 	}
 }
 
-// GetRedisCache reads the redis cache and publish it to the channel
-func (c *Client) GetRedisCache() {
+// GetRedisCache reads the redis cache and publishes it to the channel
+func (c *Client) GetRedisCache(ctx context.Context) {
```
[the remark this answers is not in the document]
kwitsch: Correct this is the plan for the next Pars.
In the first one I tried implementing the current logic with rueidis.

There will be no need for a pub/sub channel in the future.

redis/redis.go
```diff
@@ -56,74 +57,75 @@ type EnabledMessage struct {
...
-				EnabledChannel: make(chan *EnabledMessage, chanCap),
-			}
+	roption.ClientName = fmt.Sprintf("blocky-%s", util.HostnameString())
 
-			// start channel handling go routine
-			err = res.startup()
+	var client rueidis.Client
 
-			return res, err
+	for i := 0; i < cfg.ConnectionAttempts; i++ {
```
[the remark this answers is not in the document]
kwitsch: Will look into it. 👍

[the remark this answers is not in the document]
kwitsch: Will look into it. 👍

kwitsch:
> I think it would be better to merge this PR on "fb-redis-integration" branch and not main, what do you think?

Sorry I completely forgot that branch. 🤦‍♂️
I'll change it.

PR #1066, opened 2023-07-06
docs: Update prometheus_grafana.md with correct grafana config

aman207:
Hi,
This PR is to update the new/correct anchor link for disable_sanitize_html
It also uses underscores rather than dashes in the config.ini


0xERR0R (approved):
👍 Thanks!

PR #1067, opened 2023-07-09
feat(sudn): refactor and expand supported special use domains RFCs

ThinkChaos:
Expand the SUDN resolver to support all the domains listed on [Wikipedia](https://en.wikipedia.org/wiki/Special-use_domain_name).

I made `rfc6762-appendixG` default to `true`, I think it's unlikely to break people's setups (see the comment on the struct field), but could easily be changed to `false` to be extra safe.

Rendered docs: https://thinkchaos.github.io/blocky/feat-expand-sudn/configuration/#special-use-domain-names

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,10 +26,11 @@ jobs:
           git config --local user.email "github-actions[bot]@users.noreply.github.com"
           git config --local user.name "github-actions[bot]"
 
-      - name: Deploy version for branch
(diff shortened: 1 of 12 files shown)


kwitsch (approved):
Especially the handler mapping looks good. 👍

0xERR0R (approved):

ThinkChaos:
Thanks for the reviews!

PR #1086, opened 2023-07-28
refactor(config): allow more configuration for `upstreams`

ThinkChaos:
Rename the `upstream` option to `upstreams.groups` so we can have more `upstreams` options.


kwitsch (approved):
👍

DerRockWolf:
Thanks for already working on the config part.

What do you think of the idea with both "global" upstream and group specific settings?
Instead of the group name being the list of upstream servers they would be under the "resolvers" key. Alternatively we could get rid of the "global" upstream settings and only have group specific setting. What do you think?

example for both global and group specific configuration:
```yaml
upstreams:
  timeout: 5s
  upstreamStrategy: parallel_best
  otherSettingUsedByAllGroups: xyz
  groups:
    <name>:
      timeout: 5s
      upstreamStrategy: parallel_best
      resolvers:
        - 1.1.1.1
        - 8.8.8.8
```
example for only group specific configuration:
```yaml
upstreams:
  groups:
    <name>:
      timeout: 5s
      upstreamStrategy: parallel_best
      resolvers:
        - 1.1.1.1
        - 8.8.8.8
```

config/config.go
```diff
@@ -489,6 +490,8 @@ func unmarshalConfig(data []byte, cfg *Config) error {
 
 func (cfg *Config) migrate(logger *logrus.Entry) bool {
 	usesDepredOpts := Migrate(logger, "", cfg.Deprecated, map[string]Migrator{
+		"upstream":        Move(To("upstreams.groups", &cfg.Upstreams)),
```
0xERR0R, line 493: 👍

docs/configuration.md
```diff
@@ -107,17 +107,18 @@ CIDR notation.
...
-        - 5.9.164.112
-        - 1.1.1.1
-        - tcp-tls:fdns1.dismail.de:853
-        - https://dns.digitale-gesellschaft.ch/dns-query
-      laptop*:
-        - 123.123.123.123
-      10.43.8.67/28:
-        - 1.1.1.1
-        - 9.9.9.9
+    upstreams:
```
0xERR0R, line 110: 👍

0xERR0R (approved):
Looks very good, thanks!

0xERR0R:
> Thanks for already working on the config part.
> 
> What do you think of the idea with both "global" upstream and group specific settings? Instead of the group name being the list of upstream servers they would be under the "resolvers" key. Alternatively we could get rid of the "global" upstream settings and only have group specific setting. What do you think?
> 
> example for both global and group specific configuration:
> 
> ```yaml
> upstreams:
>   timeout: 5s
>   upstreamStrategy: parallel_best
>   otherSettingUsedByAllGroups: xyz
>   groups:
>     <name>:
>       timeout: 5s
>       upstreamStrategy: parallel_best
>       resolvers:
>         - 1.1.1.1
>         - 8.8.8.8
> ```
> 
> example for only group specific configuration:
> 
> ```yaml
> upstreams:
>   groups:
>     <name>:
>       timeout: 5s
>       upstreamStrategy: parallel_best
>       resolvers:
>         - 1.1.1.1
>         - 8.8.8.8
> ```

IMHO, one global configuration is enough. Configuration per group makes it too over-complicated and is useful only for very special edge cases.

ThinkChaos:
We can always add per group config later and be back compatible with a custom `UnmarshalYAML`.

PR #1150, opened 2023-09-17
fix: don't cache EDNS pseudo records

### LINKED ISSUE #1097: COOKIE mismatch
Upstream server responding with a cookie leads to (after the first cached request):
> ;; Warning: Client COOKIE mismatch

Indeed the request is being answered but the output is altered which is not very convenient. Here is what what I mean:
 ```
dig +short -4 rt-01.test1.home @192.168.1.3
;; Warning: Client COOKIE mismatch
192.168.1.1
```

Tested using blocky v0.21 with the following configuration:
```
ports:
  dns: 192.168.1.3:53
  http: 127.0.0.1:8081

log:
  level: debug

queryLog:
  type: console

upstream:
  default:
    - 192.168.1.2
```

Two requests to the upstream DNS server: https://pastebin.com/raw/pVkGCu1x
Requests to blocky - the error appears after the first request: https://pastebin.com/raw/Cvw4wcCr
Here is the log: https://pastebin.com/raw/KHjfQ6av

It seems that the cookie is being cached with the response from the upstream server.
This is causing the warning to be present till the cache expires and after one single request it reappears again.

Would it be possible to introduce a way of changing this behavior?

### PR DESCRIPTION
**Pull request by 0xERR0R**
fixes #1097

### DIFF
--- a/resolver/caching_resolver.go
+++ b/resolver/caching_resolver.go
@@ -215,7 +215,7 @@ func (r *CachingResolver) trackQueryDomainNameCount(domain, cacheKey string, log
 // removes EDNS OPT records from message
 func removeEdns0Extra(msg *dns.Msg) {
 	if len(msg.Extra) > 0 {
-		extra := make([]dns.RR, 0)
+		extra := make([]dns.RR, 0, len(msg.Extra))
 
 		for _, rr := range msg.Extra {
 			if rr.Header().Rrtype != dns.TypeOPT {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by kwitsch — APPROVED** [review]

File: resolver/caching_resolver.go [hunk 1]
```diff
@@ -212,21 +212,41 @@ func (r *CachingResolver) trackQueryDomainNameCount(domain, cacheKey string, log
 	}
 }
 
+// removes EDNS OPT records from message
+func removeEdns0Extra(msg *dns.Msg) {
+	if len(msg.Extra) > 0 {
+		extra := make([]dns.RR, 0)
```
ThinkChaos, line 218: We can pre-alloc the array:
```suggestion
		extra := make([]dns.RR, 0, len(msg.Extra))
```

0xERR0R: Thanks, this solution is much better

PR #1158, opened 2023-09-19
feat: add timeout to bootstrap

0xERR0R:
Bootstrap uses default HttpTransport, which has (IMHO) no timeout defined. I think, it is a good idea to set a timeout.

I'm not so familiar with bootstrap logic, so pleas check it, maybe it is not correct?


resolver/bootstrap.go
```diff
@@ -112,14 +118,16 @@ func (b *Bootstrap) resolveUpstream(r Resolver, host string) ([]net.IP, error) {
...
+		timeout := defaultTimeout
+		if cfg.Upstreams.Timeout.IsAboveZero() {
+			timeout = cfg.Upstreams.Timeout.ToDuration()
 		}
 
+		var cancel context.CancelFunc
+
+		ctx, cancel = context.WithTimeout(ctx, timeout)
+		defer cancel()
+
```
ThinkChaos, line 130: I think we should move the timeout determination to `NewBootstrap` and create `b.dialer` with it.  
That way we can access the potentially user configured timeout via `b.dialer.Timeout` and re-use the dialer as is in `NewHTTPTransport`.

```suggestion
		ctx, cancel := context.WithTimeout(context.Background(), b.dialer.Timeout)
		defer cancel()

```

(Also I think the `var cancel context.CancelFunc` shouldn't be needed even in the previous code and `:=` would work)

resolver/bootstrap.go
```diff
@@ -112,14 +118,16 @@ func (b *Bootstrap) resolveUpstream(r Resolver, host string) ([]net.IP, error) {
 		cfg := config.GetConfig()
 		ctx := context.Background()
```
ThinkChaos, line 120: To go with the other suggestion:
```suggestion
```

resolver/bootstrap.go
```diff
@@ -48,7 +52,9 @@ func NewBootstrap(cfg *config.Config) (b *Bootstrap, err error) {
 		connectIPVersion: cfg.ConnectIPVersion,
 
 		systemResolver: net.DefaultResolver,
-		dialer:         &net.Dialer{},
+		dialer: &net.Dialer{
+			Timeout: defaultTimeout,
```
ThinkChaos, line 56: I unfortunately can't make a suggestion above this block since the lines are unchanged. Can you add the following above the comment?
```go
		timeout := defaultTimeout
		if cfg.Upstreams.Timeout.IsAboveZero() {
			timeout = cfg.Upstreams.Timeout.ToDuration()
		}
```

And then we can use that here:
```suggestion
			Timeout: timeout,
```

resolver/bootstrap.go
```diff
@@ -134,7 +142,13 @@ func (b *Bootstrap) resolveUpstream(r Resolver, host string) ([]net.IP, error) {
...
 func (b *Bootstrap) NewHTTPTransport() *http.Transport {
 	if b.resolver == nil {
-		return &http.Transport{}
+		dialer := &net.Dialer{
+			Timeout: defaultTimeout,
+		}
+
+		return &http.Transport{
+			DialContext: dialer.DialContext,
+		}
```
ThinkChaos, line 151: Let's re-use `b.dialer` so we also use `cfg.Upstreams.Timeout` if it is set:
```suggestion
		return &http.Transport{
			DialContext: b.dialer.DialContext,
		}
```

ThinkChaos (commented):
Great idea, thanks for the implementation!

resolver/bootstrap.go
0xERR0R: Yes it reduces some code duplication and we can reuse dialer. Your suggestion with "b.dialer.Timeout" doesn't work, because dialer is an interface and not net.Dialer. I introduced a new field "timeout" to avoid some "hacks" with casting etc.

resolver/bootstrap.go
0xERR0R: Changed as suggested

resolver/bootstrap.go
0xERR0R: :+1:

resolver/bootstrap.go
ThinkChaos: Ah right, thanks for adjusting!

ThinkChaos (approved):

PR #1159, opened 2023-09-19
chore(text): fix ipv6 representation

0xERR0R:
need to merged before #1153 

Test fails with dns 1.1.56


ThinkChaos (approved):

ThinkChaos:
Meta: if you want to use the auto-merge on approve I think it's a good idea. I just won't approve if I make a suggestion, which was my bad last time :)

0xERR0R:
Yes, it is a good feature, but the button doesn't appear always. There is some logic behindvthe scene🤷

PR #1166, opened 2023-09-20
fix: return error code on server starup failure (#1165)

### LINKED ISSUE #1165: blocky exits with status 0 (success) even when it fails at startup
I noticed this when the listening port was already taken:

```
$ blocky --config /path/to/config.yaml
[...]
[2023-09-20 11:55:23]  INFO server: Starting server
[2023-09-20 11:55:23]  INFO server: http server is up and running on addr/port 4000
[2023-09-20 11:55:23] ERROR server start failed: start tcp listener failed: listen tcp :53: bind: address already in use
$ echo $?
0
```

I think blocky should exit with non-zero status in this case. That should signal the error to systemd, unlike now where systemd thinks the exited blocky process is completely fine/expected.

blocky version: 0.21

EDIT: Pasted wrong log -- this one shows the `address already in use` error.

0xERR0R:
closes #1165

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -19,6 +19,7 @@ import (
 var (
 	done              = make(chan bool, 1)
 	isConfigMandatory = true
+	signals           = make(chan os.Signal, 1)
 )
 
 func newServeCommand() *cobra.Command {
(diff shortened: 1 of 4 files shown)


ThinkChaos (approved):
We could make `done` be a `chan error` to make what it receives matter, but this is perfectly fine :)

PR #1167, opened 2023-09-20
fix: don't cache DNS responses with CD flag

### LINKED ISSUE #455: DNS cache poison of DNSSEC?

0xERR0R:
Don't cache responses with CD (Checking Disabled) flag (DNSSEC)

closes #455

### DIFF
--- a/resolver/caching_resolver.go
+++ b/resolver/caching_resolver.go
@@ -229,6 +229,11 @@ func removeEdns0Extra(msg *dns.Msg) {
 	}
 }
 
+func shouldBeCached(msg *dns.Msg) bool {
+	// we don't cache truncated responses and responses with CD flag
+	return !msg.Truncated && !msg.CheckingDisabled
+}
+
 func (r *CachingResolver) putInCache(cacheKey string, response *model.Response, ttl time.Duration,
 	prefetch, publish bool,
 ) {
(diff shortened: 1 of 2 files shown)


ThinkChaos (approved):
You're on a roll!

PR #1177, opened 2023-09-29
chore: remove unnecessary lock

0xERR0R:


resolver/blocking_resolver.go
```diff
@@ -137,8 +141,11 @@ func NewBlockingResolver(
 			enabled:     true,
 			enableTimer: time.NewTimer(0),
 		},
-		clientGroupsBlock: cgb,
+		clientGroupsBlock: clientGroupsBlock(cfg),
 		redisClient:       redis,
+		fqdnIPCache: expirationcache.NewCache(
+			expirationcache.WithCleanUpInterval[[]net.IP](defaultBlockingCleanUpInterval),
+			expirationcache.WithOnExpiredFn(r.queryForFQIdentifierIPs)),
 	}
```
ThinkChaos, line 149: I was confused by using `r`, which is `nil` at this point, to get the address of `queryForFQIdentifierIPs`.  
IIUC it works because the address of the method is common for all receivers of the same type, so we don't need to deref the receiver to know it. Same as when you call a method on a `nil` receiver.  
But I think initializing this in two steps makes it clearer what our intention is:
```suggestion
	}

	res.fqdnIPCache = expirationcache.NewCache(
			expirationcache.WithCleanUpInterval[[]net.IP](defaultBlockingCleanUpInterval),
			expirationcache.WithOnExpiredFn(res.queryForFQIdentifierIPs))
```

ThinkChaos: When merging this and #1174 together, if we keep the single step init it won't work anymore since [the anonymous function](https://github.com/0xERR0R/blocky/pull/1174/commits/4d6aebaf20300f8acc80327e3e0a0f1b18ee779f#diff-1f338c619cdcf82e75f7ba954b415bd20ac0109f8cbd472cf6a3068d2215bf86R608-R610) capturing `r` would capture the wrong var and you'd call `queryForFQIdentifierIPs` with a `nil` receiver.  
So I think we should definitely change this to prepare for that.

0xERR0R: Yes, you are right. It works because the *r variable is already initialized (pointer). At this point of time it points to nil but at the end of the func it will be initialized. It is better to use the local variable name here.

ThinkChaos (approved):

kwitsch (approved):

0xERR0R:
will be merged after #1174

kwitsch (approved):

PR #1178, opened 2023-09-30
feat: add API endpoint to flush the DNS Cache

### LINKED ISSUE #541: [Feature Request] Add API endpoint to flush the DNS Cache

0xERR0R:
closes #541

### DIFF
--- a/server/server_endpoints.go
+++ b/server/server_endpoints.go
@@ -30,7 +30,6 @@ const (
 	dohMessageLimit   = 512
 	contentTypeHeader = "content-type"
 	dnsContentType    = "application/dns-message"
-	jsonContentType   = "application/json"
 	htmlContentType   = "text/html; charset=UTF-8"
 	yamlContentType   = "text/yaml"
 	corsMaxAge        = 5 * time.Minute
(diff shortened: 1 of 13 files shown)


server/server_endpoints.go
```diff
@@ -57,7 +56,12 @@ func (s *Server) createOpenAPIInterfaceImpl() (impl api.StrictServerInterface, e
 		return nil, fmt.Errorf("no refresh API implementation found %w", err)
 	}
 
-	return api.NewOpenAPIInterfaceImpl(bControl, s, refresher), nil
+	cacheControl, err := resolver.GetFromChainWithType[api.CacheControl](s.queryResolver)
```
ThinkChaos, line 59: This won't clear the bootstrap cache. I don't think that's really necessary, but just wanted to make note of it in case you want to document/implement it.

ThinkChaos (approved):

PR #1179, opened 2023-10-01
Feature/devcontainer

kwitsch:
Since I am currently in the process of switching to devcontainers I defined a first version for blocky.

Current features & configurations:
- predefined defaultFormatter for Go, Markdown & Json
- toolset for Go development
- preinstalled make & Docker
- go linter set to golangci-lint
- go formater set to gofumpt
- autoformat for Go, Markdown & Json enabled

For everyone who has problems building it:
There is a bug in the current release which is fixed in the prerelease.
Setting the extension version in VSCode to prerelease solves the problem.


0xERR0R:
@kwitsch That looks good 👍 I did some experiments with devcontainers and it worked pretty good.I tried to replace the root user with an unprivileged one, this resulted in permission errors and then I gave up. 🤦 I assume, devcontainer runs as root inside, right?


Did you try all commands from the Makefile? I think, we need mkdocs-material for "serve_docs" task.

0xERR0R:
I found the Dockerfile from my last try, maybe we can take the mkdocs installation part from it:
```Dockerfile
FROM golang:1-bookworm

ARG USERNAME=blocky-dev
ARG USER_UID=1000
ARG USER_GID=$USER_UID

# install python package manager
RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && apt-get -y install --no-install-recommends pipx

# create non-root user
RUN groupadd --gid $USER_GID $USERNAME \
    && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
    && apt-get update \
    && apt-get install -y sudo \
    && echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME \
    && chmod 0440 /etc/sudoers.d/$USERNAME	 
	 
USER $USERNAME

ENV PATH="${PATH}:/home/$USERNAME/.local/bin"

# install mkdocs-material
RUN pipx install mkdocs-material --include-deps
```

kwitsch:
> I did some experiments with devcontainers and it worked pretty good.

We switched from IntelliJ to VSCode with devcontainers at work which speed up the onboarding process of new members in projects and ensure coherence among the team.

> I tried to replace the root user with an unprivileged one, this resulted in permission errors and then I gave up. 

According to my experience this should be done after everything else since the other way around can be tricky. 😅

>I assume, devcontainer runs as root inside, right?

Yes. Since the devcontainer is your actual environment it should be considered as "long term work in process". I wanted to create a first step to start development again and enhance the devcontainer in future PRs. Sorry I should have clarified it in the description.

> Did you try all commands from the Makefile? I think, we need mkdocs-material for "serve_docs" task. 

No I only tested it with the common development commands: fmt,generate,lint,test,build,docker-build
I'll add it to completely support make commands.

0xERR0R:
> We switched from IntelliJ to VSCode with devcontainers at work which speed up the onboarding process of new members in projects and ensure coherence among the team.

Just a side note: I think IntelliJ supports devcontainers in the latest version. I didn't try it, but I'll do it as soon we merged you PR :)

kwitsch:
> Just a side note: I think IntelliJ supports devcontainers in the latest version. I didn't try it, but I'll do it as soon we merged you PR :)

It didn't during our evaluation trial and even now it supports less features.
Furthermore we reduced the license costs by canceling the licenses and reinvested roughly the same amount in Copilot licenses for a further performance boost. 😉

@0xERR0R if you haven't already used it: you should have a free Copilot license because of your open source contribution

kwitsch:
@0xERR0R all `make` options are now working inside the devcontainer without errors

> I assume, devcontainer runs as root inside, right?

I have to correct my previous answer, it's actually not.
The user inside the devcontainer is actually called `vscode` which can use `sudo` without a password.
This can be changed in the `devcontainer.json` later on.

ThinkChaos:
I use [direnv](https://github.com/direnv/direnv) to get project dependencies automatically and have a local setup for blocky. It doesn't configure your editor for you, but for VSCode you could commit a `.vscode` folder in the repo with settings and extension recommendations.  
The main advantages are:
- it doesn't depend on a container, so build and load is faster
- it integrates in your shell, so you also get the whole setup just by `cd`-ing into the repo (after marking it as trusted the first time)
- it's not linked to a specific editor (devcontainers are still very VSCode centric AFAIK) and has plugins for most ([VSCode](https://marketplace.visualstudio.com/items?itemName=mkhl.direnv), [IntelliJ/other Jetbrains IDEs](https://plugins.jetbrains.com/plugin/15285-direnv-integration))

If that sounds interesting I can make a PR for you to try it out.

kwitsch:
@ThinkChaos I skimmed over direnv but if I get it correct it's primary usecase are Unix systems.
My development computer runs on Windows at the moment and I really enjoy that devcontainers provide fully configured Linux environments.😅

I currently only have Git, VSCode and Docker installed.
Everything else runs in containers. 🫣

0xERR0R:
Test with Goland 2023.3 (Arch Linux):

- [x] Starting dev container -> downloads 1,17 GB "IDE backend on the remote host" :flushed: 
   - Error in log `[backend] kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 42: Expected start of the object '{', but had '[' instead at path: $.forwardPorts['ports']
JSON input: {"gitConfig":null,"forwardPorts":{"ports":[]}}` -> :shrug: don't know if it is related to devcontainers file or an internal IntelliJ bug
   - terminal window in Goland shows root user :shrug: interesting, in vscode it was non-root user

- [ ] make all -> error, go is not installed

0xERR0R:
Runs fine with vscode. IntelliJ/Goland -> as you mentioned, not all features are supported. Is very hungry (system load 15 on i9 CPU).

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "blocky development",
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-22.04",
+  "features": {
+    "ghcr.io/devcontainers/features/go:1": {
+      "version": "1.21"
```
0xERR0R, line 6: I would try to avoid the fix golang version here, otherwise we need to maintain the current version twice, here and in go.mod. Imho it is better to use the latest 1.x here

0xERR0R (approved):
LGTM, just minor note regarding the golang version. You can merge it :+1:

.devcontainer/devcontainer.json
kwitsch: Sorry had automerge enabled.
Will open another PR for this.

PR #1187, opened 2023-10-06
refactor: pass context for goroutine shutdown

0xERR0R:
- pass context to server/resolvers
- each goroutine checks for cancellation
- I think, maybe we'll need it for config reload -> we can just close the context and recreate all resolvers


cmd/serve.go
```diff
@@ -43,15 +44,19 @@ func startServer(_ *cobra.Command, _ []string) error {
 
 	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
 
-	srv, err := server.NewServer(cfg)
+	ctx, cancelFn := context.WithCancel(context.Background())
```
ThinkChaos, line 47: We can `defer` the `cancelFn` here to make this more foolproof :)

config/config.go
```diff
@@ -331,20 +334,29 @@ func (c *SourceLoadingConfig) StartPeriodicRefresh(refresh func(context.Context)
...
 	defer ticker.Stop()
 
-	for range ticker.C {
-		err := refresh(context.Background())
-		if err != nil {
-			logErr(err)
+	for {
+		select {
+		case <-ticker.C:
+			err := refresh(context.Background())
```
ThinkChaos, line 353:
```suggestion
			err := refresh(ctx)
```

ThinkChaos (commented):
This is great!

kwitsch:
Agreed this looks really great. 👍

cmd/serve.go
0xERR0R: yes, it makes absolutely sense. And we save one line of code :)

ThinkChaos (approved):

kwitsch (approved):

PR #1189, opened 2023-10-06
chore: improve database writer tests

0xERR0R:
I have seen sporadic errors in this test (only on slow environments). I was able to reproduce the error: it is due to the fact that the changes from the DB migration (creation of tables etc) may become "visible" to other sessions a bit later. The first write attempt (insert into the table) goes wrong.

This fix sets the number of connections to 1 and forces the db consumer to reuse the same db connection

### DIFF
--- a/querylog/database_writer_test.go
+++ b/querylog/database_writer_test.go
@@ -34,6 +34,14 @@ var _ = Describe("DatabaseWriter", func() {
 			BeforeEach(func() {
 				writer, err = newDatabaseWriter(sqliteDB, 7, time.Millisecond)
 				Expect(err).Should(Succeed())
+
+				db, err := writer.db.DB()
+				Expect(err).Should(Succeed())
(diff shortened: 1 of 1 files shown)


kwitsch (approved):

ThinkChaos:
I'm thinking we might need to make a similar fix for blocky outside of tests since, IIUC, this situation could happen for someone running blocky for the first time with a DB.  
So maybe we need to flush at the end of migration? Cause the periodic flush starts by waiting, so there's no immediate flush right now. And IMO it would make sense to have an explicit one in the migration function.

0xERR0R:
I'm not sure what flush should do. We use only DML operations (like create table) in the migration. Normal database like postgres will execute it atomar (there is no transaction which can be committed or rollbacked). Due to transaction isolation, this change should be immediately visible.

In test we are using in memory sqlite. I think they made some performance optimisations and therefore this works different.

PR #1191, opened 2023-10-07
opencontainer labels

kwitsch:
Fixed and added opencontainer labels in accordance to: https://github.com/opencontainers/image-spec/blob/main/annotations.md

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,7 +52,7 @@ LABEL org.opencontainers.image.title="blocky" \
   org.opencontainers.image.description="Fast and lightweight DNS proxy as ad-blocker for local network with many features" \
   org.opencontainers.image.url="https://github.com/0xERR0R/blocky#readme" \
   org.opencontainers.image.source="https://github.com/0xERR0R/blocky" \
-  org.opencontainers.image.documentation="https://0xerr0r.github.io/blocky"
+  org.opencontainers.image.documentation="https://0xerr0r.github.io/blocky/${VERSION}/"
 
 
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -41,9 +41,20 @@ RUN --mount=type=bind,target=. \
...
+
+LABEL org.opencontainers.image.title="blocky" \
+  org.opencontainers.image.vendor="0xERR0R" \
+  org.opencontainers.image.licenses="Apache-2.0" \
+  org.opencontainers.image.version=${VERSION} \
+  org.opencontainers.image.created=${BUILD_TIME} \
+  org.opencontainers.image.description="Fast and lightweight DNS proxy as ad-blocker for local network with many features" \
+  org.opencontainers.image.url="https://github.com/0xERR0R/blocky#readme" \
+  org.opencontainers.image.source="https://github.com/0xERR0R/blocky" \
+  org.opencontainers.image.documentation="https://0xerr0r.github.io/blocky"
```
ThinkChaos, line 55: Maybe link to the docs for the specific version?  
Not sure if `VERSION` includes a `v` already or not, so suggestion might be wrong:
```suggestion
  org.opencontainers.image.documentation="https://0xerr0r.github.io/blocky/v${VERSION}/"
```

kwitsch: The v is already in the version string if the tag contains it.
Since this was done in the past we could commit it.

kwitsch:
```suggestion
  org.opencontainers.image.documentation="https://0xerr0r.github.io/blocky/${VERSION}/"
```

ThinkChaos (approved):

PR #1196, opened 2023-10-10
feat: reduce DNS response cache memory consumption

0xERR0R:
currently, we store dns.Msg struct in the DNS response cache. In my tests, a typical response for example "A google.com" needs approx. 270 bytes. If we cache the binary representation (wire format), we'll need only 70 bytes.

This PR changes the result cache struct from dns.Msg to []byte. This avoids also copying of dns struct on one place.

### DIFF
--- a/resolver/caching_resolver.go
+++ b/resolver/caching_resolver.go
@@ -30,7 +30,7 @@ type CachingResolver struct {
 
 	emitMetricEvents bool // disabled by Bootstrap
 
-	resultCache expirationcache.ExpiringCache[dns.Msg]
+	resultCache expirationcache.ExpiringCache[[]byte]
 
 	redisClient *redis.Client
 }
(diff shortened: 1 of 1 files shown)


resolver/caching_resolver.go
```diff
@@ -197,6 +202,24 @@ func (r *CachingResolver) Resolve(request *model.Request) (response *model.Respo
...
+		if err != nil {
+			log.Log().Error("can't unpack cached entry. Cache malformed?", err)
+
+			return nil, 0
+		}
+
+		return res, ttl
+	}
+
+	return nil, 0
```
ThinkChaos, line 220: Change to early return to reduce right-drift, and replaced `log.Log` with `r.log`:
```suggestion
	val, ttl := r.resultCache.Get(key)
	if val == nil {
		return nil, 0
	}

	res := new(dns.Msg)

	err := res.Unpack(*val)
	if err != nil {
		r.log().Error("can't unpack cached entry. Cache malformed?", err)

		return nil, 0
	}

	return res, ttl
```

ThinkChaos (commented):
Seems like a pretty big win!

ThinkChaos (approved):

PR #1202, opened 2023-10-14
feat: configuration to enable/disable external interfaces

### LINKED ISSUE #1109: Configuration to disable/enable of the REST API
I have currently deployed blocky with the following port configuration:
```
ports:
  dns: 53
  tls: 853
  https: 443
```
The [interfaces documentation](https://0xerr0r.github.io/blocky/v0.21/interfaces/) indicates this:
```
If http listener is enabled, blocky provides REST API.
```
Which made me think that disabling `4000` port (http) would also block access to the management API.

But it seems like I'm still able to access the API over `https`.

Given that `https` is required for DNS-over-HTTPS I'm at a loss on how I'd go about blocking access to this management API since I don't want anonymous access to this API.

I could go ahead and put blocky behind a Caddy server and block the management API from there but I was wondering if there's something I'm missing here.

0xERR0R:
closes #1109

### DIFF
--- a/config/interfaces.go
+++ b/config/interfaces.go
@@ -1,0 +1,15 @@
+package config
+
+// InterfacesConfig configuration for blocky's interfaces (REST, DoH, ...)
+type InterfacesConfig struct {
+	Rest RESTConfig `yaml:"rest"`
+	DoH  DoHConfig  `yaml:"dns-over-http"`
+}
+
+type RESTConfig struct {
+	Enable bool `yaml:"enable" default:"true"`
+}
+
+type DoHConfig struct {
+	Enable bool `yaml:"enable" default:"true"`
+}
--- a/config/config.go
+++ b/config/config.go
@@ -212,6 +212,7 @@ type Config struct {
 	Filtering           FilteringConfig           `yaml:"filtering"`
 	Ede                 EdeConfig                 `yaml:"ede"`
 	SUDN                SUDNConfig                `yaml:"specialUseDomains"`
+	Interfaces          InterfacesConfig          `yaml:"interfaces"`
 
 	// Deprecated options
 	Deprecated struct {
--- a/docs/config.yml
+++ b/docs/config.yml
@@ -331,3 +331,12 @@ specialUseDomains:
   # optional: block recomended private TLDs
   # default: true
   rfc6762-appendixG: true
+
+# optional: configuration for external interfaces
+interfaces:
+  rest:
+    # optional: if true and http port is defined, provides REST API to control blocky. Default: true
+    enable: false
+  dns-over-http:
+    # optional: if true and http port is defined, provides DNS-over-HTTP endpoint. Default: true
+    enable: true
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -12,7 +12,7 @@ configuration properties as [JSON](config.yml).
 ## Basic configuration
 
 | Parameter           | Type                | Mandatory | Default value | Description                                                                                                |
-|---------------------|---------------------|-----------|---------------|------------------------------------------------------------------------------------------------------------|
+| ------------------- | ------------------- | --------- | ------------- | ---------------------------------------------------------------------------------------------------------- |
 | certFile            | path                | no        |               | Path to cert and key file for SSL encryption (DoH and DoT); if empty, self-signed certificate is generated |
 | keyFile             | path                | no        |               | Path to cert and key file for SSL encryption (DoH and DoT); if empty, self-signed certificate is generated |
 | dohUserAgent        | string              | no        |               | HTTP User Agent for DoH upstreams                                                                          |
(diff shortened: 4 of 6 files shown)


config/interfaces.go
```diff
@@ -0,0 +1,15 @@
+package config
+
+// InterfacesConfig configuration for blocky's interfaces (REST, DoH, ...)
+type InterfacesConfig struct {
+	Rest RESTConfig `yaml:"rest"`
```
kwitsch, line 5: The option name is a bit misleading as almost all http endpoints could be considered as rest endpoints.
I would suggest `control-api` as name since this option only enables/disables the blocky API.

config/interfaces.go
```diff
@@ -0,0 +1,15 @@
...
+	DoH  DoHConfig  `yaml:"dns-over-http"`
+}
```
kwitsch, line 7: Would it be sensible to add another option like `debug-api` to disable the `/debug` endpoint?
Or do you consider it out of scope and should be added later on?

kwitsch (changes requested):
Since the debug endpoint is technically also rest and not disabled through the rest config option (unless I missed something) the name is misleading.

config/interfaces.go
```diff
@@ -0,0 +1,15 @@
...
+
+type RESTConfig struct {
+	Enable bool `yaml:"enable" default:"true"`
+}
+
+type DoHConfig struct {
+	Enable bool `yaml:"enable" default:"true"`
+}
```
ThinkChaos, line 15: It would be nice to be able to make each HTTP service use a different port: that way you can, for instance, expose DoH on all interfaces but the API only on localhost, and use a firewall as a line of defense without a reverse proxy.
I think that would need more code changes though since there would need to be one `http.Server` for each port, so that can definitely be a follow-up.

kwitsch: Agreed 👍

That would be the cleanest and most secure approach.

Considering changing to an port<->service mapping the config options should move to the ports section or maybe create a new service section? 🤔

0xERR0R: 👍 
Definition of port per services would make this PR obsolete. User can decide, which service should be provided by defining the port, otherwise the service is disabled. 
I think it should also be possible to use the same port for different services. Or should we force to use different ports? 🤔 

`http` configuration will not be used anymore. We have also `https` config property, I think it is too much to provide http and https ports for each service, since only for DoH is https mandatory. I would suggest to use https for DoH automatically and http for other services, but it could be complicated if user wants to use the same port for DoH and REST API.

Something like this: 

oldConfig:
```yaml
ports:
  dns: 53
  tls: 853
  http: 4000
  https: 443
```

new config with different ports
```yaml
ports:
  dns: 53 # plain
  tls: 853 # encrypted
  dns-over-https: 443 # encrypted
  prometheus: 4001 # plain http
  control-api: 4000 # plain http
```

new config with same port (like old behavior)
```yaml
ports:
  dns: 53 # plain
  tls: 853 # encrypted
  dns-over-https: 443 # encrypted
  prometheus: 4001 # plain http
  control-api: 4000 # plain http
```

index.html (links to services and version number) will be removed, instead we can provide a new REST-endpoint in control-API with version number.

Any thoughts?

ThinkChaos: Not sure if we actually want this, but maybe allowing prometheus and the API to use HTTPS would be useful?  
Config could look something like:
```yaml
ports:
  dns:
    plain: 53
    tls: 853
    https: 443
  metrics:
    http: 4001
    https: 443 # would use the same router and server as DoH
  control-api:
    http: 4000
```

Having a section unspecified would mean disabled.  
The old options would need to be migrated to set multiple values which is doable with `Apply`. And we'd probably need to remove the old migration because it would not work consistently since it would depend on the older migration to run before the newer one, but iterating on a `map` is unordered so that would not be guaranteed.

> but it could be complicated if user wants to use the same port for DoH and REST API

Yeah the code would need to re-use the `*chi.Mux` per port+proto combo.  
So maybe build a `map[struct{port, protocol}]*chi.Mux` and then loop over that to create each `net.Listener` and server.

> index.html (links to services and version number) will be removed, instead we can provide a new REST-endpoint in control-API with version number.

The REST-endpoint sounds good, and I have no opinion on `index.html`.

kwitsch: I like the approach with multiple ports per service config wise(clear structure & extensible).
But is there really a scenario where https connection to the metrics or control API is needed?
Since they aren't secured with credentials or API keys they shouldn't be exposed outside of a secured network anyway...🤔

ThinkChaos: Yeah might not be really useful to add `https` options. But even without that, having the protocol in the option name would be future proof and make it more explicit IMO.

kwitsch: Adding quick later on might be interesting. 🤔

kwitsch: There are systems for auto creating certificates in multi container environments which could handle it for you (kubernetes has something build in).

I think unsecured + option to add more is fine.

0xERR0R: > I like the approach with multiple ports per service config wise(clear structure & extensible). But is there really a scenario where https connection to the metrics or control API is needed? Since they aren't secured with credentials or API keys they shouldn't be exposed outside of a secured network anyway...🤔

Maybe we should rename the config from "ports" to "interfaces" and use it for further interface configs (api-key, ...). We can also move the metrics config (prometheus.enable and prometheus.path) there. Something like:

```yaml
interfaces:
  dns:
    ports:
       plain: 53
       tls: 853
       https: 443
  rate-limit: 60 # just example for further config
  metrics:
    ports:
       http: 4001
       https: 443 
    path: /mymetrics # moved from prometheus.path
  control-api:
    ports:
       http: 3000
    api-key: ASBC123 # just example for further config
```

kwitsch: I like this structure the most so far.

ThinkChaos: I like it! Just think `interfaces` could be confusing with "network interfaces" meaning something else. What about `listeners` or `servers`?

kwitsch: Agreed with the confusing part.

How about `services`?

Technically they are services blocky provides...🤔

0xERR0R: I created #1206 to track new implementation. I'll close this PR

PR #1220, opened 2023-10-31
devcontainer & unit test refactoring 

kwitsch:
Devcontainer changes:
- removed slow extensions
- corrected autolinting
- added test extensions
- added lcov converter
- added volume for go package caching
- added task & launch configurations

Unit test changes:
- surrounded improper sections with BeforeEach
- fixed local variables by declaring them in a parent section
- use dedicated ports for each runner if tests are executed parallel 

Workflow changes:
- disabled coverage upload on forks

Makefile changes:
- runs test and race with `-p` if GINKGO_PROCS isn't specified


kwitsch:
Thought of a better way to achieve automatic background testing.
Will push it after testing.

kwitsch:
@0xERR0R & @ThinkChaos I started refactoring the unit tests in this PR to enable parallel execution. Since this is still flaky I disabled it in GitHub workflow execution. The GitHub runners provide 2 cores therefore enabling `-p` speeds `test` and `race` by only <1 minute during workflow execution.
Running `make test` locally with 8< cores the performance increase is significant.

I fixed the most problematic parts but there are still way too many local variables which are against the ginkgo principal (variables shouldn't be declared inside the deepest level). Also initialization should be done outside the test itself in BeforeEach, BeforeAll, etc.

I enabled parallel execution in `make test` and `make race` therefore these are flaky locally which helps identify the problematic tests so they can be fixed gradually.

cmd/serve_test.go
```diff
@@ -12,9 +13,17 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+const (
+	basePort = 5000
```
ThinkChaos, line 16: Not sure if it's happening in practice, but using a predefined port could be a cause of flakiness if something else already uses it, or even a previous blocky run made it unavailable for a bit.  
From a quick look it seems possible to use port 0 and then use a listener which can give you the actual port via `.Addr()`. `http.ListenAndServe` would need to be replaced with creating a server and passing the listener to `.Serve`, like we do in `server.go`.  
I guess making a custom `listenAndServe` func to wrap that would make it easier.

ThinkChaos (approved):
Looks good to me. The port thing doesn't necessarily need to be changed now, do as you wish :)

cmd/serve_test.go
kwitsch: That's why I created a helper function which adds the test process number to the base port.
This way the parallel running tests don't block each others port.

ThinkChaos: Ah my bad I missed that!

ThinkChaos: Turns out I didn't review the full changes, just one push I think!
Too late I guess :stuck_out_tongue:

PR #1221, opened 2023-11-04
feat: add upstream strategy `random`

DerRockWolf:
If you want these changes in separate PRs I'll do so :slightly_smiling_face: 

This PR includes multiple changes:
- removal of the `resolversPerClient` struct used in all upstream group/branch resolvers and replaced by a separate `groupName` string and a flat `upstreamResolverStatus` slice.
  - The `NewXYZResolver()` func now accept the new `UpstreamGroupConfig` struct instead of `UpstreamsConfig`.
  - Doing this allows removal of a lot of redundant code used to select the only existing `upstreamResolverStatus` slice in the map (#1093 implemented a "tree" like architecture which resulted in only one slice existing in the map).
- some fixed comments and logs
- ~~removing the `context.Canceled` check for the `parallel_best` resolver. I'm not 100% certain but from what I can tell the `upstream_resolver` `Resolve()` won't ever return `context.Canceled`.~~ Edit: we keep the check to not miss it when we will pass a context to `upstream_resolver` `Resolve()` func. 
- **And the actual feature**: the new `random` upstream strategy :tada:
  - It's basically the same as `parallel_best` without the race. If the first randomly selected upstream fails a second will be asked.

There are also a few things we could discuss or create issues for:
- Passing a context into `upstream_resolver` `Resolve()` to allow cancellation of the `retry.Do` if e.g. a timeout is reached or the other resolver won the race (for `parallel_best`).
  - Currently if a timeout is reached or once one resolver won the race the `retry.Do` func will still retry any failed calls.

- In #1093 and also in the PR I've used the `Config.Upstreams.Timeout` to "fail" early and continue with the "try next resolver" logic even though `retry.Do` might be able to successfully get an answer on one of the two remaining retries.
    - We could either get rid of the `retry.Do` or multiply the timeout for the contexts I've set by 3 to only continue with the next resolvers if all retries failed.

- The structs returned by `NewXYZResolver` all embed the `Configurable` interface and I'm not 100% sure if this is still needed/useful because the only configuration they currently have is the group name and the `Upstreams` slice.

Thanks for maintaining blocky :heart:


kwitsch:
Looks promising. ♥️

I'll look into it after I fix the parallel processing of unit tests. 🥴

ThinkChaos (commented):
I like the idea and most of the changes are great. I wonder if we could make random a specific parallel best configuration: add an option (could just be in the code for now, not necessarily user facing) of how many resolvers to race.  
Then random would just set that to 1 :)

DerRockWolf:
> I like the idea and most of the changes are great. I wonder if we could make random a specific parallel best configuration: add an option (could just be in the code for now, not necessarily user facing) of how many resolvers to race. Then random would just set that to 1 :)

Good idea, I'll look into it :)

DerRockWolf:
> Good idea, I'll look into it :)

I've now implemented the `random` strategy as custom options for the `parallel_best` resolver.

I'm not 100% happy with the implementation and it still feels a bit rough.
If you have an idea what could be improved please tell me^^

After my changes the random selection is a little bit different than before (I needed to increase the `BeNumerically` range from `80-100` to `80-120`). I could not found out why, from my point of view the changes shouldn't have impact on the randomness, but maybe you have an idea...
 https://github.com/0xERR0R/blocky/blob/0e74a3a5526e431cd36ab65584da4d9ce02402d6/resolver/parallel_best_resolver_test.go#L292-L294

resolver/parallel_best_resolver.go
```diff
@@ -152,89 +134,149 @@ func (r *ParallelBestResolver) Name() string {
...
 
-	go r2.resolve(request, ch)
+	response, collectedErrors := evaluateResponses(ctx, logger, ch, resolvers)
+	if response != nil {
+		return response, nil
+	}
+
+	if !r.retryWithDifferentResolver {
+		return nil, fmt.Errorf("resolution was not successful, used resolvers: %s errors: %v",
+			usedResolvers, collectedErrors)
```
ThinkChaos, line 196:
```suggestion
		return nil, fmt.Errorf("resolution failed: %w", errors.Join(collectedErrors...))
```

resolver/parallel_best_resolver.go
```diff
@@ -152,89 +134,149 @@ func (r *ParallelBestResolver) Name() string {
...
-			collectedErrors = append(collectedErrors, result.err)
-		} else {
-			logger.WithFields(logrus.Fields{
-				"resolver": *result.resolver,
-				"answer":   util.AnswerToString(result.response.Res.Answer),
-			}).Debug("using response from resolver")
+func evaluateResponses(
+	ctx context.Context, logger *logrus.Entry, ch chan requestResponse, resolvers []*upstreamResolverStatus,
+) (*model.Response, []error) {
+	var collectedErrors []error
```
ThinkChaos, line 205: We can pre-allocate the array:
```suggestion
	collectedErrors := make([]error, 0, len(resolvers))
```

resolver/parallel_best_resolver.go
```diff
@@ -152,89 +134,149 @@ func (r *ParallelBestResolver) Name() string {
...
+
+	ch := make(chan requestResponse, 1)
+
+	resolver.resolve(request, ch)
+
+	result := <-ch
+	if result.err != nil {
+		logger.Debug("resolution failed from resolver, cause: ", result.err)
+
+		return nil, errors.New("resolution was not successful, no resolver returned answer in time")
```
ThinkChaos, line 249: I think we're better off forwarding the error to not lose that info in release builds:
```suggestion
		return nil, fmt.Errorf("resolution retry failed: %w", err)
```
```

File: resolver/parallel_best_resolver.go [hunk 4]
```diff
@@ -152,89 +134,149 @@ func (r *ParallelBestResolver) Name() string {
...
+	return result.response, nil
 }
 
-// pick 2 different random resolvers from the resolver pool
-func pickRandom(resolvers []*upstreamResolverStatus) (resolver1, resolver2 *upstreamResolverStatus) {
-	resolver1 = weightedRandom(resolvers, nil)
-	resolver2 = weightedRandom(resolvers, resolver1.resolver)
+// pickRandom picks n (resolverCount) different random resolvers from the given resolver pool
+func pickRandom(resolvers []*upstreamResolverStatus, resolverCount int) (choosenResolvers []*upstreamResolverStatus) {
+	for i := 0; i < resolverCount; i++ {
```
ThinkChaos, line 262: We can pre-alloc too:
```suggestion
	choosenResolvers = make([]*upstreamResolverStatus, 0, resolverCount)

	for i := 0; i < resolverCount; i++ {
```

resolver/parallel_best_resolver.go
```diff
@@ -152,89 +134,149 @@ func (r *ParallelBestResolver) Name() string {
...
+func evaluateResponses(
+	ctx context.Context, logger *logrus.Entry, ch chan requestResponse, resolvers []*upstreamResolverStatus,
+) (*model.Response, []error) {
+	var collectedErrors []error
 
-			return result.response, nil
+	for len(collectedErrors) < len(resolvers) {
+		select {
+		case <-ctx.Done():
+			// this context is currently only set & canceled if resolverCount == 1
```
ThinkChaos, line 210: A bit pedantic but canceled and deadline/timeout aren't the same thing for contexts:
```suggestion
			// this context currently only a deadline when resolverCount == 1
```

resolver/parallel_best_resolver.go
```diff
@@ -152,89 +134,149 @@ func (r *ParallelBestResolver) Name() string {
...
+	if len(r.resolvers) == 1 {
+		logger.WithField("resolver", r.resolvers[0].resolver).Debug("delegating to resolver")
 
-		break
+		return r.resolvers[0].resolver.Resolve(request)
 	}
 
-	if len(resolvers) == 1 {
-		logger.WithField("resolver", resolvers[0].resolver).Debug("delegating to resolver")
+	ctx := context.Background()
```
ThinkChaos, line 157:
```suggestion
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel() // abort requests to resolvers that lost the race
```

resolver/parallel_best_resolver.go
```diff
@@ -152,89 +134,149 @@ func (r *ParallelBestResolver) Name() string {
...
+
+	// using context with timeout for random upstream strategy
+	if r.resolverCount == 1 {
+		var cancel context.CancelFunc
```
ThinkChaos, line 162: Not needed anymore:
```suggestion
```

ThinkChaos (commented):
General state is pretty good, looking forward to being able to add a config for how many resolver to query in parallel!

resolver/bootstrap.go
```diff
@@ -300,16 +300,14 @@ func newBootstrapedResolvers(b *Bootstrap, cfg config.BootstrapDNSConfig) (boots
...
 	resolvers := make([]Resolver, 0, len(br))
 
 	for resolver := range br {
 		resolvers = append(resolvers, resolver)
 	}
 
-	return map[string][]Resolver{
-		upstreamDefaultCfgName: resolvers,
-	}
+	return resolvers
```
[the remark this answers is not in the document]
DerRockWolf: Cool, I didn't know about `maps.Keys()`. I've also found two more occurrences where I could use it.

resolver/bootstrap_test.go
```diff
@@ -32,6 +32,7 @@ var _ = Describe("Bootstrap", Label("bootstrap"), func() {
 	)
 
 	BeforeEach(func() {
+		config.GetConfig().Upstreams.Strategy = config.UpstreamStrategyParallelBest
```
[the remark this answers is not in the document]
DerRockWolf: Good idea, if you want to I could also do this change in a future PR :)

resolver/parallel_best_resolver.go
```diff
@@ -152,89 +134,149 @@ func (r *ParallelBestResolver) Name() string {
 	return r.String()
 }
 
+// TODO: add resolverCount & retryWithDifferentResolver to output
```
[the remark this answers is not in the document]
DerRockWolf: Addressed

resolver/parallel_best_resolver.go
DerRockWolf: What do you mean with previous?
I'm not that familiar with contexts yet, but as far as I know, `cancel` won't result in aborting requests to the lost resolvers, as this context is only used for the `select` `case` and is not passed into any functions.

ThinkChaos: I meant the code currently in main/removed by this PR.
You're right, lets keep the code you wrote for now, and when we use contexts properly add something like this.

resolver/parallel_best_resolver.go
```diff
@@ -153,88 +137,137 @@ func (r *ParallelBestResolver) Name() string {
...
+) (*model.Response, []error) {
+	collectedErrors := make([]error, 0, len(resolvers))
+
+	for len(collectedErrors) < len(resolvers) {
+		select {
+		case <-ctx.Done():
+			// this context currently only has a deadline when resolverCount == 1
+			field := logrus.Fields{"resolver": resolvers[0].resolver}
+			if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+				logger.WithFields(field).Debug("upstream exceeded timeout, trying other upstream")
```
ThinkChaos, line 204: Noticed we can simplify the field passing setup:
```suggestion
			// this context currently only has a deadline when resolverCount == 1
			if errors.Is(ctx.Err(), context.DeadlineExceeded) {
				logger.
					WithField("resolver", resolvers[0].resolver).
					Debug("upstream exceeded timeout, trying other upstream")
```

resolver/parallel_best_resolver.go
```diff
@@ -50,7 +55,7 @@ func newUpstreamResolverStatus(resolver Resolver) *upstreamResolverStatus {
 
 func (r *upstreamResolverStatus) resolve(req *model.Request, ch chan<- requestResponse) {
 	resp, err := r.resolver.Resolve(req)
-	if err != nil && !errors.Is(err, context.Canceled) { // ignore `Canceled`: resolver lost the race, not an error
+	if err != nil {
 		// update the last error time
 		r.lastErrorTime.Store(time.Now())
 	}
```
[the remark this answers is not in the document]
ThinkChaos: With another look, you're definitely right it's not passed to the upstream resolvers. I still think we should still keep this part just to be sure we don't forget it when we pass contexts in the future cause it'd be easy to miss and it's an error path, so the (small) cost of the check doesn't really matter :)

resolver/blocking_resolver.go
```diff
@@ -206,21 +208,11 @@ func (r *BlockingResolver) RefreshLists() error {
...
-
-	for group := range r.cfg.BlackLists {
-		groups[group] = true
-	}
-
-	var result []string
-	for k := range groups {
-		result = append(result, k)
-	}
+	result := maps.Keys(r.cfg.BlackLists)
```
ThinkChaos, line 212: Nice find!

resolver/bootstrap_test.go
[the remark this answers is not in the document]
DerRockWolf: Should I create a new issue to track this or is this already somewhere on your todo list?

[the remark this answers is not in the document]
ThinkChaos: Just made #1258!

ThinkChaos (approved):

ThinkChaos:
Did some tiny changes before merging to remove the TODO in the tests and fix a race because of reading and writing `Upstreams.Timeout` in the tests.  
Also about why the numbers changed: for me the using the old numbers for the parallel tests works, so no change!  
And for the "random" tests, I think it's not an issue it doesn't exactly match parallel best, as long as we keep it consistant-ish from now on.

Thanks for all the work on this!

DerRockWolf:
> Did some tiny changes before merging to remove the TODO in the tests and fix a race because of reading and writing `Upstreams.Timeout` in the tests. Also about why the numbers changed: for me the using the old numbers for the parallel tests works, so no change! And for the "random" tests, I think it's not an issue it doesn't exactly match parallel best, as long as we keep it consistant-ish from now on.
> 
> Thanks for all the work on this!

Thank you for reviewing the PR :smile:

Issue #1222, opened 2023-11-05
Memory usage with large regex-based blocklists


priitohlo:
I'm using Blocky in a container on a Mikrotik RB5009 which is an ARM-based router with 1 GB of memory.
Generally, with plain domain lists (even large ones) it works great but with large regex based lists (e.g. this one: https://big.oisd.nl/regex) I'm seeing that the memory usage of the container starts growing really fast until the device is out of memory.

Is this behaviour with regex-based lists expected and normal or is it fixable in Blocky's code? Am happy to test if need be.

kwitsch:
I looked into it and it seems that the overhead storing the regex as structs accumulate quite a bit.  

There are multiple options to reduce it:
- storing the regex as string and compile them during matching(reduces memory usage / slows matching down)
- fusing all regex to one big(increases load and memory usage during refresh / speed up matching)
- use a different regex engine(most external ones match faster / memory usage has to be evaluated)

0xERR0R:
Using such big regex files (over 200K in this case) has very poor performance: we match each regex in a loop that means, for each request we have to do 200K regex evaluations

priitohlo:
Maybe support for wildcard domains would be an option here? OISD also provides that list in a wildcard format, which Blocky, sadly, does not support.

kwitsch:
> I also thought a bit about this and came up with the same ideas as @kwitsch.
> 
> > - storing the regex as string and compile them during matching(reduces memory usage / slows matching down)
> 
> I think we can eliminate this option from the start, or at least try everything else first, cause that sounds very slow.

Agreed 👍 
I only included it for completeness.

> > - fusing all regex to one big(increases load and memory usage during refresh / reduce long term memory usage)
> > - use a different regex engine(most external ones match faster / memory usage has to be evaluated)
> 
> I think my preferred option would be to combine both of these. We could start with just the first one, and then switch the engine we use to improve performance and hopefully further improve memory usage.  

Thought the same. 👍

> To build the pattern, we should definitely use a `string.Builder` to avoid growing a buffer in a loop and only allocating the giant sting at the end.

Agreed 👍

> Ideally the engine would have an API that lets us assemble the regex so we can avoid handling a `string.Builder` and regex syntax ourselves. The idea being such an API could be optimized to never actually build the full string since the regex engine compiles it to a state machine anyways. The exception being to print the pattern, but seems like a fair tradeoff.  
[...]

PR #1227, opened 2023-11-09
feat(cache): change regex cache to use a single regex to save memory

ThinkChaos:
A quick test w/ https://big.oisd.nl/regex shows memory goes down from 436 MB to 303 MB, which seems stable across runs.

Progress for #1222


ThinkChaos:
Closing because after more testing, I noticed Go's regex library can fail with `error parsing regexp: expression too large`.  
I don't feel like spending time working around that, especially now that I'm almost done with wildcards.

PR #1245, opened 2023-11-14
Unit test refactoring

kwitsch:
Changes:
- `lcov.info` is generated seperately for each subfolder(devcontainer)
- added `dot-import-whitelist` to `golangci.yml` for comonly used global test imports
- moved `mock_call_sequence.go` to `helpertest` package
- consolidated some strings in tests
- updated ginkgo to `v2.13.1`


.devcontainer/scripts/runItOnGo.sh
```diff
@@ -0,0 +1,34 @@
...
+
+FOLDER_NAME=${FOLDER_PATH#"$BASE_PATH/"}
+FILE_NAME="$(echo "$FOLDER_NAME" | sed 's/\//-/g').ginkgo"
+FILE_PATH="/tmp/$FILE_NAME"
+
+
+echo "-- Start $FOLDER_NAME ($(date '+%T')) --"
+
+TIMEFORMAT=' - Ginkgo tests finished in: %R seconds'
+time ginkgo --label-filter="!e2e" --keep-going --timeout=5m --output-dir=/tmp --coverprofile="$FILE_NAME" --covermode=atomic --cover -r -p "$FOLDER_PATH" || true
```
ThinkChaos, line 26: I have a local change to the `Makefile` to set `output-dir` to the equivalent of `${WORKSPACE_FOLDER}/coverage` and just ignore that whole folder in `.gitignore`. Maybe we can do that to avoid putting files in every dir?  
That also seems to match what the `markis.code-coverage` plugin expects as per its [README](https://github.com/markis/vscode-code-coverage#getting-started), so we wouldn't need a custom setting.

kwitsch: Seems like I lost track over all the things I tried to get this features performance up. 😅

I'll change it. 👍

0xERR0R (approved):
LGTM 👍

.devcontainer/scripts/runItOnGo.sh
kwitsch: I moved all lcov files in a coverage folder but had to change the default search pattern since the default one behaved flaky.
All container specifc settings are now in the devcontainer.json.
All generic ones in the settings.json.

kwitsch:
Closes #1251

ThinkChaos (approved):
Looks  like there's conflicts now though!

EDIT: too slow :smile:

PR #1255, opened 2023-11-16
fix: log.privacy option no longer hides domain names being resolved

### LINKED ISSUE #1254: `log.privacy` option no longer hides domain names being resolved
From correlating my Blocky and package manager logs, this seems to have changed after upgrading from 0.20 to 0.21 (here using the [blocky-bin](https://aur.archlinux.org/packages/blocky-bin) AUR package for Arch), and is still the case in 0.22:

```
[2023-08-19T10:28:29+0100] [ALPM] upgraded blocky-bin (0.20-1 -> 0.21-1)
```

Log entry before upgrade:

```
2023-08-19 10:26:53     127.0.0.1       localhost       0       CACHED  * (**.*.*****.***.)     ***** (*.*****.*********.***.), ***** (*****.****.******.***.), * (*.**.***.***), * (*.**.***.***)      NOERROR
```

Log entry after upgrade:

```
2023-08-19 10:48:13     127.0.0.1       localhost       0       CACHED  r3.o.lencr.org. ***** (*.*****.*********.***.), ***** (*****.****.******.***.), * (*.**.***.**), * (*.**.***.**)        NOERROR CACHED  A       server
```

The relevant config:

```yaml
#logLevel: warn
#logPrivacy: true
log:
  level: warn
  privacy: true

queryLog:
  type: csv
  target: /var/log/blocky
```

Looking at the docs I see there's a `queryLog.fields` option, is that now required instead of / as well as `log.privacy`?

0xERR0R:
fixes #1254

found and fixed some log statement where DNS question/answer are logged without obfuscation

### DIFF
--- a/resolver/rewriter_resolver.go
+++ b/resolver/rewriter_resolver.go
@@ -126,9 +126,8 @@ func (r *RewriterResolver) rewriteRequest(logger *logrus.Entry, request *dns.Msg
 			rewritten.Question[i].Name = dns.Fqdn(domainRewritten)
 
 			logger.WithFields(logrus.Fields{
-				"domain":  util.Obfuscate(domainOriginal),
-				"rewrite": rewriteKey + ":" + r.cfg.Rewrite[rewriteKey],
-			}).Debugf("rewriting %q to %q", domainOriginal, domainRewritten)
+				"rewrite": util.Obfuscate(rewriteKey) + ":" + util.Obfuscate(r.cfg.Rewrite[rewriteKey]),
+			}).Debugf("rewriting %q to %q", util.Obfuscate(domainOriginal), util.Obfuscate(domainRewritten))
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


resolver/rewriter_resolver.go
```diff
@@ -126,7 +126,7 @@ func (r *RewriterResolver) rewriteRequest(logger *logrus.Entry, request *dns.Msg
 			rewritten.Question[i].Name = dns.Fqdn(domainRewritten)
 
 			logger.WithFields(logrus.Fields{
-				"domain":  domainOriginal,
+				"domain":  util.Obfuscate(domainOriginal),
 				"rewrite": rewriteKey + ":" + r.cfg.Rewrite[rewriteKey],
```
ThinkChaos, line 130: Should we obfuscate this too? Cause it pretty much contains part of the domain?

0xERR0R: I added additional obfuscation

ThinkChaos (approved):

PR #1261, opened 2023-11-19
refactor: make use of contexts in more places

ThinkChaos:
> - `CacheControl.FlushCaches`
> - `Querier.Query`
> - `Resolver.Resolve`
> 
> Besides all the API churn, this leads to `ParallelBestResolver`,
> `StrictResolver` and `UpstreamResolver` simplification: timeouts only
> need to be setup in one place, `upstreamResolverStatus`.
> 
> We also benefit from using HTTP request contexts, so if the client
> closes the connection we stop processing on our side.

I split the changes to make reviewing easier since most of the changes are just adding `ctx` to function definitions and calls. Here's the interesting commits:
- [refactor: make use of contexts in more places](https://github.com/0xERR0R/blocky/commit/5a01265f60280cb663d019057ce486b60570c1f8) all the interesting `ctx` changes
- [fix(parallel_best): set typeName to "random" when appropriate](https://github.com/0xERR0R/blocky/commit/9af89728c225090fca7a5eae471d9b49d5dbebb7) not related to the `ctx` stuff but did it at the same time
- [refactor: remove deprecated and no-op SingleInflight](https://github.com/0xERR0R/blocky/commit/0dc01c3ee1f75c3027df844cc31d90363631ffbe) also a small but unrelated change

I'd say it's easier to read the commits mentioned above individually, and then scan through the whole lot of changes more quickly.

Progress #1264

### DIFF
--- a/resolver/upstream_resolver.go
+++ b/resolver/upstream_resolver.go
@@ -80,21 +80,18 @@ func createUpstreamClient(cfg config.Upstream) upstreamClient {
 	case config.NetProtocolTcpTls:
 		return &dnsUpstreamClient{
 			tcpClient: &dns.Client{
-				TLSConfig:      &tlsConfig,
-				Net:            cfg.Net.String(),
-				SingleInflight: true,
+				TLSConfig: &tlsConfig,
+				Net:       cfg.Net.String(),
 			},
 		}
 
 	case config.NetProtocolTcpUdp:
 		return &dnsUpstreamClient{
 			tcpClient: &dns.Client{
-				Net:            "tcp",
-				SingleInflight: true,
+				Net: "tcp",
 			},
 			udpClient: &dns.Client{
-				Net:            "udp",
-				SingleInflight: true,
+				Net: "udp",
 			},
 		}
 
--- a/resolver/upstream_resolver.go
+++ b/resolver/upstream_resolver.go
@@ -2,6 +2,7 @@ package resolver
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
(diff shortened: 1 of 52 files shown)


kwitsch:
Since the redis package is also heavily context based and has the somewhat worst implementation(sorry for that 🫣) would you consider also refactoring it?

Or is this out of scope and should be done in a follow up PR?

ThinkChaos:
Yeah there's a couple other places where we should do a pass. I have more changes locally but didn't want to make a too giant PR since this is already big!
Made issue #1264 to track this.

kwitsch (approved):

resolver/upstream_resolver.go
```diff
@@ -169,35 +162,37 @@ func (r *dnsUpstreamClient) fmtURL(ip net.IP, port uint16, _ string) string {
...
-		response, rtt, err = r.tcpClient.Exchange(msg, upstreamURL)
-		if err != nil {
+		response, rtt, err = r.tcpClient.ExchangeContext(ctx, msg, upstreamURL)
+		if err != nil && r.udpClient != nil {
 			// try UDP as fallback
 			var opErr *net.OpError
-			if errors.As(err, &opErr) && opErr.Op == "dial" && r.udpClient != nil {
-				return r.udpClient.Exchange(msg, upstreamURL)
+			if errors.As(err, &opErr) && opErr.Op == "dial" {
+				return r.udpClient.ExchangeContext(ctx, msg, upstreamURL)
```
ThinkChaos, line 174: This changes the behavior: we know have a single timeout for both calls. Meaning if the first one fails because of a timeout, then the second one won't even really start because the context's deadline has been reached.  
I think the best solution would be to start both queries at the same time (or remember which protocol worked last time and use something like weighted random -- but that's more complex and more state).  
That's more work than I want to do in this PR though, so I'll make an issue and get back to it.

In the mean I'll write a test to trigger that condition and do a very ugly thing: ignore the passed-in context for the second query.

kwitsch: How about doubling the context timout and splitting it to different contexts for each request? 🤔

ThinkChaos: > How about doubling the context timout and splitting it to different contexts for each request? 🤔

I'd rather not mess with the timeout value the user provides if we can.

When writing the test, I realized this isn't an issue: the only case we retry is in case of an error during "dial", aka connection refused which shouldn't take long so we'll be fine.  
I still think racing TCP and UDP could be worth it and will create an issue.

Issue #1264, opened 2023-11-21
Moar Go context usage


ThinkChaos:
Let's make a checklist (please edit this post if you have more ideas)!

- [ ] `.golang-ci.yml`: enable context related linters
- [ ] `log/`
  I want to support passing a logger via a context, so we can get rid of `Request.Log` and just have the right logger be chosen by `r.log(ctx)`. Started work on that, but depends on other stuff first.
- [ ] `lists`:
  - [ ] `FileDownloader.DownloadFile`
- [ ] `redis/`
- [ ] `resolver.Resolver` #1261

kwitsch:
>   I want to support passing a logger via a context, so we can get rid of `Request.Log` and just have the right logger be chosen by `r.log(ctx)`. Started work on that, but depends on other stuff first.

That sounds interesting. ♥️

kwitsch:
> `.golang-ci.yml`: enable context related linters


Did this on a dev branch and fell in a rabbit hole of removing `context.Background` from the unit tests 😵‍💫

I'm fairly certain that we fix a lot flakiness from the tests if we actually use the ginkgo contexts...

ThinkChaos:
> I'm fairly certain that we fix a lot flakiness from the tests if we actually use the ginkgo contexts...

Hmm I don't see how it would but could be wrong. What are you thinking?

kwitsch:
> Hmm I don't see how it would but could be wrong. What are you thinking?

Sorry had to be more specific.
I was talking about the e2e-tests which are currently flaky ridden. 😅
The false error runs in most cases looked like a timing problem in the resource allocation and freeing to me.

My theory: since the background context isn't finished after the spec execution it doesn't close services that aren't expected to be in the following spec. But sometimes they are there and produce errors by blocking things.

Background: I tried to remove the `http.Get` with a context aware one and passing the ginkgo context to it.
That made the tests even more flaky...
After I slowly replaced the background context with the ginkgo one it got better.


Do I miss something and it's only a coincidence? 🤔

ThinkChaos:
I'm definitely not that familiar with the e2e tests, but sounds quite plausible!

0xERR0R:
@ThinkChaos @kwitsch Hey, I'd like to make release in the next 1-2 weeks. This is last issue scheduled for 0.23. Is there something in progress and will be implemented or should I change fix version to 0.24?

kwitsch:
[...]

PR #1271, opened 2023-11-24
Refactoring Redis

kwitsch:
Changes:
- config.RedisConfig renamed to config.Redis
- config.Redis got its own file
- config.Redis implements config.Configurable
- config.Redis got unittests
- redis.Client doesn't store it's own context
- methodes of redis.Client expect a context where it's needed
- redis.Client closes its channels when the context is done

Part of #1264

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,6 +8,7 @@
     "source.organizeImports": true,
     "source.fixAll": true
   },
+  "editor.rulers": [120],
   "go.showWelcome": false,
   "go.survey.prompt": false,
   "go.useLanguageServer": true,
@@ -18,8 +19,5 @@
     "ui.semanticTokens": true,
     "formatting.gofumpt": true,
     "build.standaloneTags": ["ignore", "tools"]
-  },
-  "[go]": {
-    "editor.rulers": [120]
   }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -20,6 +20,6 @@
     "build.standaloneTags": ["ignore", "tools"]
   },
   "[go]": {
-    "editor.rulers": [111]
+    "editor.rulers": [120]
   }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -18,5 +18,8 @@
     "ui.semanticTokens": true,
     "formatting.gofumpt": true,
     "build.standaloneTags": ["ignore", "tools"]
+  },
+  "[go]": {
+    "editor.rulers": [111]
   }
 }
--- a/config/redis.go
+++ b/config/redis.go
@@ -1,0 +1,15 @@
+package config
+
+// Redis configuration for the redis connection
+type Redis struct {
+	Address            string   `yaml:"address"`
+	Username           string   `yaml:"username" default:""`
+	Password           string   `yaml:"password" default:""`
+	Database           int      `yaml:"database" default:"0"`
+	Required           bool     `yaml:"required" default:"false"`
+	ConnectionAttempts int      `yaml:"connectionAttempts" default:"3"`
+	ConnectionCooldown Duration `yaml:"connectionCooldown" default:"1s"`
+	SentinelUsername   string   `yaml:"sentinelUsername" default:""`
+	SentinelPassword   string   `yaml:"sentinelPassword" default:""`
+	SentinelAddresses  []string `yaml:"sentinelAddresses"`
+}
--- a/config/redis.go
+++ b/config/redis.go
@@ -1,7 +1,7 @@
 package config
 
 import (
-	"regexp"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 )
--- a/config/redis.go
+++ b/config/redis.go
@@ -53,11 +53,5 @@ func (c *Redis) LogConfig(logger *logrus.Entry) {
 
 // obfuscatePassword replaces all characters of a password except the first and last with *
 func obfuscatePassword(pass string) string {
-	if pass == "" {
-		return ""
-	}
-
-	re := regexp.MustCompile(`^(.).*(.)$`)
-
-	return re.ReplaceAllString(pass, `$1*****$2`)
+	return strings.Repeat("*", len(pass))
 }
--- a/config/redis.go
+++ b/config/redis.go
@@ -1,5 +1,11 @@
 package config
 
+import (
+	"regexp"
+
+	"github.com/sirupsen/logrus"
+)
+
 // Redis configuration for the redis connection
 type Redis struct {
 	Address            string   `yaml:"address"`
@@ -12,4 +18,46 @@ type Redis struct {
 	SentinelUsername   string   `yaml:"sentinelUsername" default:""`
 	SentinelPassword   string   `yaml:"sentinelPassword" default:""`
 	SentinelAddresses  []string `yaml:"sentinelAddresses"`
+}
+
+// IsEnabled implements `config.Configurable`
+func (c *Redis) IsEnabled() bool {
+	return c.Address != ""
+}
+
+// LogConfig implements `config.Configurable`
+func (c *Redis) LogConfig(logger *logrus.Entry) {
+	if len(c.SentinelAddresses) == 0 {
+		logger.Info("Address: ", c.Address)
+	}
+
+	logger.Info("Username: ", c.Username)
+	logger.Info("Password: ", obfuscatePassword(c.Password))
+	logger.Info("Database: ", c.Database)
+	logger.Info("Required: ", c.Required)
+	logger.Info("ConnectionAttempts: ", c.ConnectionAttempts)
+	logger.Info("ConnectionCooldown: ", c.ConnectionCooldown)
+
+	if len(c.SentinelAddresses) > 0 {
+		logger.Info("Sentinel:")
+		logger.Info("  MasterName: ", c.Address)
+		logger.Info("  Username: ", c.SentinelUsername)
+		logger.Info("  Password: ", obfuscatePassword(c.SentinelPassword))
+		logger.Info("  Addresses:")
+
+		for _, addr := range c.SentinelAddresses {
+			logger.Info("    - ", addr)
+		}
+	}
+}
+
+// obfuscatePassword replaces all characters of a password except the first and last with *
+func obfuscatePassword(pass string) string {
+	if pass == "" {
+		return ""
+	}
+
+	re := regexp.MustCompile(`^(.).*(.)$`)
+
+	return re.ReplaceAllString(pass, `$1*****$2`)
 }
--- a/config/redis.go
+++ b/config/redis.go
@@ -28,22 +28,22 @@ func (c *Redis) IsEnabled() bool {
 // LogConfig implements `config.Configurable`
 func (c *Redis) LogConfig(logger *logrus.Entry) {
 	if len(c.SentinelAddresses) == 0 {
-		logger.Info("Address: ", c.Address)
+		logger.Info("address: ", c.Address)
 	}
 
-	logger.Info("Username: ", c.Username)
-	logger.Info("Password: ", obfuscatePassword(c.Password))
-	logger.Info("Database: ", c.Database)
-	logger.Info("Required: ", c.Required)
-	logger.Info("ConnectionAttempts: ", c.ConnectionAttempts)
-	logger.Info("ConnectionCooldown: ", c.ConnectionCooldown)
+	logger.Info("username: ", c.Username)
+	logger.Info("password: ", obfuscatePassword(c.Password))
+	logger.Info("database: ", c.Database)
+	logger.Info("required: ", c.Required)
+	logger.Info("connectionAttempts: ", c.ConnectionAttempts)
+	logger.Info("connectionCooldown: ", c.ConnectionCooldown)
 
 	if len(c.SentinelAddresses) > 0 {
-		logger.Info("Sentinel:")
-		logger.Info("  MasterName: ", c.Address)
-		logger.Info("  Username: ", c.SentinelUsername)
-		logger.Info("  Password: ", obfuscatePassword(c.SentinelPassword))
-		logger.Info("  Addresses:")
+		logger.Info("sentinel:")
+		logger.Info("  master: ", c.Address)
+		logger.Info("  username: ", c.SentinelUsername)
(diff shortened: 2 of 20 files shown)


redis/redis_test.go
```diff
@@ -33,52 +34,52 @@ var _ = Describe("Redis client", func() {
...
-		var rcfg config.RedisConfig
+		var rcfg config.Redis
 		err = defaults.Set(&rcfg)
 
 		Expect(err).Should(Succeed())
 
 		rcfg.Address = redisServer.Addr()
 		redisConfig = &rcfg
-		redisClient, err = New(redisConfig)
+		redisClient, err = New(context.TODO(), redisConfig)
```
ThinkChaos, line 44: You could use the same technique I used for resolvers:
```go
	var (
		ctx       context.Context
		cancelFn  context.CancelFunc
	)

	BeforeEach(func() {
		ctx, cancelFn = context.WithCancel(context.Background())
		DeferCleanup(cancelFn)
	})
```

Or ideally we'd use ginkgo's support for having a context per test, but that only works when calling`New` from a specific test and not in setup nodes ([docs](https://onsi.github.io/ginkgo/#interruptible-node-function-signatures-a-quick-reference)):
```go
It("is interruptible", func (ctx context.Context) {
  ...
})
```

config/redis_test.go
```diff
@@ -0,0 +1,104 @@
...
+				c.LogConfig(logger)
+
+				Expect(hook.Messages).Should(
+					SatisfyAll(ContainElement(ContainSubstring("Address: ")),
+						ContainElement(ContainSubstring("Username: ")),
+						ContainElement(ContainSubstring("Password: ")),
+						ContainElement(ContainSubstring("Database: ")),
+						ContainElement(ContainSubstring("Required: ")),
+						ContainElement(ContainSubstring("ConnectionAttempts: ")),
+						ContainElement(ContainSubstring("ConnectionCooldown: "))))
```
ThinkChaos, line 57: To match the `LogConfig` suggestion:
```suggestion
					SatisfyAll(ContainElement(ContainSubstring("address: ")),
						ContainElement(ContainSubstring("username: ")),
						ContainElement(ContainSubstring("password: ")),
						ContainElement(ContainSubstring("database: ")),
						ContainElement(ContainSubstring("required: ")),
						ContainElement(ContainSubstring("connectionAttempts: ")),
						ContainElement(ContainSubstring("connectionCooldown: "))))
```

config/redis_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+			It("should log sentinel addresses", func() {
+				c.LogConfig(logger)
+
+				Expect(hook.Messages).Should(
+					SatisfyAll(
+						ContainElement(ContainSubstring("Sentinel:")),
+						ContainElement(ContainSubstring("  Addresses:")),
+						ContainElement(ContainSubstring("  - localhost:26379")),
+						ContainElement(ContainSubstring("  - localhost:26380"))))
```
ThinkChaos, line 86: To match the `LogConfig` suggestion:
```suggestion
					SatisfyAll(
						ContainElement(ContainSubstring("sentinel:")),
						ContainElement(ContainSubstring("  addresses:")),
						ContainElement(ContainSubstring("  - localhost:26379")),
						ContainElement(ContainSubstring("  - localhost:26380"))))
```

config/redis_test.go
```diff
@@ -0,0 +1,104 @@
...
+	Describe("obfuscatePassword", func() {
+		When("password is empty", func() {
+			It("should return empty string", func() {
+				Expect(obfuscatePassword("")).Should(Equal(""))
+			})
+		})
+
+		When("password is not empty", func() {
+			It("should return obfuscated password", func() {
+				Expect(obfuscatePassword("test123")).Should(Equal("t*****3"))
```
ThinkChaos, line 100: To match the `obfuscatePassword` suggestion:
```suggestion
				Expect(obfuscatePassword("test123")).Should(Equal("*******"))
```

.vscode/settings.json
```diff
@@ -18,5 +18,8 @@
     "ui.semanticTokens": true,
     "formatting.gofumpt": true,
     "build.standaloneTags": ["ignore", "tools"]
+  },
+  "[go]": {
+    "editor.rulers": [111]
```
ThinkChaos, line 23: Isn't the linter using `120`?

config/redis.go
```diff
@@ -0,0 +1,63 @@
...
+	logger.Info("Required: ", c.Required)
+	logger.Info("ConnectionAttempts: ", c.ConnectionAttempts)
+	logger.Info("ConnectionCooldown: ", c.ConnectionCooldown)
+
+	if len(c.SentinelAddresses) > 0 {
+		logger.Info("Sentinel:")
+		logger.Info("  MasterName: ", c.Address)
+		logger.Info("  Username: ", c.SentinelUsername)
+		logger.Info("  Password: ", obfuscatePassword(c.SentinelPassword))
+		logger.Info("  Addresses:")
```
ThinkChaos, line 46: The other code mostly uses lower-case, and it has the benefit of mostly matching the option names:
```suggestion
		logger.Info("address: ", c.Address)
	}

	logger.Info("username: ", c.Username)
	logger.Info("password: ", obfuscatePassword(c.Password))
	logger.Info("database: ", c.Database)
	logger.Info("required: ", c.Required)
	logger.Info("connectionAttempts: ", c.ConnectionAttempts)
	logger.Info("connectionCooldown: ", c.ConnectionCooldown)

	if len(c.SentinelAddresses) > 0 {
		logger.Info("sentinel:")
		logger.Info("  master: ", c.Address)
		logger.Info("  username: ", c.SentinelUsername)
		logger.Info("  password: ", obfuscatePassword(c.SentinelPassword))
		logger.Info("  addresses:")
```

config/redis.go
```diff
@@ -0,0 +1,63 @@
...
+
+// obfuscatePassword replaces all characters of a password except the first and last with *
+func obfuscatePassword(pass string) string {
+	if pass == "" {
+		return ""
+	}
+
+	re := regexp.MustCompile(`^(.).*(.)$`)
+
+	return re.ReplaceAllString(pass, `$1*****$2`)
```
ThinkChaos, line 62: I think we shouldn't print any of the password:
```suggestion
	return strings.Repeat("*", len(pass))
```

config/redis_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+		When("Address is set", func() {
+			BeforeEach(func() {
+				c.Address = "localhost:6379"
+			})
+
+			It("should log address", func() {
+				c.LogConfig(logger)
+
+				Expect(hook.Messages).Should(ContainElement(ContainSubstring("Address: localhost:6379")))
```
ThinkChaos, line 69: To match the `LogConfig` suggestion:
```suggestion
				Expect(hook.Messages).Should(ContainElement(ContainSubstring("address: localhost:6379")))
```

ThinkChaos (commented):
Have very minor suggestions, looks good and splitting `config.go` more is always great!

.vscode/settings.json
kwitsch: Nope the linter `111` has the default length of 111 characters per line. 😅

The default ruler of VSCode is at character 80 which is misleading if the linter is at 111.

redis/redis_test.go
kwitsch: Missed the part of the docs, I'll read into em.

.vscode/settings.json
kwitsch: Ok I re read the linter docs and the default should be at 120 as you mentioned.

I'm a bit confused why I get linter errors over 111. 🤨

kwitsch: Ok it's getting late...I did read lll as 111 and got confused...🤦‍♂️

It's as you said 120.

I'll change the editor config.

redis/redis_test.go
kwitsch: According to the documentation setup nodes can also use contexts:

"Most Ginkgo nodes can be made interruptible. Setup and Subject nodes typically take a simple func() {} but can be made interruptible like so:"

kwitsch:
@ThinkChaos I rewrote the redis tests to match the ginkgo context aware specs approach. A review would be appreciated since I was struggling quite a bit getting it to work. 🥲

ThinkChaos:
The approach you did looks fine to me. I had gone with the context per describe technique to not have to add initialization to each test and limit the changes I had to do, but I think what you did is what ginkgo actually expects.  
Up to you with how you prefer it, I think as long as the tests work, we're good :)

kwitsch:
I try to stay as close to the ginkgo principal/best practice since they seem to get flaky otherwise when running in parallel.

I'll fix the tests tomorrow when I'm back home. 👍

kwitsch:
Currently uncovered are `ctx.Done()` returns and the sentinel client since I don't know how to test them properly. 🫣

redis/redis.go
```diff
@@ -147,22 +145,33 @@ func (c *Client) PublishEnabled(state *EnabledMessage) {
...
+			response, err := c.getResponse(ctx, iter.Val())
 			if err == nil {
 				if response != nil {
-					c.CacheChannel <- response
+					select {
+					case <-ctx.Done():
+						return
+					default:
+						c.CacheChannel <- response
+					}
```
ThinkChaos, line 174: I think we're better off checking the context first thing in the loop body: here if `getResponse` fails because the context is done, we'll never exit the loop.  
In practice the program will still exit could be an issue if we someday want to recreate the client while blocky is running.

Also to check a context in fewer lines you can do:
```go
if err := ctx.Err(); err != nil {
	// ...
}
```

redis/redis.go
```diff
@@ -213,61 +227,77 @@ func (c *Client) publishMessageFromBuffer(s *bufferMessage) {
...
+
+				return
+			}
+
+			select {
+			case <-ctx.Done():
+				return
 			default:
-				c.l.Warn("Unknown message type: ", rm.Type)
+				c.CacheChannel <- cm
```
ThinkChaos, line 272: GitHub doesn't let me put a suggestion here, but we should do the send in a `case` so it's cancelled if `ctx` ends before it can send:
```go
case c.CacheChannel <- cm:
```

redis/redis.go
```diff
@@ -213,61 +227,77 @@ func (c *Client) publishMessageFromBuffer(s *bufferMessage) {
...
+			}
 
-	err := json.Unmarshal(redisMsg.Message, &msg)
-	if err == nil {
-		c.EnabledChannel <- &msg
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				c.EnabledChannel <- &msg
```
ThinkChaos, line 287: Same as above, the send could be blocking, so let's make sure it cancelable:
```suggestion
			case c.EnabledChannel <- &msg:
```

redis/redis.go
```diff
@@ -213,61 +227,77 @@ func (c *Client) publishMessageFromBuffer(s *bufferMessage) {
...
 }
 
-func (c *Client) processReceivedMessage(msg *redis.Message) {
+func (c *Client) processReceivedMessage(ctx context.Context, msg *redis.Message) {
+	select {
+	case <-ctx.Done():
+		return
+	default:
+	}
+
```
ThinkChaos, line 246: I think we can skip this check: worse case we'll notice the context is done after parsing the json but we're not doing anything potentially blocking:
```suggestion
```

ThinkChaos: And we get rid of the coverage warning :)

ThinkChaos:
If you really want to get coverage for the paths where the context is done, you can call the functions with a already cancelled context, that works in most cases.  
And maybe make the channel send/`ctx.Done()` receive pattern a function and test it directly with a cancelled context.

But yeah it's can be annoying to test and worse case it doesn't have too much impact if broken so I'd say it's not mandatory to get 100%, testing the logic is more important :)

redis/redis.go
kwitsch: I disagree on this one 😅

If getResponse fails the next call would be Next which is also context aware and would terminate the loop.
And if not the redis client would be closed in the startup go routine which would force Next to terminate the loop.

In my opinion the check is at the right place since all other function calls are already context aware.

kwitsch: > Also to check a context in fewer lines you can do:
> ```go
> if err := ctx.Err(); err != nil {
> 	// ...
> }
> ```

I used that approach first but context checks get easily mixed up with other conditions which made rereading the code troublesome.

I like checking it in a select more since it is more obvious when skimming over code.

But that's only my personal preference I guess. 😅

kwitsch:
> If you really want to get coverage for the paths where the context is done, you can call the functions with a already cancelled context, that works in most cases. 

Thought about that but got stuck at `GetRedisCache` since there are surrounding function calls that are context aware it has to be closed at the right time and I couldn't think of a way to get the timing right. 🫣

> And maybe make the channel send/`ctx.Done()` receive pattern a function and test it directly with a cancelled context.

I've already started working on something like that but as part of utility functions and determined it would be out of scope for this PR. So maybe add it later on? 🤔

> But yeah it's can be annoying to test and worse case it doesn't have too much impact if broken so I'd say it's not mandatory to get 100%, testing the logic is more important :)

I like numbers and seeing em drop is discouraging... 😅

ThinkChaos:
> I've already started working on something like that but as part of utility functions and determined it would be out of scope for this PR. So maybe add it later on? 🤔

Yeah for sure can be done later just trying to give ideas for the tests.  
IMO last thing I'd like to see done is just not have any `context.TODO()` in the tests, but if you want to do that later it's fine too.

kwitsch:
> Yeah for sure can be done later just trying to give ideas for the tests.  

Was meant like: I was thinking in the same direction 😉

kwitsch:
I added `util.CtxSend` which provides a context aware way to send something to a channel(should be usefull outside of redis).

Im currently satisfied with the result even if the coverage dropped since there are now some error checks which are troublesome to test but should work(`if err != nil {log & return}`) in theory. 😅

redis/redis.go
ThinkChaos: I personally like the shorter one cause it's not the main logic of the function and it makes it less obtrusive, also forces you to think about the error, but I use the channel when there's another channel in play.  
Both are fine anyways, was more to get the coverage complaints down at the same time!

ThinkChaos (approved):

PR #1274, opened 2023-11-26
ginkgolinter: enable all checks and fix findings

nunnatsa:
ginkgolinter configuration argument names may be a bit misleading. Assuming the linter added to use it, this PR remove all the supress-warning configurations.

Two error checks now found by the linter and fixed as well.


kwitsch (approved):
Thanks for your contribution. 👍

PR #1282, opened 2023-11-29
refactor(tests): simplify some assertions and move some generic cleanup/assertions in helper functions

ThinkChaos:
Continues with the current stream of general cleanup:  
I changed a bunch of assertions to standardize stuff like `Should(Not(x))` to `ShouldNot(x)`, and moved some `DeferCleanup`/`Should(Succeed())` calls in central locations to simplify setup in the actual tests, most notably `TmpFolder`/`TmpFile` no longer need the `Error` field.  
Probably easier to review commit by commit.


querylog/file_writer_test.go
```diff
@@ -117,21 +115,11 @@ var _ = Describe("FileWriter", func() {
 					})
 				})
 
-				Eventually(func(g Gomega) int {
-					filesCount, err := tmpDir.CountFiles()
-					g.Expect(err).Should(Succeed())
-
-					return filesCount
-				}, "20s", "1s").Should(Equal(2))
+				Expect(tmpDir.ReadDir()).Should(HaveLen(2))
```
kwitsch, line 118: I think that will potentially increase flakiness since in theory your approach is correct but will fail on filesystem lags because the directory lookup is performed directly after the second write.

If I remember correctly the `Eventually` was added because it was flaky on the GitHub runner. 🤔

flake.nix
```diff
@@ -0,0 +1,31 @@
+{
```
kwitsch, line 1: Idk anything about NixOS, can't review that 🫣

kwitsch (commented):
Great readability improvement 🧹

One comment and the NixOS part should probably someone else 😅

querylog/file_writer_test.go
ThinkChaos: Yeah `git blame` confirms that, I thought is was because of doing the cleanup in another goroutine.  
I reverted to `Eventually` and decided to remove `TempFolder.ReadDir()` since this was the only use.

kwitsch (approved):

PR #1290, opened 2023-12-03
Bugfix in ECS forward

### LINKED ISSUE #1288: ECS forward is overwritten if mask is set
By looking at this documentation (https://github.com/0xERR0R/blocky/blob/main/docs/configuration.md), I saw that the ecs extension protocol is supported.
However, it is still in an unreleased state, so I compiled a binary blocky myself.

I configured the configuration to enable ECS and when tested via dig the ecs does not seem to be forwarded to the upstream properly. Among them, 5355 is the port of blocky

The dig command is as follows:

```dig a +subnet=218.201.17.0/24 @127.0.0.1 -p 5355 www.taobao.com```

ecs configuration
```
ecs:
   useAsClient: true
   forward: true
   ipv4Mask: 24
   ipv6Mask: 56
```
Execute under normal circumstances

```dig a +subnet=218.201.17.0/24 @127.0.0.1 -p 5355 www.taobao.com```

The obtained ip should be:
111.10.61.149, 111.10.61.150
But actually get:
47.246.2.233, 47.246.2.234

7NLQKLNBCUNPOMYI){0 MKA

### PR DESCRIPTION
**Pull request by kwitsch**
Changes:
- ecs resolver: bugfix for incorrect overrides of ecs option if forward is enabled and mask is set
- ecs resolver test: use ginkgo context instead of `context.Background()`
- devcontainer: set prettier as default formatter for yaml files
- example config: added ecs section

Closes #1288

### DIFF
--- a/resolver/ecs_resolver.go
+++ b/resolver/ecs_resolver.go
@@ -84,6 +84,7 @@ func (r *ECSResolver) setSubnet(so *dns.EDNS0_SUBNET, request *model.Request) {
 	}
 
 	var edsOption *dns.EDNS0_SUBNET
+
 	if ip := subIP.To4(); ip != nil && r.cfg.IPv4Mask > 0 {
 		if mip, err := maskIP(ip, r.cfg.IPv4Mask); err == nil {
 			edsOption = newEdnsSubnetOption(mip, ecsFamilyIPv4, r.cfg.IPv4Mask)

### CODE REVIEW

File: resolver/ecs_resolver.go [hunk 1]
```diff
@@ -53,17 +53,19 @@ func (r *ECSResolver) Resolve(ctx context.Context, request *model.Request) (*mod
 		// Set the client IP from the Edns0 subnet option if the option is enabled and the correct subnet mask is set
 		if r.cfg.UseAsClient && so != nil && ((so.Family == ecsFamilyIPv4 && so.SourceNetmask == ecsMaskIPv4) ||
 			(so.Family == ecsFamilyIPv6 && so.SourceNetmask == ecsMaskIPv6)) {
+			request.Log.Debugf("set client IP from edns0 subnet option: %s", so.Address)
```
ThinkChaos, line 56:
```suggestion
			request.Log.Debugf("using request's edns0 address as internal client IP: %s", so.Address)
```

resolver/ecs_resolver.go
```diff
@@ -73,25 +75,24 @@ func (r *ECSResolver) Resolve(ctx context.Context, request *model.Request) (*mod
...
-			e.Address = mip
+func (r *ECSResolver) setSubnet(so *dns.EDNS0_SUBNET, request *model.Request) {
+	var subIP net.IP
+	if so != nil && r.cfg.Forward && so.Address != nil {
+		subIP = so.Address
+	} else {
+		subIP = request.ClientIP
+	}
+
+	request.Log.Debugf("set edns0 subnet option: %s", subIP)
```
ThinkChaos, line 86: I think this might be confusing for users because it's not logging the masked IP and the message doesn't make that fully clear.

Maybe we can do something like:
```go
var res *dns.EDNS0_SUBNET

if ... {
	res = ...
else if ... {
	res = ...
}

if res != nil {
	// log w/ res.Address
	util.SetEdns0Option(request.Req, res)
}
```

resolver/ecs_resolver.go
```diff
@@ -73,25 +75,24 @@ func (r *ECSResolver) Resolve(ctx context.Context, request *model.Request) (*mod
...
+
+	if ip := subIP.To4(); ip != nil && r.cfg.IPv4Mask > 0 {
+		if mip, err := maskIP(ip, r.cfg.IPv4Mask); err == nil {
+			e := getEdnsSubnetOption(mip, ecsFamilyIPv4, r.cfg.IPv4Mask)
```
ThinkChaos, line 90: Want to avoid `get` meaning "create":
```suggestion
			e := newEdnsSubnetOption(mip, ecsFamilyIPv4, r.cfg.IPv4Mask)
```

resolver/ecs_resolver.go
```diff
@@ -73,25 +75,24 @@ func (r *ECSResolver) Resolve(ctx context.Context, request *model.Request) (*mod
...
 		}
-	} else if ip := request.ClientIP.To16(); ip != nil && r.cfg.IPv6Mask > 0 {
-		mip, err := maskIP(ip, r.cfg.IPv6Mask)
-		if err == nil {
-			e.Family = ecsFamilyIPv6
-			e.SourceNetmask = uint8(r.cfg.IPv6Mask)
-			e.Address = mip
+	} else if ip := subIP.To16(); ip != nil && r.cfg.IPv6Mask > 0 {
+		if mip, err := maskIP(ip, r.cfg.IPv6Mask); err == nil {
+			e := getEdnsSubnetOption(mip, ecsFamilyIPv6, r.cfg.IPv6Mask)
```
ThinkChaos, line 95:
```suggestion
			e := newEdnsSubnetOption((mip, ecsFamilyIPv6, r.cfg.IPv6Mask)
```

resolver/ecs_resolver.go
```diff
@@ -103,3 +104,14 @@ func maskIP[maskType ECSMask](ip net.IP, mask maskType) (net.IP, error) {
 
 	return mip.IP, err
 }
+
+// getEdnsSubnetOption creates a new EDNS0 subnet option with the given IP, family and mask
+func getEdnsSubnetOption[maskType ECSMask](ip net.IP, family uint16, mask maskType) *dns.EDNS0_SUBNET {
```
ThinkChaos, line 109:
```suggestion
// newEdnsSubnetOption( creates a new EDNS0 subnet option with the given IP, family and mask
func newEdnsSubnetOption([maskType ECSMask](ip net.IP, family uint16, mask maskType) *dns.EDNS0_SUBNET {
```

ThinkChaos (commented):
That was fast since the report!

Just a couple minor things for clarity, mostly relating to the debug logs.

ThinkChaos (approved):

PR #1302, opened 2023-12-13
feat(resolver): race UDP and TCP when connecting upstream

### LINKED ISSUE #1266: Race TCP and UDP upstream connections instead of trying sequentially
Right now when a client uses TCP, we try TCP and then UDP if that failed to connect. I think we should race both connections and use the winner (obviously if the UDP response wins but is truncated we shouldn't use it). That's similar in concept to the [Happy Eyeballs algorithm](https://en.wikipedia.org/wiki/Happy_Eyeballs).  
This would also have the benefit of falling back to UDP for any kind of failure, not just connection refused. I think this could make a difference in practice since a lot of firewalls are set to `DROP` by default.

At the same time, we should investigate and fix the inconsistency between `dnsUpstreamClient.callExternal` and `createUpstreamClient`: the later always creates the struct with both `tcpClient` and `udpClient`, but `callExternal` seems to have been written before that was the case as it does `nil` checks.

See https://github.com/0xERR0R/blocky/pull/1261#discussion_r1400919635

ThinkChaos:
Inspired by https://en.wikipedia.org/wiki/Happy_Eyeballs this should improve latency and fixes the long standing behavior where a single resolve attempt could take 2x the timeout.
UpstreamResolver.Resolve can still take more than the configured timeout so maybe that can be improved by splitting the retry algorithm into its own resolver type.

Fixes #1266

### DIFF
--- a/resolver/upstream_resolver.go
+++ b/resolver/upstream_resolver.go
@@ -188,24 +188,71 @@ func (r *dnsUpstreamClient) fmtURL(ip net.IP, port uint16, _ string) string {
 func (r *dnsUpstreamClient) callExternal(
 	ctx context.Context, msg *dns.Msg, upstreamURL string, protocol model.RequestProtocol,
 ) (response *dns.Msg, rtt time.Duration, err error) {
-	if protocol == model.RequestProtocolTCP {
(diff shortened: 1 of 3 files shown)


0xERR0R (approved):
👍

PR #1316, opened 2023-12-21
Refactoring/e2e tests

### LINKED ISSUE #1224: add "withNetwork" for testcontainers to e2e-implementation
With following PR, testcontainers introduced "withNetwork" function. We should use it instead of own function.

see https://github.com/testcontainers/testcontainers-go/pull/1894

kwitsch:
Changes:
- removed tmpDir in favor of individual testfiles
- replaced `WithNetwork` with a modified version of the [testcontainer PR](https://github.com/testcontainers/testcontainers-go/pull/1894)
- moved some functions to different files to declutter `containers.go`

I opted for a modified version since the original one doesn't remove the network after the tests finish and also use `context.Background()`(which could be avoided with the Ginkgo context).

Behavior change to ensure only containers for the current test are inside the network:
- it is created during the first `WithNetwork` costomization
- it is deleted after the last container detaches
- during deletion a new network name is generated

Closes #1224

### DIFF
--- a/e2e/basic_test.go
+++ b/e2e/basic_test.go
@@ -25,7 +25,7 @@ var _ = Describe("Basic functional tests", func() {
 		})
 		When("wrong port configuration is provided", func() {
 			BeforeEach(func(ctx context.Context) {
-				blocky, err = createBlockyContainer(ctx, tmpDir,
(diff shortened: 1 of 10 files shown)


0xERR0R (approved):
👍

PR #1326, opened 2024-01-07
Use a variable to select the job in the grafana dashboard

shizunge:
Do not hardcode the job name in the grafana dashboard.

User cannot set the job name in Grafana Cloud. (non self hosted prometheus)

### DIFF
--- a/docs/blocky-grafana.json
+++ b/docs/blocky-grafana.json
@@ -160,7 +160,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "exemplar": false,
-          "expr": "sum(up{job=\"blocky\"})",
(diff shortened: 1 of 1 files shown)


0xERR0R (approved):

0xERR0R:
Thank you!

FutureCow:
Tried this new JSON, but for me it's not working. This is my grafana page:
image
When I change $job to blocky, I can see the values again.
Also get an error when pushing one of the buttons, but dont know if that's related.

FutureCow:
I'm running the latest stable: Grafana v10.3.3

Grafana is running native on the system
Blocky: v0.23-44-g07b864e inside docker container
Prometheus: version="2.45.3+ds" running native on the system

PR #1331, opened 2024-01-09
Fix docs at index being outdated

### LINKED ISSUE #1329: Documentation pages at index (`/*`) are outdated
Hello, I started blocky and I got a few warnings that my yaml configuration is deprecated :

```
WARN config option "upstream" is deprecated, please use "upstreams.groups" instead
WARN config option "blocking.downloadTimeout" is deprecated, please use "blocking.loading.downloads.timeout" instead
WARN config option "blocking.refreshPeriod" is deprecated, please use "blocking.loading.refreshPeriod" instead
WARN config option "hostsFile.refreshPeriod" is deprecated, please use "hostsFile.loading.refreshPeriod" instead
WARN config option "hostsFile.filePath" is deprecated, please use "hostsFile.sources" instead
```
The problem is : the documentation of the "reference file" ( https://0xerr0r.github.io/blocky/configuration/ ) hasn't been updated, could you update your documentation please ?

ThinkChaos:
`/configuration` will redirect to `/latest/configuration` which actually gets updated by the CI.

Fix #1329


0xERR0R:
Branch gh-pages will be overridden on next run of the regular docs job.

ThinkChaos:
Do you mean manual commits would be lost?
I've tested and that's not the case, see for instance a65bb21710b1642acd6f9a270f5ba3dac94cfd40. The CI does pull + push every time.

If you mean the contents of the files, then these ones are not being updated by the CI, and `mike` seems to not support files at the index, so it's not something we can update via the CI without hacks. Thus this adding a redirect to `/latest`, which does get updated.  
I could've missed something with mike, but I really don't see any other solution to this, and unfortunately it looks like we don't have git history from before the "single dev branch" change so, I'm not sure how these files were added in the first place.

If you have another idea for a solution, I'm all for it, but I know this will work (fully tested on my fork, including tagging a new release after this is merged) and really don't see any other way.

0xERR0R:
I'm not against this solution, I just have concerns that your work will be lost. I'm not sure, but my understanding was: documentation source is the main branch (docs directory). CI job creates all needed files and pushes them to gh-pages branch. Some file manipulation on gh-pages branch manually will be overridden (I think the job will do some kind of git push --force if there are some conflicts).

We can try to merge it and we can then observe what happens in the future.

0xERR0R (approved):

ThinkChaos:
Yeah that happens to files that are under `latest/` since the CI generates them every time, but since the files at the root are never generated by the CI they'll stay as is.

ThinkChaos:
Redirect works fine: the link in the repo sidebar (https://0xerr0r.github.io/blocky/) now redirects to, ATM, https://0xerr0r.github.io/blocky/v0.23/
Which has a versions drop down the old page didn't have since it was generated once and never updated, thus only knew about one version, itself, and didn't include the dropdown. Anyways good news is old docs are easily accessible for users now :)

PR #1352, opened 2024-01-26
feat: support CNAME records in customDNS mappings

BenMcH:
This PR is related to #465. Currently, I have only implemented CNAME records, but it would be simple to add additional record types mentioned in #996. 

Please let me know if you'd prefer a different approach, but this PR being merged would allow me to replace pihole with Blocky in my home network. I'm happy to add more record types if they make sense. My understanding of DNS is that most other record types typically have more complex use-cases that may not make sense to implement, as they're usually reserved for serving data such as mail configuration or DNS ownership, neither of which I would expect Blocky to be utilized for, given its local-network scope and risk associated with hosting a publicly accessible DNS server.

Further DNS response types can be added with a simple if statement in both the custom_dns.go and custom_dns_resolver.go.


ThinkChaos (commented):
Thanks for picking this up!

I've got a couple comments about details, but also wanted to bring up maybe we should use [dns.ZoneParser](https://pkg.go.dev/github.com/miekg/dns#ZoneParser) which supports the standard zonefile format instead of making something custom? `UnmarshalYAML` could error if any RR is not an `A`/`AAAA`/`CNAME`.  
That also would have the benefit of allowing to split the mapping into another file via `$INCLUDE`.  
That shouldn't be a blocker for this PR though.

kwitsch:
Agreed that the zone parser looks promising but is out of scope and could be implemented later on.

config/custom_dns.go
```diff
@@ -16,14 +17,13 @@ type CustomDNS struct {
 	FilterUnmappedTypes bool             `yaml:"filterUnmappedTypes" default:"true"`
 }
 
-// CustomDNSMapping mapping for the custom DNS configuration
 type CustomDNSMapping struct {
-	HostIPs map[string][]net.IP `yaml:"hostIPs"`
+	Entries map[string][]dns.RR `yaml:"entries"`
 }
```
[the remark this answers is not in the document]
BenMcH: I'm taking a look at this and don't mind cleaning up while I'm in here :smile: Should input be a string? I was thinking it should be a `map[string]string`. 

I need to run some errands, but I'll check back later and finish cleaning this up then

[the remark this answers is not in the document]
BenMcH: Nevermind, I see what I'm doing wrong! I was implementing CustomDNSMapping, not Entries.

resolver/custom_dns_resolver.go
```diff
@@ -120,26 +137,45 @@ func (r *CustomDNSResolver) processRequest(request *model.Request) *model.Respon
 		}
 	}
 
-	return nil
+	logger.WithField("next_resolver", Name(r.next)).Trace("go to next resolver")
+	forwardResponse, _ := r.next.Resolve(ctx, request)
```
ThinkChaos, line 141: We need to return this error, can you adjust the function signatures? `processCNAME` will need to change too.

resolver/custom_dns_resolver.go
```diff
@@ -85,12 +92,22 @@ func (r *CustomDNSResolver) processRequest(request *model.Request) *model.Respon
...
 		if found {
-			for _, ip := range ips {
-				if isSupportedType(ip, question) {
-					rr, _ := util.CreateAnswerFromQuestion(question, ip, r.cfg.CustomTTL.SecondsU32())
-					response.Answer = append(response.Answer, rr)
+			for _, entry := range entries {
+				switch v := entry.(type) {
+				case *dns.A:
+					ip := net.ParseIP(v.A.String())
+					result := r.processIP(ip, question)
```
ThinkChaos, line 101: Sorry if it wasn't clear but I mean directly passing the IP without going through a string:
```suggestion
					result := r.processIP(v.A, question)
```

resolver/custom_dns_resolver.go
```diff
@@ -85,12 +92,22 @@ func (r *CustomDNSResolver) processRequest(request *model.Request) *model.Respon
...
+					response.Answer = append(response.Answer, result...)
+				case *dns.AAAA:
+					ip := net.ParseIP(v.AAAA.String())
+					result := r.processIP(ip, question)
```
ThinkChaos, line 105:
```suggestion
					result := r.processIP(v.AAAA, question)
```

resolver/custom_dns_resolver.go
```diff
@@ -120,26 +137,45 @@ func (r *CustomDNSResolver) processRequest(request *model.Request) *model.Respon
...
+	cname := new(dns.CNAME)
+	cname.Hdr = dns.RR_Header{Class: dns.ClassINET, Ttl: r.cfg.CustomTTL.SecondsU32(), Rrtype: dns.TypeCNAME, Name: question.Name}
+	cname.Target = dns.Fqdn(targetCname.Target)
+	result = append(result, cname)
+
+	targetWithoutDot := strings.TrimSuffix(targetCname.Target, ".")
+
+	// Resolve target recursively
+	targetResp := r.processRequest(ctx, newRequestWithClientID(targetWithoutDot, dns.Type(dns.TypeA), request.ClientIP.String(), request.RequestClientID))
+	result = append(result, targetResp.Res.Answer...)
```
ThinkChaos, line 178: We'll need to do the same thing for `dns.TypeAAAA` here.

resolver/custom_dns_resolver.go
```diff
@@ -120,26 +137,45 @@ func (r *CustomDNSResolver) processRequest(request *model.Request) *model.Respon
...
-		}
+	resp := r.processRequest(ctx, request)
+	return resp, nil
+}
+
+func (r *CustomDNSResolver) processIP(ip net.IP, question dns.Question) (result []dns.RR) {
+	result = make([]dns.RR, 0)
+
+	if isSupportedType(ip, question) {
+		rr, _ := util.CreateAnswerFromQuestion(question, ip, r.cfg.CustomTTL.SecondsU32())
```
ThinkChaos, line 161: I don't think this error will ever happen, but since we're changing the other functions to return an error, can you  check and return it if it occurs?

resolver/custom_dns_resolver.go
```diff
@@ -85,12 +92,22 @@ func (r *CustomDNSResolver) processRequest(request *model.Request) *model.Respon
 	domain := util.ExtractDomain(question)
 
 	for len(domain) > 0 {
```
ThinkChaos, line 94: Let's add a context cancellation check here. That way if we're in an infinite CNAME loop, we'll have an exit strategy:
```suggestion
	for len(domain) > 0 {
		if err := ctx.Err(); err != nil {
			return nil, err
		}

```

We also need another change to actually trigger it, since ATM we never cancel the context.  
Can you also add a timeout in `Server.resolve`, just above the `switch`?
```go
	ctx, cancel := context.WithTimeout(ctx, 100*s.cfg.Upstreams.Timeout.ToDuration())
	defer cancel()

	switch {
		// ...
```

That way we'll reduce the impact of a bad config, and generally having a high global timeout per request is a good safety measure.

ThinkChaos (commented):
A couple new minor things and some follow-up from the previous threads.
Progress is looking great :)

BenMcH:
I think that I've implemented most of the PR feedback. I'm not a regular golang user and have been having a heck of a time implementing the `CustomDNSEntries` unmarshalling. I didn't want to force the rest of the changes to come after that, so I skipped it for now. I'll keep hacking away at it on and off tonight, but if you feel this is ready to ship, I can do a follow-up PR once I figure it out.

ThinkChaos:
If you want to post something that's blocking you feel free and I can take a look :)

I resolved most threads, there's just that and one error in `processIP` left, and then I think you'll get some trouble from the linter.

BenMcH:
> If you want to post something that's blocking you feel free and I can take a look :)

I'm mostly just not sure how to get the nested unmarshalling working. I took the sample posted earlier and tried to run with it. Conceptually, the code makes sense to me, however, when I run it, the UnmarshallYAML function gets called and I'm getting empty strings as the input. I'm sure that I've simply mapped something incorrectly. 

Here's a gist of the error: https://gist.github.com/BenMcH/32137efc5f5d5f0ce9f33bc94ccda796

> I resolved most threads, there's just that and one error in `processIP` left, and then I think you'll get some trouble from the linter.

Sorry for my ignorance, but which error in processIP am I missing? I thought that I had updated all of the calls that could return an error to bubble the error up and out of the function.

I'll take a look at the linter soon! :rocket:

ThinkChaos:
> I'm mostly just not sure how to get the nested unmarshalling working.

The code you posted works for me!  
I see one potential issue: the `fmt.Println("Here", input)` runs before `unmarshal(&input)` so it'll see `input` as an empty string. So maybe it's your debugging that's misleading you?  
And another minor thing, but that shouldn't break anything: `result := make(CustomDNSEntries, len(input))` should be `result := make(CustomDNSEntries, len(parts))`.

> Sorry for my ignorance, but which error in processIP am I missing? I thought that I had updated all of the calls that could return an error to bubble the error up and out of the function.

The one from this thread: https://github.com/0xERR0R/blocky/pull/1352#discussion_r1468227613
`rr, _ := ...` should become `rr, err := ...`, and then do the usual `err != nil` dance.

BenMcH:
Thank you for all of your help on this, @ThinkChaos! Turns out I was overthinking it and failed to assign `*c` to result at the end, which made my DNS queries fail :man_facepalming: 

I'll go ahead and implement the rest of the feedback tonight. 

I think that the processIPs change is implemented. The current diff has the following:

```
rr, err := util.CreateAnswerFromQuestion(question, ip, r.cfg.CustomTTL.SecondsU32())
if err != nil {
	return nil, err
}
```

BenMcH:
Alright, `make lint` is now happy locally. I think the tests should be fine as well, but I could be wrong. When I build from `main` or my branch, reverse DNS is not functional, which could probably be attributed to something in my home network or my blocky config (I don't typically use reverse lookups, so this is the first I've tried them at home).

The reverse DNS and old log file deletion tests are failing on both branches on my machine, so I don't _think_ that I introduced it. If I did introduce them, let me know, and I can continue looking into it.

BenMcH:
Seems I just had to walk away and come back with fresh eyes. I was able to fix the reverse DNS tests and now both `make lint` and `make test` work locally!

BenMcH:
I went ahead and added a variety of tests throughout my branch to validate functionality this PR has added and to hopefully surpass the codecov failures. :crossed_fingers:

BenMcH:
Alright, I put the create answer function on the custom resolver and made a setter func, like I saw in the NextResolver struct and forced an error for the create answer path. My local coverage.html file now shows that the file is 100% covered :grin:

docs/configuration.md
```diff
@@ -278,11 +278,14 @@ domain must be separated by a comma.
       mapping:
         printer.lan: 192.168.178.3
         otherdevice.lan: 192.168.178.15,2001:0db8:85a3:08d3:1319:8a2e:0370:7344
+        anothername.lan: CNAME(otherdevice.lan)
     ```
 
 This configuration will also resolve any subdomain of the defined domain, recursively. For example querying any of
 `printer.lan`, `my.printer.lan` or `i.love.my.printer.lan` will return 192.168.178.3.
 
+CNAME records are supported by setting the value of the mapping to `CNAME(target)`. Note that the target will be recursively resolved and there are no checks for CNAME loops, meaning that it is possible for CNAMES to never resolve due to a loop.
```
kwitsch, line 287: No breaker but just an idea:
If every CNAME that is looped through is added to a slice you could check if it already was in the resolving CNAME chain and break the loop with an error. 🤔

kwitsch (approved):
Looks good to me.

One small comment how to possibly prevent CNAME loops.

Nice work & communication. 👍

docs/configuration.md
BenMcH: Great idea! I just pushed up a change to add this to the custom resolver!

BenMcH: But it broke the linter :man_facepalming: I'll have that fixed in a few mintues

resolver/custom_dns_resolver.go
```diff
@@ -120,26 +144,113 @@ func (r *CustomDNSResolver) processRequest(request *model.Request) *model.Respon
...
+
+	result = append(result, targetResp.Res.Answer...)
+
+	// Resolve ipv6 target recursively
+	targetResp, resolvedCnames, err = r.processRequestTrackingCnames(ctx, aaaaRequest, append(resolvedCnames, targetWithoutDot))
+	if err != nil {
+		return nil, err
+	}
+
+	result = append(result, targetResp.Res.Answer...)
```
ThinkChaos: Sorry to tack this on last minute, but the logic needs to be adjusted. My bad for not realizing earlier and suggesting the AAAA thing.

Based on [RFC-1034 Section 3.6.2](https://www.rfc-editor.org/rfc/rfc1034#section-3.6.2):
- we don't need to do both A and AAAA, but should use what the question asks for. Looking at what `RewriterResolver.rewriteRequest` does should be a good starting point
- if the question is for a CNAME, then we must not recurse but send the CNAME itself
- a CNAME RR must be the only one for a domain

For the last point I'm not sure what's the best thing to do if the user's config doesn't respect that. Here's a couple options:
- error in `UnmarshalYAML`, the user will notice immediately since they're presumably editing and restarting blocky
- log an error and return a ServFail response. EDE, if enabled, could point the user in the right direction.
- discard the extra RRs and just return the result from the CNAME query

resolver/custom_dns_resolver.go
```diff
@@ -120,26 +144,113 @@ func (r *CustomDNSResolver) processRequest(request *model.Request) *model.Respon
...
+	switch v := entry.(type) {
+	case *dns.A:
+		return r.processIP(v.A, question)
+	case *dns.AAAA:
+		return r.processIP(v.AAAA, question)
+	case *dns.CNAME:
+		return r.processCNAME(ctx, request, *v, resolvedCnames, question)
+	}
+
+	return nil, fmt.Errorf("unsupported type %T", entry)
```
ThinkChaos, line 192: Trying to make what it means obvious even without the context of the code:
```suggestion
	return nil, fmt.Errorf("unsupported customDNS RR type %T", entry)
```

ThinkChaos (changes requested):
Can you delete the `blocky` binary and add it to `.gitignore`?

resolver/custom_dns_resolver.go
```diff
@@ -75,7 +92,7 @@ func (r *CustomDNSResolver) handleReverseDNS(request *model.Request) *model.Resp
 	return nil
 }
 
-func (r *CustomDNSResolver) processRequest(request *model.Request) *model.Response {
+func (r *CustomDNSResolver) processRequestTrackingCnames(ctx context.Context, request *model.Request, resolvedCnames []string) (*model.Response, []string, error) {
```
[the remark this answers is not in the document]
BenMcH: You're correct. we can remove it from the return signature!

resolver/custom_dns_resolver.go
BenMcH: The other changes requested have been pushed. I'll start taking a look at this one now, as it seems like a bigger change than the others

BenMcH: Would it be correct to say that a domain could have multiple CNAME RRs associated but that all RRs should be CNAMEs if any one is a CNAME? 

I believe that it is valid to have 2 CNAMEs for a single domain, pointing to different targets.

BenMcH: Looks like my understanding is wrong. Only a single CNAME value is valid for a given domain. I'll add a check to the UnmarshalYAML function so that users are immediately forced to correct the error rather than blindly ignoring configuration.

ThinkChaos: > Only a single CNAME value is valid for a given domain

Yeah that's my understanding too!

BenMcH:
@ThinkChaos thanks for the thorough review! I think that I've implemented all of your suggestions. We now conditionally resolve A or AAAA records on CNAMEs based on the question type and we prevent blocky from starting if there is more than 1 mapping for a domain and any of them are a CNAME.

BenMcH:
Sorry, I should have run lint locally before pushing. _Now_ it should pass :crossed_fingers: I verified both tests and lint

resolver/custom_dns_resolver.go
```diff
@@ -120,26 +163,108 @@ func (r *CustomDNSResolver) processRequest(request *model.Request) *model.Respon
...
+
+	if question.Qtype == dns.TypeAAAA {
+		// Resolve ipv6 target recursively
+		targetResp, err := r.processRequest(ctx, aaaaRequest, cnames)
+		if err != nil {
+			return nil, err
+		}
+
+		result = append(result, targetResp.Res.Answer...)
+	}
```
ThinkChaos, line 263: I think we should do like the following to avoid duplication and be ready for other qtypes we add:
```suggestion
	cnames := resolvedCnames
	cnames = append(cnames, targetWithoutDot)

	// Resolve target recursively
	targetReq := newRequestWithClientID(targetWithoutDot, dns.Type(question.Qtype), clientIP, clientID)

	targetResp, err := r.processRequest(ctx, targetReq, cnames)
	if err != nil {
		return nil, err
	}

	result = append(result, targetResp.Res.Answer...)
```

resolver/custom_dns_resolver.go
```diff
@@ -120,26 +163,108 @@ func (r *CustomDNSResolver) processRequest(request *model.Request) *model.Respon
...
+	resolvedCnames []string,
+	question dns.Question,
+) (result []dns.RR, err error) {
+	cname := new(dns.CNAME)
+	ttl := r.cfg.CustomTTL.SecondsU32()
+	cname.Hdr = dns.RR_Header{Class: dns.ClassINET, Ttl: ttl, Rrtype: dns.TypeCNAME, Name: question.Name}
+	cname.Target = dns.Fqdn(targetCname.Target)
+	result = append(result, cname)
+
+	targetWithoutDot := strings.TrimSuffix(targetCname.Target, ".")
```
ThinkChaos, line 231: Can you add an early return here for the case of `question.Qtype == dns.TypeCNAME`?

resolver/custom_dns_resolver.go
```diff
@@ -14,33 +16,48 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+type createAnswerFunc func(question dns.Question, ip net.IP, ttl uint32) (dns.RR, error)
+
 // CustomDNSResolver resolves passed domain name to ip address defined in domain-IP map
 type CustomDNSResolver struct {
 	configurable[*config.CustomDNS]
+	createAnswerFromQuestion createAnswerFunc
```
ThinkChaos, line 24: Very minor but can you move this with the other data fields below, as opposed to the embedded types here? And do the same in `NewCustomDNSResolver`?

ThinkChaos (commented):
Ok last round for me I think :)

ThinkChaos:
IDK why we need to re-approve the CI on every push, I don't remember it being like that before. Maybe because it's your first MR to this repo?

BenMcH:
> IDK why we need to re-approve the CI on every push, I don't remember it being like that before. Maybe because it's your first MR to this repo?

That would make sense. Some github actions are things that you wouldn't want to run unless the code has been reviewed/trusted unless the user is a contributor/trusted. I've been making do with the makefile targets locally though :smile:

kwitsch:
> > IDK why we need to re-approve the CI on every push, I don't remember it being like that before. Maybe because it's your first MR to this repo?
> 
> That would make sense. Some github actions are things that you wouldn't want to run unless the code has been reviewed/trusted unless the user is a contributor/trusted. I've been making do with the makefile targets locally though 😄

As far as I observed the behavior:  
Until your first MR is merged the action runs need reaproval on any change.  
If the first MR is through you are somwhat trusted according to GitHub. 😉

0xERR0R:
> 
> As far as I observed the behavior: Until your first MR is merged the action runs need reaproval on any change. If the first MR is through you are somwhat trusted according to GitHub. 😉

First time contributors need always approval (this is some kind of protection from GitHub to avoid execution of project workflows on each PR) 🤷

BenMcH:
I think that the e2e test failure may have been a false positive. The errors reported were for an http port timeout and an error communicating with mariadb. I ran e2e tests on my local machine this morning and all tests passed.

ThinkChaos (approved):

hozza:
Could someone help with #1644? - I can't get simple CNAME mapping working under "customDNS":

```yaml
mappings:
    test.internal: CNAME(host.internal)
    host.internal: 192.168.1.100
```

or even with:

```yaml
mappings:
    test.internal: "CNAME(host.internal)"
    host.internal: 192.168.1.100
```

The error is `Error: unable to load configuration file '/app/config.yml': failed to unmarshal config from /app/config.yml: wrong file structure: invalid custom DNS entry 'CNAME(host.internal)': invalid IP address 'CNAME(host.internal)'`

It seems this code was merged and should be in the latest docker container (?), but I don't know what's wrong. 

I've looked through [the code here](https://github.com/0xERR0R/blocky/pull/1352/files) but can't work out what's wrong. There's also an old discussion that "unresolved" here at #412. The somewhat annoying stale bot closed down @asianpianoman issue at #1644.

hozza:
Maybe this was not merged or deployed properly as it seems like "configuration.md" on the [live site here](https://0xerr0r.github.io/blocky/latest/configuration/#custom-dns) does not match entirely with the docs submitted in this PR.

i.e. the live/latest docs mention you can use CNAME in the customDNS config instead of an IP, but it does not show how as specified on this PR's file configuration.md line 287.

hozza:
It seems I've been on a wild Goose Chase... The feature was removed, and the docs were not updated.

#1973 and file https://github.com/0xERR0R/blocky/pull/1973/files removed the doc parts referencing CNAME's only a few days ago and not published live yet. 

However, I can't find the commit/reasoning why this was removed or see it in the change log though. Can someone point me to the right commit/PR just for closure I've spent awhile on this one lol.

PR #1353, opened 2024-01-26
fix: set capability for privileged ports

buroa:
This is required for running the container in a non-priviledged way without specifically adding `NET_BIND_SERVICE`.

See here as well: https://github.com/coredns/coredns/blob/master/Dockerfile#L15

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,9 @@ RUN --mount=type=bind,target=. \
   --mount=type=cache,target=/go/pkg \
   make build
 
+# set capability for privileged ports
+RUN setcap cap_net_bind_service=+ep /bin/blocky
+
 # ----------- stage: final
 FROM scratch
 


buroa:
Hey @ThinkChaos @0xERR0R @kwitsch Easy one for you :)

kwitsch:
Not necessary since it is already done through the [Makefile](https://github.com/0xERR0R/blocky/blob/main/Makefile#L54)

buroa:
@kwitsch I just realized that, but it is definitely not working correctly. I have to add this to get it to run: https://github.com/buroa/k8s-gitops/blob/master/kubernetes/apps/networking/blocky/app/helmrelease.yaml#L70

kwitsch:
Strange, I don't have any problems using docker in swarm mode with port 53.

buroa:
> Strange, I don't have any problems using docker in swarm mode with port 53.

It will work fine, but once you start restricting the app's capabilities and running without root, that's where this issue becomes present. You shouldn't need to add `NET_BIND_SERVICE` when you drop everything.

kwitsch:
Just now looked at the [logs](https://github.com/0xERR0R/blocky/actions/runs/7433875791/job/20227417637#step:12:1877) and it should be added. I don't get why you need to specifically do it in a separate run to work. o.O

Since the application inside the container doesn't run as root it has to be set somehow to work even in my setup. 🤔

buroa:
> Just now looked at the [logs](https://github.com/0xERR0R/blocky/actions/runs/7433875791/job/20227417637#step:12:1877) and it should be added. I don't get why you need to specifically do it in a separate run to work. o.O
> 
> Since the application inside the container doesn't run as root it has to be set somehow to work even in my setup. 🤔

So I just removed the flag `BIN_AUTOCAB`, and it actually binded to the port just fine in my k8s setup. How bizarre.

kwitsch:
I am very sure that it is needed if the port inside the container is in the restricted range.
It isn't needed if you run the cap add outside the make.

But I'm unfortunately not sure how containers behave in a kubernetes environment since I only use docker.

buroa:
Can you try running `ghcr.io/buroa/blocky:develop` which has the `setcap` commented out and see if it works in your setup?

kwitsch:
I'll try it in a few days when I'm back at my test setup

PR #1356, opened 2024-01-28
refactor(log): store log in context so it's automatically propagated

### LINKED ISSUE #1307: Improve per request log metadata
It would be nice to have something like a request ID being logged with each log that relates to a specific request to make tracing a request's path easier, and possible to discern logs from different requests.

Should pretty straightforward after logs automatically pickup the request's log via a context (see #1264).

Also in the server we should log how we received the request (TCP/UDP/..., the port, etc.).

ThinkChaos:
As I had mentioned in #1264, this stores the logger in the `Context` we pass around during resolving. This helps better propagate log fields.  
The most user visible impact is `prefix` now contains more resolvers in the chain.  
There's also a new `req_id` field that's a unique ID per request which will making following logs during debug easier (related to #1307).

I also changed `util.LogOnError` to get  the log from a context, and removed `model.Request.Log`.


server/server.go
```diff
@@ -562,62 +561,102 @@ func extractClientIDFromHost(hostName string) string {
...
 
 func newRequest(
+	ctx context.Context,
 	clientIP net.IP, clientID string,
 	protocol model.RequestProtocol, request *dns.Msg,
-) *model.Request {
-	return &model.Request{
+) (context.Context, *model.Request) {
+	ctx, logger := log.CtxWithFields(ctx, logrus.Fields{
+		"req_id":    uuid.New().String(),
```
ThinkChaos, line 569: @0xERR0R in #1307 you mentioned not using UUIDs, do you have something else in mind?  
I just went with that since it was already in `go.mod`, and unique across restarts, but changing should be trivial.

[xid](https://pkg.go.dev/github.com/rs/xid) might be a good option, and its docs has a small comparison that links alternatives.

kwitsch (commented):
Just one question out of curiosity

resolver/ecs_resolver.go
```diff
@@ -49,23 +50,26 @@ func NewECSResolver(cfg config.ECS) ChainedResolver {
 // and sets the client IP from the EDNS0 option to the request if this option is enabled
 func (r *ECSResolver) Resolve(ctx context.Context, request *model.Request) (*model.Response, error) {
 	if r.cfg.IsEnabled() {
+		ctx, logger := r.log(ctx)
+		_ = ctx
```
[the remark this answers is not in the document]
ThinkChaos: Yep exactly. It complains it's unused otherwise, but I think it's better to keep it so if in the future we reference `ctx` we'll get the right one. Otherwise we'd need to notice that in `_, logger := r.log(ctx)` `_` is a context and update that code, which seems unlikely.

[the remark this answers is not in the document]
kwitsch: Agreed, I was just curious if my guess was right since I couldn't test it right now. 😅

kwitsch (approved):
👍

resolver/ecs_resolver.go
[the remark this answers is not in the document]
ThinkChaos: Thanks for taking a look! :purple_heart:

server/server.go
kwitsch: Looked into some UUID alternatives.  

I currently would prefer mid over most of em even if I never worked with it. 😅 
Reducing the string length by a lot sounds tempting.

resolver/caching_resolver.go
```diff
@@ -249,16 +250,16 @@ func isResponseCacheable(msg *dns.Msg) bool {
 	return !msg.Truncated && !msg.CheckingDisabled
 }
 
-func (r *CachingResolver) putInCache(cacheKey string, response *model.Response, ttl time.Duration,
-	publish bool,
+func (r *CachingResolver) putInCache(
+	ctx context.Context, cacheKey string, response *model.Response, ttl time.Duration, publish bool,
```
BenMcH, line 254: This isn't super relevant to the functionality of the pull request, but is there a preference within the project to keep these parameters on the same line vs listing them one per line when the line length becomes too long to pass the linter checks?

BenMcH (approved):
This looks like a positive change! :tada: Just had one (unrelated) question

resolver/caching_resolver.go
ThinkChaos: I think we've got a bit of all versions :smile:  
Personally I don't really mind the format, I think the order and grouping relating things matters more.

server/server.go
ThinkChaos: @0xERR0R any opinion on the ID format?

0xERR0R: We can use message id from dns request (dns.Id) itself, this is not to long. But UUID is also ok (I have only concerns about ID length, this could make log files difficult to read by human).

ThinkChaos: DNS message ID is only unique per client so I think it's best to avoid it here.  
I'll just keep UUID as we already have it as a dependency and if it's annoying we can replace it easily.

PR #1360, opened 2024-01-31
feat: Support defining records by dns zone format

### LINKED ISSUE #1355: Support DNS Zone File Syntax
It was pointed out in #1352 that we could change DNS mappings from the custom YAML formatting to rely on [DNS Zone File](https://en.wikipedia.org/wiki/Zone_file) syntax instead.

Adapted from @ThinkChaos's comment:
This could be implemented using [dns.ZoneParser](https://pkg.go.dev/github.com/miekg/dns#ZoneParser) which supports the standard zonefile format. This would also have the benefit of allowing to split the mapping into another file via `$INCLUDE`.  

Is this something that the project wants to move towards? Initial concerns would be the potential for new users who rely on private DNS entries not necessarily being familiar with the syntax. On the other hand, this is a standard syntax that can be linked to rather than maintaining custom syntax and documentation within this project.

BenMcH:
Closes #1355 

As I got to looking at the issue above, it became obvious to me that support for DNS zone files only requires a very small change to my previous PR. 

Outstanding questions for others:

* Do we remove the `CNAME()` method under `mappings`? This hasn't been published as far as I can tell, so we have the option to remove it before it hits the mainline if we don't like/want that syntax moving forward for non-A records.
* What is needed in terms of testing or documentation surrounding the `$INCLUDE` directive?
* Should we document the `$GENERATE` directive and expect others to use it?
* Currently, TTLs from the zone file are unused, as we always use the TTL from the customDNS.customTTL field. Is this expected behavior? Should we respect TTLs set by the zone file as it's advanced configuration?


kwitsch:
My thoughts on the open questions:

> * Do we remove the `CNAME()` method under `mappings`? This hasn't been published as far as I can tell, so we have the option to remove it before it hits the mainline if we don't like/want that syntax moving forward for non-A records.

I opt for removal since it's not released yet.

> * What is needed in terms of testing or documentation surrounding the `$INCLUDE` directive?

There is a struct in helpertest to generate temporary folders and subsequently string based files for testing. If you look for the usages it should be obvious how to use it in tests. 😉

> * Should we document the `$GENERATE` directive and expect others to use it?

It should be enough to list what is currently not/supported and a link to an proper zonefile documentation (documenting it ourselves is a bit out of scope). 🤔

> * Currently, TTLs from the zone file are unused, as we always use the TTL from the customDNS.customTTL field. Is this expected behavior? Should we respect TTLs set by the zone file as it's advanced configuration? 

customDNS.customTTL should only apply to the old format (documentation should mention it) in my opinion.

docs/configuration.md
```diff
@@ -264,6 +264,7 @@ domain must be separated by a comma.
 | customTTL           | duration (no unit is minutes)               | no        | 1h            |
 | rewrite             | string: string (domain: domain)             | no        |               |
 | mapping             | string: string (hostname: address or CNAME) | no        |               |
+| zoneFileMapping     | multiline string containing a DNS Zone File | no        |               |
```
kwitsch, line 267: Since it supports already more than the simple mapping and is no file by itself I would opt for ` zone`  as config key. 🤔

BenMcH:
Perfect! Thank you for your input @kwitsch. I'll implement these changes as I have some free time in the next day or 2

config/custom_dns.go
```diff
@@ -14,14 +14,56 @@ type CustomDNS struct {
...
+	ZoneFileDNS CustomDNSMapping
 )
 
+func (z *ZoneFileDNS) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var input string
+	if err := unmarshal(&input); err != nil {
+		return err
+	}
+
+	result := make(ZoneFileDNS, len(input))
```
ThinkChaos, line 34: Minor copy-paste typo (`1` seems sensible since you're probably at least going to have a domain if you specify the option):
```suggestion
	result := make(ZoneFileDNS, 1)
```

config/custom_dns.go
```diff
@@ -14,14 +14,56 @@ type CustomDNS struct {
...
+				}
+
+				// Done
+				break
+			}
+
+			domain := zoneRR.Header().Name
+
+			if _, ok := result[domain]; !ok {
+				result[domain] = make(CustomDNSEntries, 0)
```
ThinkChaos, line 55: We know there's at least one entry:
```suggestion
				result[domain] = make(CustomDNSEntries, 1)
```

config/custom_dns.go
```diff
@@ -14,14 +14,56 @@ type CustomDNS struct {
...
+func (z *ZoneFileDNS) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var input string
+	if err := unmarshal(&input); err != nil {
+		return err
+	}
+
+	result := make(ZoneFileDNS, len(input))
+
+	if input != "" {
+		zoneParser := dns.NewZoneParser(strings.NewReader(input), "", "")
```
ThinkChaos, line 37: Let's put an arbitrary file name so errors make sense:
```suggestion
		zoneParser := dns.NewZoneParser(strings.NewReader(input), "", "<YAML config>")
```

Also could you try what setting the origin arg to an empty string does to a relative name?  
Basically does`www 3600 A 1.2.3.4` end up being `www.`, does it error, or something else?  
I'm wondering if it's something we should document or error on our side if no `$ORIGIN` was specified by the user and wet get a relative name.

config/custom_dns_test.go
```diff
@@ -67,7 +69,7 @@ var _ = Describe("CustomDNSConfig", func() {
 		})
 	})
 
-	Describe("UnmarshalYAML", func() {
+	Describe("#CustomDNSEntries UnmarshalYAML", func() {
```
ThinkChaos, line 72: Does `#` do anything here?

ThinkChaos (commented):
Thanks for tackling this!

I agree with everything @kwitsch said above :)
Just have very minor comments.

config/custom_dns_test.go
BenMcH: Nope! I will remove it when I work on this next

config/custom_dns.go
BenMcH: This returns an error. I have written a test to confirm that this remains the case, but I wouldn't think that this is something that the project would document.

BenMcH:
This should be ready for another round of reviews 😄

kwitsch:
Gonna look into it tomorrow. 🙂👍

config/custom_dns.go
ThinkChaos: I agree. Good idea for the test!

config/custom_dns.go
```diff
@@ -14,14 +14,55 @@ type CustomDNS struct {
...
+		}
+
+		domain := zoneRR.Header().Name
+
+		if _, ok := result[domain]; !ok {
+			result[domain] = make(CustomDNSEntries, 1)
+			result[domain][0] = zoneRR
+		} else {
+			result[domain] = append(result[domain], zoneRR)
+		}
```
ThinkChaos, line 58: My bad for the first suggestion, I often make the mistake of creating an array with elements instead of capacity since it's different for arrays and maps. Capacity will let us just use append:
```suggestion
			result[domain] = make(CustomDNSEntries, 0, 1)
		}

		result[domain] = append(result[domain], zoneRR)
```

docs/configuration.md
```diff
@@ -278,13 +279,22 @@ domain must be separated by a comma.
...
 This configuration will also resolve any subdomain of the defined domain, recursively. For example querying any of
 `printer.lan`, `my.printer.lan` or `i.love.my.printer.lan` will return 192.168.178.3.
 
-CNAME records are supported by setting the value of the mapping to `CNAME(target)`. Note that the target will be recursively resolved and will return an error if a loop is detected.
+CNAME records are supported by utilizing the `zone` parameter. The zone file is a multiline string containing a [DNS Zone File](https://en.wikipedia.org/wiki/Zone_file#Example_file).
+For records defined using the `zone` parameter, the `customTTL` parameter is unused. Instead, the TTL is defined in the zone directly.
+The following directives are supported in the zone file:
+* `$ORIGIN` - sets the origin for relative domain names
+* `$TTL` - sets the default TTL for records in the zone
+* `$INCLUDE` - includes another zone file relative to the blocky executable
```
ThinkChaos, line 296: I think it would be better UX to make it relative to the config path, my bad for not realizing this before.

It requires a bit of a hack because the Go YAML unmarshaling API doesn't let you easily pass in additional data (also applies to pretty much any other Go unmarshaling API).  

The idea is to make `ZoneFileDNS` a struct with two fields:
```go
type ZoneFileDNS struct{
	RRs []dns.RR

	configPath string // only used during unmarshaling
}
```

And then in `loadConfig`, we can set `configPath` before calling `unmarshalConfig`, so that `ZoneFileDNS.UnmarshalYAML` can read `z.configPath`.

Here's what that code should look like (additions highlighted):
```go
var (
	data       []byte
	prettyPath = path // addition
)

if fs.IsDir() {
	prettyPath = filepath.Join(path, "*") // addition

	data, err = readFromDir(path, data)
	if err != nil {
		return nil, fmt.Errorf("can't read config files: %w", err)
	}
} else {
	data, err = os.ReadFile(path)
	if err != nil {
		return nil, fmt.Errorf("can't read config file: %w", err)
	}
}

cfg.CustomDNS.Zone.configPath = prettyPath // addition

err = unmarshalConfig(logger, data, &cfg)
```

And finally in `UnmarshalYAML` we can do `dns.NewZoneParser(strings.NewReader(input), "", z.configPath)`.

The downside of this technique is you need a struct with a field that's only used during unmarshaling, but I think it's worth it in this case.

Let me know if I can clarify anything because it's a bit of a long winded comment.

docs/configuration.md
```diff
@@ -259,12 +259,13 @@ You can define your own domain name to IP mappings. For example, you can use a u
...
-| customTTL           | duration (no unit is minutes)               | no        | 1h            |
-| rewrite             | string: string (domain: domain)             | no        |               |
-| mapping             | string: string (hostname: address or CNAME) | no        |               |
-| filterUnmappedTypes | boolean                                     | no        | true          |
+| Parameter           | Type                                                   | Mandatory | Default value |
+| ------------------- | ------------------------------------------------------ | --------- | ------------- |
+| customTTL           | duration used for simple mappings (no unit is minutes) | no        | 1h            |
+| rewrite             | string: string (domain: domain)                        | no        |               |
+| mapping             | string: string (hostname: address or CNAME)            | no        |               |
+| zone                | multiline string containing a DNS Zone File            | no        |               |
```
ThinkChaos, line 267: I think we have similar phrases elsewhere but nothing forces it to be multiline:
```suggestion
| zone                | string containing a DNS Zone                           | no        |               |
```

config/custom_dns_test.go
```diff
@@ -123,12 +112,18 @@ cname 3600 CNAME www
...
+					for _, record := range records {
+						switch r := record.(type) {
+						case *dns.A:
+							Expect(r.A).Should(Equal(net.ParseIP("1.2.3.4")))
+						case *dns.AAAA:
+							Expect(r.AAAA).Should(Equal(net.ParseIP("2001:db8:85a3::8a2e:370:7334")))
+						default:
+							Fail("unexpected record")
+						}
+					}
```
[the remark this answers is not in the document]
BenMcH: Oh I like this a lot! I ended up replacing these tests and the one below using these matchers. I am only checking BeDNSRecord and HaveTTL, however. The reason is that we are only storing dns RRs from the zone file. I don't think we have the rest of the properties, however I could be wrong. 

Going to take a look at the `$INCLUDE` comment now and will hopefully have some code to push yet tonight :crossed_fingers:

docs/configuration.md
BenMcH: This was super helpful. Thank you! I ended up writing 2 additional tests to cover bare filenames (Files in the same dir), relative filenames, `./other.zone`, and kept the original absolute path test so that we can ensure however a user defines their zone file, we _should_ have reasonable assurances that it will correctly resolve.

config/custom_dns_test.go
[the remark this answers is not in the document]
ThinkChaos: Yeah this was more to show what's possible, what you did looks good to me :)

config/custom_dns.go
```diff
@@ -14,14 +14,56 @@ type CustomDNS struct {
...
 )
 
+func (z *ZoneFileDNS) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var input string
+	if err := unmarshal(&input); err != nil {
+		return err
+	}
+
+	z.RRs = make(CustomDNSMapping)
+	result := z.RRs
```
ThinkChaos, line 38: This doesn't matter in our code ATM, but can you change this so we do `result := make(CustomDNSMapping)` here, and only assign `z.RRs` at the very end?  
That way the method is a no-op in case of error, which can be useful if you're unmarshaling into an instance with existing data.

config/config.go
```diff
@@ -470,6 +475,8 @@ func loadConfig(logger *logrus.Entry, path string, mandatory bool) (rCfg *Config
 		}
 	}
 
+	cfg.CustomDNS.Zone.configPath = prettyPath
```
ThinkChaos, line 478: Could you move the two non-absolute `$INCLUDE` tests to `config_test.go` and change them to parse the config from YAML instead of creating a `ZoneFileDNS` manually in Go?  
That way we're testing the whole flow of the feature.

There's already a bunch of tests in there that create files we run through the whole config parsing pipeline.  
One example:
```go
It("should ignore non YAML files", func() {
	tmpDir.CreateStringFile("ignore-me.txt", "THIS SHOULD BE IGNORED!")

	_, err := LoadConfig(tmpDir.Path, true)
	Expect(err).Should(Succeed())
})
```

ThinkChaos (commented):
Two details, besides that everything looks good to me!

config/config.go
BenMcH: This was a good move. It showed that we hadn't accounted for cases where blocky is launched with a directory as the config parameter. `blocky -c ./config/` would have previously failed :sweat_smile:

ThinkChaos: Nice, quick payoff!

ThinkChaos (approved):
Thanks again, this is great!

kwitsch:
> @kwitsch I didn't merge in case you had somethign to add. Anything left? If not can you press the green button? :)

✅

PR #1129, opened 2023-09-07
feat: API-first approach for REST interface

0xERR0R:
Motivation and detailed description: see #808 

short summary:
- API specification with OpenAPI standard (yaml file under docs/api)
- We generate server stub and REST client from this OpenAPI file
- Implementation of server stub
- CLI commands use generated REST client
- OpenAPI and RapiDoc UI are embedded in final binary and are accessible from index.html
- Documentation (GitHub pages) provides OpenAPI and RapiDoc

closes #808

PR #1373, opened 2024-02-10
fix: Corrects ability to disable all blocking from the CLI by not providing any groups

### LINKED ISSUE #1336: v0.23 broke `blocky blocking disable`
when I ran `blocky blocking disable`

blocky 0.22 responded:
> [2024-02-10 02:55:02]  INFO OK

blocky 0.23 responds:
> Error: response NOK, 400 Bad Request group '' is unknown

works totally fine in 0.23 when I run `curl localhost:4000/api/blocking/disable`

CC: @ThinkChaos  @0xERR0R @BenMcH

BenMcH:
This PR fixes #1336 and restores the v0.22 behavior of calling `blocky blocking disable` without any additional arguments. It appears that in #1129, we lost the following check from api_endpoints.go, lines 123-126:

```go
	groupsParam := req.URL.Query().Get("groups")
	if len(groupsParam) > 0 {
		groups = strings.Split(groupsParam, ",")
	}
```

This parsing, being done by hand, allowed us to receive a query string, `groups=` and implicitly drop this variable. This PR adds this check back in

### DIFF
--- a/api/api_interface_impl.go
+++ b/api/api_interface_impl.go
@@ -103,7 +103,7 @@ func (i *OpenAPIInterfaceImpl) DisableBlocking(ctx context.Context,
 		}
 	}
 
-	if len(*request.Params.Groups) > 0 {
+	if request.Params.Groups != nil && len(*request.Params.Groups) > 0 {
 		groups = strings.Split(*request.Params.Groups, ",")
 	}
 
--- a/api/api_interface_impl.go
+++ b/api/api_interface_impl.go
@@ -105,8 +105,6 @@ func (i *OpenAPIInterfaceImpl) DisableBlocking(ctx context.Context,
 
 	if request.Params.Groups != nil && len(*request.Params.Groups) > 0 {
 		groups = strings.Split(*request.Params.Groups, ",")
-	} else {
-		groups = []string{}
 	}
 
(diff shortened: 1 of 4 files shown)


api/api_interface_impl.go
```diff
@@ -103,8 +103,10 @@ func (i *OpenAPIInterfaceImpl) DisableBlocking(ctx context.Context,
 		}
 	}
 
-	if request.Params.Groups != nil {
+	if request.Params.Groups != nil && len(*request.Params.Groups) > 0 {
 		groups = strings.Split(*request.Params.Groups, ",")
+	} else {
+		groups = []string{}
```
ThinkChaos, line 109: I think we can skip the `else` since usually a `nil` and an empty slice behave the same (`len(x) == 0` for both).

ThinkChaos (commented):
Thanks for looking into it and finding the root cause!

api/api_interface_impl.go
BenMcH: Oh I just noticed I didn't get rid of the nil check. I'll do that shortly

ThinkChaos: I think we do need that `nil` check cause it's for the pointer to the `Groups` string. It's likely `nil` when the query arg is absent.

ThinkChaos (approved):

PR #1376, opened 2024-02-13
chore(build): update testcontainers dependency

0xERR0R:
This PR updates all testcontainers dependencies to v0.27.0. This should close 4 dependabot PRs

In the version v0.27.0 the network creation was changed:
- `testcontainers.GenericNetwork` and `testcontainers.NetworkRequest` used in helper.go are deprecated, I added linter exclusion here
- I have tried to rewrite the code and use network.New() as suggested by testcontainers team, but this breaks current logic with testNetwork in helper.go: It is not possible to pass the network name, new API always generates new unique name, see https://github.com/testcontainers/testcontainers-go/pull/1993
- I think, we can leave the deprecated code for now, but we need to rework the network creation in e2e (maybe to create one network on test start (before suite) and use it for all tests?)


kwitsch:
That looks promising. 
Most of our current logic regarding the e2e-networks exists to combat problems with the old implementation. 
I skimmed over the merge and it looks like we could discard much of it. 👍
I'll try changing the tests to the new methods. 😊

kwitsch (approved):
I'll refactor the e2e-networks in a follow up PR.

mdelapenya:
Hey team, please let us know if we can be of any help, but it feels you are already doing a great job keeping up with the changes 🙇

PR #1402, opened 2024-03-14
Build Cache Optimization

kwitsch:
Changes:
- use bind mounts instead of COPY in Dockerfile
- layer cache optimization during Docker build
- preload go modules in makefile workflow for better cache usage in matrix steps


ThinkChaos (approved):
Nice!

PR #1424, opened 2024-04-02
fix(log): don't print querylog target password when using a database

### LINKED ISSUE #1421: Blocky leaks credentials while logging config at stratup
Blocky prints out the config that it is using at startup. 

It seems that the raw config is printed out, including credentials to connect to a database for query logging.

```
│ [2024-03-31 13:17:02]  INFO server: -> query_logging:                                                                                                                                                                                                                                                                                            │
│ [2024-03-31 13:17:02]  INFO server:      type: postgresql                                                                                                                                                                                                                                                                                        │
│ [2024-03-31 13:17:02]  INFO server:      target: postgresql://blocky_app_user:wG7SNYaf9tGu01w-GYgK@main.psql.svc.cluster.local:5432/blocky                                                                                                                                                                                                       │
│ [2024-03-31 13:17:02]  INFO server:      logRetentionDays: 90                                                                                                                                                                                                                                                                                    │
│ [2024-03-31 13:17:02]  INFO server:      flushInterval: 30 seconds                                                                                                                                                                                                                                                                               │
│ [2024-03-31 13:17:02]  INFO server:      fields: [clientIP clientName responseReason responseAnswer question duration]                                                                                                                                                                                                                           │
│ [2024-03-31 13:17:02]  INFO server: -> rewrite:                                                                         
```

I've already rotated this password, so no need to worry about posting it.

![image](https://github.com/0xERR0R/blocky/assets/17928966/eba713cb-ddca-4dc9-baa9-138cc8033ecf)

ThinkChaos:
Ideally we could add a e2e test that configures trace level logging and all config options that have a secret, does a query, and then checks the container logs don't contain any secret (making all secrets contain a unique token would make that easy).

Fixes #1421

### DIFF
--- a/config/query_log.go
+++ b/config/query_log.go
@@ -57,14 +57,10 @@ func (c *QueryLog) censoredTarget() string {
 		return c.Target
 	}
 
-	if target.User == nil {
+	pass, ok := target.User.Password()
+	if !ok {
 		return c.Target
 	}
 
-	// Drop the password since special chars like * get URL escaped
-	if pass, hasPass :=target.User.Password(); hasPass {
-		return strings.Replace(target.String(), pass, strings.Repeat("*", len(pass)), 1)
-	}
-
-	return target.String()
+	return strings.ReplaceAll(c.Target, pass, secretObfuscator)
 }
--- a/config/query_log.go
+++ b/config/query_log.go
@@ -1,6 +1,9 @@
 package config
 
 import (
+	"net/url"
+	"strings"
+
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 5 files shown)


config/query_log.go
```diff
@@ -41,3 +44,25 @@ func (c *QueryLog) LogConfig(logger *logrus.Entry) {
...
+	if err != nil {
+		return c.Target
+	}
+
+	if target.User == nil {
+		return c.Target
+	}
+
+	// Drop the password since special chars like * get URL escaped
+	target.User = url.User(target.User.Username())
```
kwitsch, line 65: If the password is completly droped it might seem like it isen't set at all which could indicate a configuration error.

It might be better to just replace the password at string level
```suggestion
	if pass, hasPass :=target.User.Password(); hasPass {
		return strings.Replace(target.String(), pass, strings.Repeat("*", len(pass)), 1)
	}
```

ThinkChaos: Good point!

ThinkChaos:
Added another commit to use constant a length obfuscation string (including for redis).

kwitsch (approved):

PR #1426, opened 2024-04-02
ci: minor changes

ThinkChaos:
- Add `on: push: paths:` to limit when some workflows are run
- Add a hint that the fork sync workflow can be disabled without modifying the file
- Remove the concurrency limit for the Makefile workflow:

   I like to push my commits one by one so they all get the full CI validation.
   This limits forces me to push a commit, and wait for the result before pushing another.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,10 +2,12 @@ name: docs
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


kwitsch (approved):
👍

Issue #1206, opened 2023-10-18
Enabling/Disabling of services/ports


0xERR0R:
Configuration of ports should be changed to:

```yaml
services:
  dns:
    ports:
       plain: 53
       tls: 853
       https: 443
    rate-limit: 60 # just example for further config
  metrics:
    ports:
       http: 4001
       https: 443 
    path: /mymetrics # moved from prometheus.path
  control-api:
    ports:
       http: 3000
    api-key: ASBC123 # just example for further config
  pprof:
    ports:
       http: 12345
```

prometheus configuration should be removed (migrated to services.metrics)

kwitsch:
Just to verify: most settings of the basic configuration would also be moved(certFile, key file,minTlsServeVersion) since they are currently connected to ports, correct?

ThinkChaos:
Sounds like that's what ` # just example for further config` shows, so I'd say yeah.  
Could be done in two steps if that's easier.

ThinkChaos:
Started working on this though I'm doing a preliminary step without all the config changes, and just focusing on getting DoH separate from other HTTP services ATM, but doing it in a way that'll be re-usable for further splitting.  
I think I can get a PR up this week. I assigned myself here just until then, but might unassign after that depending on motivation for refactoring the rest.

ThinkChaos:
I like the original proposal, but ran into some ambiguity: if we add `dns: ports: unix:`, it's not clear what protocol should be sent over the Unix socket. For instance it could be DoH, or plain DNS.  

I think we should add an optional transport layer to the config to solve that:
```yaml
services:
  dns:
    expose: # instead of ports since you can specify an IP/hostname
      # Short version uses the protocol's default transport(s)
      plain: 53 # TCP+UDP
      http: localhost:80 # TCP, single value
      https: # TCP, YAML native list to replace splitting on `,`
        - 443
        - localhost:4430

      # Explicit transport(s)
      http:
        unix: /run/blocky/http.sock
      https:
        tcp: 443
        unix: /run/blocky/https.sock

  # Same thing applies to other services
  # Though not all services support all protocols
  metrics:
    listen:
      # `plain:` would be invalid in this context
      http:
        unix: /run/blocky/http.sock
      https: 443

    path: /mymetrics # support for extra config is the same as initial proposal
```

What do you think of this `services` format?

What do you think of the `expose` name?  
[...]

PR #1427, opened 2024-04-03
refactor: add service package and start splitting HTTP handling

ThinkChaos:
Progress for #1206.

No user visible changes in this PR: it's just refactoring and the new config cannot be set from the YAML file.  
Instead the new config structs are filled with the info from `ports:` until we're ready to deprecate `ports` fully.

The first part of the commits refactor how we setup the HTTP listeners.
The second part adds a new `service` package with interfaces used to split up `server.Server`.  
The commits after adding that package refactor the DoH, metrics, and API to be services.

Here's the docs that package as an overview of the design:
> Package service exposes types to abstract services from the networking.
> 
> The idea is that we build a set of services and a set of network
> endpoints (Listener). The services are then assigned to endpoints based
> on the address(es) they were configured for.
> 
> Actual service to endpoint binding is not handled by the abstractions in
> this package as it is protocol specific.
> The general pattern is to make a "server" that wraps a service, and can
> then be started on an endpoint using a `Serve` method,
> similar to `http.Server`.
> 
> To support exposing multiple compatible services on a single endpoint
> (example: DoH + metrics on a single port), services can implement `Merger`.

This doesn't touch DNS listener/server code (besides moving the TLS config to a single spot), but I think we'll be able to re-use the same code to setup DNS listeners, and split the DNS server logic into a bunch of services + one server.

The follow-up work is to continue removing things from `configureHTTPRouter` until it's empty and we can remove the transitional `httpMiscService`.  
At that point we'll be ready to deprecate `ports` and switch to the new `services` config format under discussion in #1206.

### DIFF
--- a/api/api_interface_impl.go
+++ b/api/api_interface_impl.go
@@ -53,7 +53,7 @@ type CacheControl interface {
 	FlushCaches(ctx context.Context)
 }
 
-func RegisterOpenAPIEndpoints(router chi.Router, impl StrictServerInterface) {
+func registerOpenAPIEndpoints(router chi.Router, impl StrictServerInterface) {
 	middleware := []StrictMiddlewareFunc{ctxWithHTTPRequestMiddleware}
 
 	HandlerFromMuxWithBaseURL(NewStrictHandler(impl, middleware), router, "/api")
--- a/api/api_interface_impl_test.go
+++ b/api/api_interface_impl_test.go
@@ -105,7 +105,7 @@ var _ = Describe("API implementation tests", func() {
 	Describe("RegisterOpenAPIEndpoints", func() {
 		It("adds routes", func() {
 			rtr := chi.NewRouter()
-			RegisterOpenAPIEndpoints(rtr, sut)
+			registerOpenAPIEndpoints(rtr, sut)
 
 			Expect(rtr.Routes()).ShouldNot(BeEmpty())
 		})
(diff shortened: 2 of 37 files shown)


ThinkChaos:
I still need to write tests for the new package, and the various new "services", but besides that it's in a reviewable state :)

0xERR0R:
Skimmed over your PR and it looks very good! :+1:

kwitsch (approved):
Read through it and I think you did an amazing job. 👍

This approach looks like a really good foundation for a step wise implementation without end-user obstruction.

ThinkChaos:
I'm satisfied with where tests are at now. Codecov doesn't seem to understand the PR being based on a branch that's not targeting main (or maybe it's because I changed it), so there's no explicit percentage diff, but most new code is covered.  
The total coverage dropped 2% because I added a suite to the metrics package, but that's just more realistic :smile:

Also the e2e tests have been pretty flaky for me today, not sure if it's my changes or not.
Here's an example failure: https://github.com/0xERR0R/blocky/actions/runs/10639898980/job/29498704949

ThinkChaos:
@t-e-s-tweb thanks for the interesting and trying it out, there's no

ThinkChaos:
Then building the branch and making no changes to the config should work.  
The `CopyPortsToServices` function in `config.go` takes care of mapping the existing YAML structure to the new code.

ThinkChaos:
GitHub auto closed this with no option for me to reopen because the target branch was merged into main.
So I made a new PR: #1595.

(I was hoping GitHub had the same logic as GitLab where merging a MR changes the other MRs targeting that branch to now target where the first MR was merged, but no I guess  not)

PR #1428, opened 2024-04-04
Add support for time-base scheduling of blacklists

laghoule:
My use case:

I have created a blacklist URL of YouTube, and I want to create a schedule for when I want it to block access.

Example of config:

```yaml
blocking:
  blackLists:
    youtube:
      - https://raw.githubusercontent.com/gieljnssns/Social-media-Blocklists/master/pihole-youtube.txt
  schedules:
    youtube:
      - days: ["Mon", "Tue", "Wed", "Thu"]
        hoursRanges:
          - "06:45-08:30"
          - "19:00-23:30"
      - days: ["Sun"]
        hoursRanges:
          - "20:00-23:30" 
```

So, I have implemented it, with updated doc, readme & unit-tests. It's working fine on my K8S cluster for 1 month.
Let me know what you think.

Thanks for the project.


kwitsch:
This looks interesting. 👍

A case I just thought about which could be problematic with the current implementation:
- blocklist has schedule 8 to 17 o'clock 
- block TTL is set to 1 hour 
- temporary blocked URL is requested at 16.45
- same URL(not blocked anymore) is requested at 16.15

Same applies the other way around(pre and post 8). 

This is just a theoretical scenario since such huge TTLs aren't beneficial most of the time(I know some users configure such TTLs tho). 
This scenario isn't limited to this change but also applies to disabling blocklists manually. 

Best practice in my opinion would be to clear cache entries covered by newly enabled/disabled blocklists.

laghoule:
@kwitsch Thanks for your feedback, I will look into it.

PR #1442, opened 2024-04-10
tests(resolver): fix data race introduced by querylog ignore

ThinkChaos:
Race only happens in tests: the write removed here can happen at the same time as `writeLog` is reading the struct field since cancelling the context doesn't guarantee immediate shutdown.
We can just use the existing channel so the field becomes read-only, avoiding the race.

Example CI failure:
https://github.com/0xERR0R/blocky/actions/runs/8632315589/job/23662702020

For local repro, adding a 2ms sleep to `writeLog`'s startup was enough for me.


kwitsch (approved):
Nice find 👍

ThinkChaos:
Wouldn't have found it without the CI or Makefile, so thanks to you and @0xERR0R :)

PR #1445, opened 2024-04-12
Feature: binary build workflow

kwitsch:
Added a new workflow for binary artifact creation. 
It runs only on dispatch and creates a zip achive with a blocky binary in it compiled acording to the selected parameters(os, arch & arm version).
The artifact is only stored for 5 days to limit storage polution.

The idea behind it is to simplify binary distribution to users for testing purpose while working on issues.


ThinkChaos (approved):
Nice!

PR #1446, opened 2024-04-12
Prefetch syncs to redis

### LINKED ISSUE #1422: Prefetch updates aren't published to Redis
Prefetch update results aren't published to Redis.

Implementation At caching_resolver: https://github.com/0xERR0R/blocky/blob/main/resolver%2Fcaching_resolver.go#L94

kwitsch:
Changes:
- reloadCacheEntry removes EDNS0 records from response
- reloadCacheEntry publishes its result to Redis

Closes #1422

Related #1420

### DIFF
--- a/resolver/caching_resolver.go
+++ b/resolver/caching_resolver.go
@@ -111,21 +111,33 @@ func (r *CachingResolver) reloadCacheEntry(ctx context.Context, cacheKey string)
 	logger.Debugf("prefetching '%s' (%s)", util.Obfuscate(domainName), qType)
 
 	req := newRequest(dns.Fqdn(domainName), qType)
+
 	response, err := r.next.Resolve(ctx, req)
-
-	if err == nil {
-		if response.Res.Rcode == dns.RcodeSuccess {
-			packed, err := response.Res.Pack()
-			if err != nil {
-				logger.Error("unable to pack response", err)
-
-				return nil, 0
-			}
-
-			return &packed, r.adjustTTLs(response.Res.Answer)
-		}
-	} else {
+	if err != nil {
 		util.LogOnError(ctx, fmt.Sprintf("can't prefetch '%s' ", domainName), err)
+
+		return nil, 0
+	}
+
+	if response.Res.Rcode == dns.RcodeSuccess {
+		respCopy := response.Res.Copy()
+
+		// don't cache any EDNS OPT records
+		util.RemoveEdns0Record(respCopy)
+
+		packed, err := respCopy.Pack()
+		if err != nil {
+			logger.Error("unable to pack response", err)
+
+			return nil, 0
+		}
+
+		if r.redisClient != nil {
+			res := *respCopy
+			r.redisClient.PublishCache(cacheKey, &res)
+		}
+
+		return &packed, r.adjustTTLs(response.Res.Answer)
 	}
 
 	return nil, 0
--- a/resolver/caching_resolver.go
+++ b/resolver/caching_resolver.go
@@ -17,8 +17,6 @@ import (
 
 const (
 	defaultCachingCleanUpInterval = 5 * time.Second
-	// noCacheTTL indicates that a response should not be cached
-	noCacheTTL = uint32(0)
 )
 
 // CachingResolver caches answers from dns queries with their TTL time,
@@ -122,7 +120,7 @@ func (r *CachingResolver) reloadCacheEntry(ctx context.Context, cacheKey string)
 	}
 
 	ttl, res := r.createCacheEntry(logger, response.Res)
-	if ttl == noCacheTTL || len(res) == 0 {
+	if ttl == 0 || len(res) == 0 {
 		return nil, 0
 	}
 
@@ -142,9 +140,7 @@ func (r *CachingResolver) redisSubscriber(ctx context.Context) {
 			if rc != nil {
 				_, domain := util.ExtractCacheKey(rc.Key)
 
-				dlogger := logger.WithField("domain", util.Obfuscate(domain))
-
-				dlogger.Debug("received from redis")
+				logger.WithField("domain", util.Obfuscate(domain)).Debug("received from redis")
 
 				r.resultCache.Put(rc.Key, &rc.Entry, util.ToTTLDuration(rc.TTL))
 			}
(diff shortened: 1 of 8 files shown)


resolver/caching_resolver.go
```diff
@@ -111,21 +111,33 @@ func (r *CachingResolver) reloadCacheEntry(ctx context.Context, cacheKey string)
...
 		}
-	} else {
-		util.LogOnError(ctx, fmt.Sprintf("can't prefetch '%s' ", domainName), err)
+
+		if r.redisClient != nil {
+			res := *respCopy
+			r.redisClient.PublishCache(cacheKey, &res)
+		}
+
+		return &packed, r.adjustTTLs(response.Res.Answer)
```
ThinkChaos, line 140: This is also a slight difference compared to the `putInCache` flow: here Redis sees the unadjusted TTLs whereas when we add values using `putInCache` Redis gets the adjusted TTLs.  
My initial reaction is this is correct, and the other code should change since other blocky instances might have different TTL configuration.

resolver/caching_resolver.go
```diff
@@ -111,21 +111,33 @@ func (r *CachingResolver) reloadCacheEntry(ctx context.Context, cacheKey string)
...
+	}
 
-				return nil, 0
-			}
+	if response.Res.Rcode == dns.RcodeSuccess {
+		respCopy := response.Res.Copy()
 
-			return &packed, r.adjustTTLs(response.Res.Answer)
+		// don't cache any EDNS OPT records
+		util.RemoveEdns0Record(respCopy)
```
ThinkChaos, line 126: Nice catch!

Based on my other comments and this being missing before, I think we should try and factorize the code to have a single func we can call from here and `putInCache`. Maybe something that adds the response to Redis, and returns the local cache TTL + the entry with adjusted TTL.

ThinkChaos (commented):
I have a couple comments but it's mostly stuff I noticed about the existing code when comparing with the new one!

resolver/caching_resolver.go
kwitsch: Sending the unadjusted TTL to Redis isn't a problem since received entries are TTL adjusted upon receiving before they are put in the local cache. 

This way instances with different configs can  share the same sync channel without breaking each other's settings. 

That was the base idea when I implemented it. I get that this is counterintuitive and should be changed.

resolver/caching_resolver.go
```diff
@@ -111,21 +111,33 @@ func (r *CachingResolver) reloadCacheEntry(ctx context.Context, cacheKey string)
...
+		packed, err := respCopy.Pack()
+		if err != nil {
+			logger.Error("unable to pack response", err)
+
+			return nil, 0
 		}
-	} else {
-		util.LogOnError(ctx, fmt.Sprintf("can't prefetch '%s' ", domainName), err)
+
+		if r.redisClient != nil {
```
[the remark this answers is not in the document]
kwitsch: Ur correct sending it when pack fails is an error in putInCache

kwitsch:
@ThinkChaos started refactoring the whole caching_resolver. 
May take a bit more time but I think there's some enhancement potential regarding code duplication and inconsistent behavior. 

Thanks for the input. 👍 

I'll let you know when I'm done.

kwitsch:
@ThinkChaos still needs some tests but may I get your opinion on the current changes? 
I'm a bit unsure if the direction I took was an advancement? 🤔

resolver/caching_resolver.go
```diff
@@ -107,28 +108,36 @@ func configureCaches(ctx context.Context, c *CachingResolver, cfg *config.Cachin
...
+
+		return nil, 0
 	}
 
-	return nil, 0
+	if r.redisClient != nil {
+		r.redisClient.PublishCache(cacheKey, cacheCopy)
+	}
+
+	return &packed, util.ToTTLDuration(ttl)
```
ThinkChaos, line 140: I think the biggest cause of dupe code is that here we need to return the new cache entry,
but in `r.Resolver` we need to call `r.resultCache.Put`.
So we could change the prefetch API to be more like the normal caching flow:
`PrefetchingOptions.ReloadFn func(ctx context.Context, key string) (*T, time.Duration)`
could be something like `OnAboutToExpire func(ctx context.Context, key string)`,
where the callback is expected to call `Put` if they want to update the value.
That way `reloadCacheEntry` could become a tiny wrapper around `r.Resolve` (or a new `resolve` method they both call).

From a quick check, `ReloadFn` is only used by this code (and the cache tests),
and `Put` allows replacing a value, so I think that should be doable!

From there I think most things will fall into place :)

resolver/caching_resolver.go
```diff
@@ -172,63 +185,56 @@ func (r *CachingResolver) Resolve(ctx context.Context, request *model.Request) (
...
 		if err == nil {
-			cacheTTL := r.adjustTTLs(response.Res.Answer)
-			r.putInCache(ctx, cacheKey, response, cacheTTL, true)
+			ttl := r.modifyResponseTTL(response.Res)
+			if ttl > noCacheTTL {
+				cacheCopy := r.putInCache(logger, cacheKey, response)
+				if cacheCopy != nil && r.redisClient != nil {
+					r.redisClient.PublishCache(cacheKey, cacheCopy)
+				}
+			}
```
ThinkChaos, line 212: This still adjusts the TTL twice, but if we can get everything down to a single code path, it means `putInCache` can also publish to Redis, which will make this whole part simpler.

ThinkChaos:
I had to checkout the branch locally to browse with my editor cause it is indeed a lot of functions calling combined in a couple different ways so definitely hard to follow!

kwitsch:
> I had to checkout the branch locally to browse with my editor cause it is indeed a lot of functions calling combined in a couple different ways so definitely hard to follow!

Yeah I know, that's one of the reasons why I was a bit unsure about the direction. 🫣

resolver/caching_resolver.go
kwitsch: Maybe wrap it in another function? 🤔

I tried removing the publish Boolean parameter since it seemed misplaced. 
On the other hand also a method to publish without put is needed (prefetching)... 🥴
I'm a little troubled fitting these 3 requirements in one logical block.

resolver/caching_resolver.go
kwitsch: Thanks for the input I think that was the part which caused me the most headaches. 😅

I also try to refactor the caches to streamline the resolver behavior. 👍

resolver/caching_resolver.go
ThinkChaos: >  On the other hand also a method to publish without put is needed (prefetching)... 🥴

Not sure if you saw my other comment, but I think we can make this go away without too much trouble.

kwitsch: Already commented on it. 👍

Sorry, read and replayed in reverse order. 🫣

kwitsch:
Good thing I looked deeper into the caches since I found a remaining `context.Background()` : https://github.com/0xERR0R/blocky/blob/main/cache%2Fexpirationcache%2Fexpiration_cache.go#L129

PR #1460, opened 2024-04-23
Log the rule which is the cause of blocking

### LINKED ISSUE #1458: Show the rule which is the cause of blocked request
Recently I've added a few converted AGH lists to Blocky (`v0.23`) and `hub.docker.com` started to being blocked.

I have `blocking request BLOCKED CNAME (ads)`, but there is no link to the rule, which is the cause. I had to bisect my lists. It was `dandelion_sprouts_anti_malware_list`. But it doesn't contain `docker`, `dckr`, `elb.us-east-1.amazonaws.com`.

It would be great if the rule is showed in the logs:
```
DEBUG blacklist_resolver: blocking request 'BLOCKED CNAME (ads: *.docler.com)'
INFO queryLog: query resolved ... question_name=hub.docker.com. question_type=A response_code=NXDOMAIN response_reason=BLOCKED CNAME (ads) block_rule=*.docler.com response_type=BLOCKED
```
or even point to the list and rule number:
```
DEBUG blacklist_resolver: blocking request 'BLOCKED CNAME (ads: anti_malware_list:100500)'
INFO queryLog: query resolved ... question_name=hub.docker.com. question_type=A response_code=NXDOMAIN response_reason=BLOCKED CNAME (ads) block_rule=anti_malware_list:100500 response_type=BLOCKED
```

[hub.docker.com-blocky.log](https://github.com/0xERR0R/blocky/files/15064019/hub.docker.com-blocky.log)

Problem are getting worse, if we use a bunch of regex rules, I think.

---

Searched how lists are processed, found #558. So, this is duplicate and probably should be closed.

zc-devs:
Part of #1458

Examples:
```
DEBUG Trie: wildcard block rule 'crashlytics.com' matched with 'firebase-settings.crashlytics.com'
DEBUG Trie: wildcard block rule 'telemetry.gfe.nvidia.com' matched with 'telemetry.gfe.nvidia.com'
DEBUG Trie: wildcard block rule 'mobile.events.data.microsoft.com' matched with 'mobile.events.data.microsoft.com'

DEBUG stringMap: block rule 'appboy.com' matched with 'appboy.com'
DEBUG stringMap: block rule 'app.adjust.com' matched with 'app.adjust.com'
```

### DIFF
--- a/cache/stringcache/string_caches.go
+++ b/cache/stringcache/string_caches.go
@@ -50,7 +50,11 @@ func (cache stringMap) contains(searchString string) bool {
 	})
 
 	if idx < searchBucketLen {
-		return cache[searchLen][idx*searchLen:idx*searchLen+searchLen] == strings.ToLower(normalized)
+		blockRule := cache[searchLen][idx*searchLen : idx*searchLen+searchLen]
+		if blockRule == normalized {
+			log.PrefixedLog("stringMap").Debugf("block rule '%s' matched with '%s'", blockRule, searchString)
+			return true
+		}
 	}
 
 	return false
--- a/cache/stringcache/string_caches.go
+++ b/cache/stringcache/string_caches.go
@@ -52,7 +52,8 @@ func (cache stringMap) contains(searchString string) bool {
 	if idx < searchBucketLen {
 		blockRule := cache[searchLen][idx*searchLen : idx*searchLen+searchLen]
 		if blockRule == normalized {
-			log.PrefixedLog("stringMap").Debugf("block rule '%s' matched with '%s'", blockRule, searchString)
+			log.PrefixedLog("string_map").Debugf("block rule '%s' matched with '%s'", blockRule, searchString)
+
 			return true
 		}
 	}
@@ -136,7 +137,7 @@ func (cache regexCache) elementCount() int {
 func (cache regexCache) contains(searchString string) bool {
 	for _, regex := range cache {
 		if regex.MatchString(searchString) {
-			log.PrefixedLog("regexCache").Debugf("regex '%s' matched with '%s'", regex, searchString)
+			log.PrefixedLog("regex_cache").Debugf("regex '%s' matched with '%s'", regex, searchString)
 
 			return true
 		}
--- a/trie/trie.go
+++ b/trie/trie.go
@@ -1,5 +1,10 @@
 package trie
 
+import (
+	"github.com/0xERR0R/blocky/log"
+	"strings"
+)
+
 // Trie stores a set of strings and can quickly check
 // if it contains an element, or one of its parents.
 //
@@ -108,8 +113,11 @@ func (n *parent) insert(key string, split SplitFunc) {
 }
 
 func (n *parent) hasParentOf(key string, split SplitFunc) bool {
+	searchString := key
+	rule := ""
 	for {
 		label, rest := split(key)
+		rule = strings.Join([]string{label, rule}, ".")
 
 		child, ok := n.children[label]
 		if !ok {
@@ -132,7 +140,13 @@ func (n *parent) hasParentOf(key string, split SplitFunc) bool {
 
 		case terminal:
 			// Continue down the trie
-			return child.hasParentOf(rest, split)
+			matched := child.hasParentOf(rest, split)
+			if matched {
+				rule = strings.Join([]string{child.String(), rule}, ".")
+				rule = strings.Trim(rule, ".")
+				log.PrefixedLog("Trie").Debugf("wildcard block rule '%s' matched with '%s'", rule, searchString)
+			}
+			return matched
 		}
 	}
 }
--- a/trie/trie.go
+++ b/trie/trie.go
@@ -115,6 +115,7 @@ func (n *parent) insert(key string, split SplitFunc) {
 func (n *parent) hasParentOf(key string, split SplitFunc) bool {
 	searchString := key
 	rule := ""
+
 	for {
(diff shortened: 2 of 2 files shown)


cache/stringcache/string_caches.go
```diff
@@ -50,7 +50,11 @@ func (cache stringMap) contains(searchString string) bool {
 	})
 
 	if idx < searchBucketLen {
-		return cache[searchLen][idx*searchLen:idx*searchLen+searchLen] == strings.ToLower(normalized)
+		blockRule := cache[searchLen][idx*searchLen : idx*searchLen+searchLen]
+		if blockRule == normalized {
+			log.PrefixedLog("stringMap").Debugf("block rule '%s' matched with '%s'", blockRule, searchString)
```
ThinkChaos, line 55: We usually have the prefixes/tags be `snake_case`:
```suggestion
			log.PrefixedLog("string_map").Debugf("block rule '%s' matched with '%s'", blockRule, searchString)
```

trie/trie.go
```diff
@@ -132,7 +140,13 @@ func (n *parent) hasParentOf(key string, split SplitFunc) bool {
 
 		case terminal:
 			// Continue down the trie
-			return child.hasParentOf(rest, split)
+			matched := child.hasParentOf(rest, split)
+			if matched {
+				rule = strings.Join([]string{child.String(), rule}, ".")
+				rule = strings.Trim(rule, ".")
+				log.PrefixedLog("Trie").Debugf("wildcard block rule '%s' matched with '%s'", rule, searchString)
```
ThinkChaos, line 147: Same thing about the case convention:
```suggestion
				log.PrefixedLog("trie").Debugf("wildcard block rule '%s' matched with '%s'", rule, searchString)
```

cache/stringcache/string_caches.go
```diff
@@ -50,7 +50,11 @@ func (cache stringMap) contains(searchString string) bool {
...
+			return true
```
ThinkChaos, line 56: For the linter:
```suggestion

			return true
```

trie/trie.go
```diff
@@ -108,8 +113,11 @@ func (n *parent) insert(key string, split SplitFunc) {
 }
 
 func (n *parent) hasParentOf(key string, split SplitFunc) bool {
+	searchString := key
+	rule := ""
```
ThinkChaos, line 117: Lint:
```suggestion
	rule := ""

```

trie/trie.go
```diff
@@ -132,7 +140,13 @@ func (n *parent) hasParentOf(key string, split SplitFunc) bool {
...
 		case terminal:
 			// Continue down the trie
-			return child.hasParentOf(rest, split)
+			matched := child.hasParentOf(rest, split)
+			if matched {
+				rule = strings.Join([]string{child.String(), rule}, ".")
+				rule = strings.Trim(rule, ".")
+				log.PrefixedLog("Trie").Debugf("wildcard block rule '%s' matched with '%s'", rule, searchString)
+			}
+			return matched
```
ThinkChaos, line 149:
```suggestion

			return matched
```

ThinkChaos (commented):
Thanks for the PR!  

I think ideally we should bubble up the info a bit further so the logs are associated with the request, and that we can even potentially return it through extended errors, or a blocky HTTP API, which would allow the blocky CLI to let you know what the rule is without needing to check logs.  
To do that I'd probably update `ListCache.Match` to return a `map[string]string` which would use the group names as keys, and the rule that caused the block as value. From there, the compiler should show you what else needs to be updated up the stack, and down the stack you'd need to bridge `Match` and the code you've added in the caches.

That being said I think a debug like this is already a big UX improvement and can be merged if you don't want to take on extra work! No pressure either way!
In that case just fixing the lint errors is good, hopefully my suggestions do the trick :)

cache/stringcache/string_caches.go
zc-devs: That was copy-pasted from
```
log.PrefixedLog("regexCache")
```

zc-devs:
> ideally we should bubble up the info a bit further ...

Agree. I just was afraid of changing `trie.go` too much.

> To do that ...

Thanks for suggestions, I'll take it.

> I think a debug like this is already a big UX improvement and can be merged if you don't want to take on extra work

Let's do not close the issue, but merge this PR as is. I will try to propagate the info in separate PR a little bit later.

cache/stringcache/string_caches.go
ThinkChaos: Could you change that one too then please?

The examples of snake_case I'm thinking of are mostly in the resolvers, for instance:
https://github.com/0xERR0R/blocky/blob/v0.23/resolver/fqdn_only_resolver.go#L22
https://github.com/0xERR0R/blocky/blob/v0.23/resolver/query_logging_resolver.go#L18

ThinkChaos:
> Let's do not close the issue, but merge this PR as is. I will try to propagate the info in separate PR a little bit later.

Sure works for me!

cache/stringcache/string_caches.go
zc-devs: Sure, I've already done this in [8eff14f](https://github.com/0xERR0R/blocky/pull/1460/commits/8eff14fee0f98ab23b5b8bf0d9c177f999f55337) :) I just tried to explain why it was like this :)

ThinkChaos: Ah sorry missed it!

ThinkChaos (approved):

PR #1489, opened 2024-05-21
Log blocking rule in resolver

### LINKED ISSUE #1458: Show the rule which is the cause of blocked request

zc-devs:
Follow up of https://github.com/0xERR0R/blocky/pull/1460#pullrequestreview-2018555880
Closes #1458

### DIFF
--- a/cache/stringcache/string_caches.go
+++ b/cache/stringcache/string_caches.go
@@ -11,7 +11,7 @@ import (
 
 type stringCache interface {
 	elementCount() int
-	contains(searchString string) bool
+	contains(searchString string) (bool, string)
 }
 
 type cacheFactory interface {
@@ -36,12 +36,12 @@ func (cache stringMap) elementCount() int {
 	return count
 }
 
-func (cache stringMap) contains(searchString string) bool {
+func (cache stringMap) contains(searchString string) (bool, string) {
 	normalized := normalizeEntry(searchString)
 	searchLen := len(normalized)
 
 	if searchLen == 0 {
-		return false
+		return false, ""
 	}
 
 	searchBucketLen := len(cache[searchLen]) / searchLen
@@ -54,11 +54,11 @@ func (cache stringMap) contains(searchString string) bool {
 		if blockRule == normalized {
 			log.PrefixedLog("string_map").Debugf("block rule '%s' matched with '%s'", blockRule, searchString)
 
-			return true
+			return true, blockRule
 		}
 	}
 
-	return false
+	return false, ""
 }
 
 type stringCacheFactory struct {
@@ -134,16 +134,16 @@ func (cache regexCache) elementCount() int {
 	return len(cache)
 }
 
-func (cache regexCache) contains(searchString string) bool {
+func (cache regexCache) contains(searchString string) (bool, string) {
 	for _, regex := range cache {
 		if regex.MatchString(searchString) {
 			log.PrefixedLog("regex_cache").Debugf("regex '%s' matched with '%s'", regex, searchString)
 
-			return true
+			return true, regex.String()
 		}
 	}
 
-	return false
+	return false, ""
 }
 
 type regexCacheFactory struct {
@@ -197,7 +197,7 @@ func (cache wildcardCache) elementCount() int {
 	return cache.cnt
 }
 
-func (cache wildcardCache) contains(domain string) bool {
+func (cache wildcardCache) contains(domain string) (bool, string) {
 	return cache.trie.HasParentOf(domain)
 }
 
--- a/cache/stringcache/string_caches.go
+++ b/cache/stringcache/string_caches.go
@@ -198,7 +198,11 @@ func (cache wildcardCache) elementCount() int {
 }
 
 func (cache wildcardCache) contains(domain string) (bool, string) {
-	return cache.trie.HasParentOf(domain)
+	contains, path := cache.trie.HasParentOf(domain)
+	rule := strings.Join(path, ".")
+	rule = strings.Trim(rule, ".")
+
+	return contains, rule
 }
 
 type wildcardCacheFactory struct {
--- a/resolver/blocking_resolver.go
+++ b/resolver/blocking_resolver.go
@@ -371,8 +371,8 @@ func (r *BlockingResolver) handleDenylist(ctx context.Context, groupsToCheck []s
 		domain := util.ExtractDomain(question)
 		logger := logger.WithField("domain", domain)
 
-		if groups := r.matches(groupsToCheck, r.allowlistMatcher, domain); len(groups) > 0 {
-			logger.WithField("groups", groups).Debugf("domain is allowlisted")
+		if matches := r.matches(groupsToCheck, r.allowlistMatcher, domain); len(matches) > 0 {
+			logger.WithField("groups", maps.Keys(matches)).Debugf("domain is allowlisted")
 
 			resp, err := r.next.Resolve(ctx, request)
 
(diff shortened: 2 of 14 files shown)


trie/trie.go
```diff
@@ -141,14 +142,14 @@ func (n *parent) hasParentOf(key string, split SplitFunc) bool {
 
 		case terminal:
 			// Continue down the trie
-			matched := child.hasParentOf(rest, split)
+			matched, _ := child.hasParentOf(rest, split)
 			if matched {
 				rule = strings.Join([]string{child.String(), rule}, ".")
 				rule = strings.Trim(rule, ".")
 				log.PrefixedLog("trie").Debugf("wildcard block rule '%s' matched with '%s'", rule, searchString)
 			}
```
ThinkChaos, line 150:
```suggestion
			childMatch, ok := child.hasParentOf(rest, split)
			if !ok {
				return nil, false
			}
```

trie/trie.go
```diff
@@ -112,7 +113,7 @@ func (n *parent) insert(key string, split SplitFunc) {
 	}
 }
 
-func (n *parent) hasParentOf(key string, split SplitFunc) bool {
+func (n *parent) hasParentOf(key string, split SplitFunc) (bool, string) {
 	searchString := key
 	rule := ""
```
ThinkChaos, line 118:
```suggestion
	rule := make([]string, 0, len(key)) // over-allocating is better than allocating in a loop
```

resolver/blocking_resolver.go
```diff
@@ -371,8 +371,8 @@ func (r *BlockingResolver) handleDenylist(ctx context.Context, groupsToCheck []s
 		domain := util.ExtractDomain(question)
 		logger := logger.WithField("domain", domain)
 
-		if groups := r.matches(groupsToCheck, r.allowlistMatcher, domain); len(groups) > 0 {
-			logger.WithField("groups", groups).Debugf("domain is allowlisted")
+		if matches := r.matches(groupsToCheck, r.allowlistMatcher, domain); len(matches) > 0 {
+			logger.WithField("groups", maps.Keys(matches)).Debugf("domain is allowlisted")
```
ThinkChaos, line 375: I think we can change this to log all the info:
```suggestion
			logger.WithField("matches", matches).Debug("domain is allowlisted")
```

resolver/blocking_resolver.go
```diff
@@ -415,11 +421,17 @@ func (r *BlockingResolver) Resolve(ctx context.Context, request *model.Request)
 			if len(entryToCheck) > 0 {
 				logger := logger.WithField("response_entry", entryToCheck)
 
-				if groups := r.matches(groupsToCheck, r.allowlistMatcher, entryToCheck); len(groups) > 0 {
-					logger.WithField("groups", groups).Debugf("%s is allowlisted", tName)
-				} else if groups := r.matches(groupsToCheck, r.denylistMatcher, entryToCheck); len(groups) > 0 {
-					return r.handleBlocked(logger, request, request.Req.Question[0], fmt.Sprintf("BLOCKED %s (%s)", tName,
-						strings.Join(groups, ",")))
+				if matches := r.matches(groupsToCheck, r.allowlistMatcher, entryToCheck); len(matches) > 0 {
+					logger.WithField("groups", maps.Keys(matches)).Debugf("%s is allowlisted", tName)
```
ThinkChaos, line 425:
```suggestion
					logger.WithField("matches", matches).Debug("%s is allowlisted", tName)
```

resolver/blocking_resolver.go
```diff
@@ -385,8 +385,14 @@ func (r *BlockingResolver) handleDenylist(ctx context.Context, groupsToCheck []s
...
-		if groups := r.matches(groupsToCheck, r.denylistMatcher, domain); len(groups) > 0 {
-			resp, err := r.handleBlocked(logger, request, question, fmt.Sprintf("BLOCKED (%s)", strings.Join(groups, ",")))
+		if matches := r.matches(groupsToCheck, r.denylistMatcher, domain); len(matches) > 0 {
+			var entries []string
+			for group, rule := range matches {
+				entries = append(entries, strings.Join([]string{group, rule}, ":"))
+			}
+
+			msg := fmt.Sprintf("BLOCKED (%s)", strings.Join(entries, ", "))
+			resp, err := r.handleBlocked(logger, request, question, msg)
```
ThinkChaos, line 395: Can you pass `matches` to `handleBlock` instead of `msg` so we can move the formatting there and deduplicate it with `Resolve`?

resolver/blocking_resolver.go
```diff
@@ -415,11 +421,17 @@ func (r *BlockingResolver) Resolve(ctx context.Context, request *model.Request)
...
+					logger.WithField("groups", maps.Keys(matches)).Debugf("%s is allowlisted", tName)
+				} else if matches := r.matches(groupsToCheck, r.denylistMatcher, entryToCheck); len(matches) > 0 {
+					var entries []string
+					for group, rule := range matches {
+						entries = append(entries, strings.Join([]string{group, rule}, ":"))
+					}
+
+					msg := fmt.Sprintf("BLOCKED %s (%s)", tName, strings.Join(entries, ", "))
+
+					return r.handleBlocked(logger, request, request.Req.Question[0], msg)
```
ThinkChaos, line 434: Also pass `maches` to `handleBlocked` here.

resolver/blocking_resolver.go
```diff
@@ -515,12 +527,12 @@ func (r *BlockingResolver) groupsToCheckForClient(request *model.Request) []stri
...
 func (r *BlockingResolver) matches(groupsToCheck []string, m lists.Matcher,
 	domain string,
-) (group []string) {
+) (matches map[string]string) {
 	if len(groupsToCheck) > 0 {
 		return m.Match(domain, groupsToCheck)
 	}
 
-	return []string{}
+	return map[string]string{}
```
ThinkChaos, line 535: Let's make it nil while we change it since we don't need to allocate an empty map.
```suggestion
	return nil
```

ThinkChaos (commented):
Thanks for following up with this!

I put comments about mostly minor things. but some of it does require changing a bunch of code so I tried to put suggestions those.

Could you also please add a couple checks in tests that return a positive mach for the rule? Alternatively, add some basic dedicated tests for that.
If it can be any help, the way I'd do it is update the tests to not assert anything. make the function print the rule before returning. and add an `Expect(false)` at the end so ginkgo prints the output and you can copy past the rules instead of typing them out.
Some kind of snapshot testing would be nice here, but that's more work and we don't have anything like that already.

And just to be explicit about this: nothing is set in stone feel free to push back on anything :)

cache/stringcache/string_caches.go
```diff
@@ -11,7 +11,7 @@ import (
 
 type stringCache interface {
 	elementCount() int
-	contains(searchString string) bool
+	contains(searchString string) (bool, string)
```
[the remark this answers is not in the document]
zc-devs: I would not do this with `contains` method. The name implies that the main return type should be boolean. Consider error return value:
```go
contains(searchString string) (bool, err)
```
`err` should be at the end as it is not the main response from method.

Could we rename `contains` to `match` and get rid of boolean?
```go
// returns rule if matched, otherwise empty string
func match(searchString string) (string)
```

resolver/blocking_resolver.go
zc-devs: resolver/blocking_resolver.go:383 passes a string:
```go
resp, err := r.handleBlocked(logger, request, question, "BLOCKED (ALLOWLIST ONLY)")
```

Could we return custom type from `r.matches(...)` and override `String()` method?

zc-devs:
> Could you also please add a couple checks in tests ...
> ... copy past the rules instead of typing them out.

Sorry, I don't understand what you want. There are checks for the rules in `blocking_resolver_test.go`. And I've copy-pasted tests, not typed manually. Could you provide an example?

0xERR0R:
replaced with #2091

PR #1517, opened 2024-06-11
Add Timescale DB with Hypertable and Retention support

m4rcs:
With this pull I want to add support of Timescale DB to blocky. 

It uses the postgres connection (`queryLog.target=postgres://...`) of GORM to write data, but uses [hypertabels](https://docs.timescale.com/use-timescale/latest/hypertables/) to store time-series data and uses Timescales-native [retention](https://docs.timescale.com/use-timescale/latest/data-retention/create-a-retention-policy/) with `queryLog.type=timescale`.

### DIFF
--- a/querylog/database_writer.go
+++ b/querylog/database_writer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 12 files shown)


querylog/database_writer.go
```diff
@@ -113,7 +114,30 @@ func databaseMigration(db *gorm.DB) error {
 		}
 
 	case "postgres":
-		return db.Exec("ALTER TABLE " + tableName + " ADD column if not exists id serial primary key").Error
+		return db.Exec("ALTER TABLE " + tableName + " ADD column if not exists id bigserial primary key").Error
```
0xERR0R, line 117: Why did the type change from serial to bigserial?

0xERR0R (commented):
I didn't test it, but it looks good so far for me :+1: . We do have some e2e tests with testcontainers (real database as docker container) for query log. Maybe you could create e2e test for timescale DB? If not, I can also do it.

querylog/database_writer.go
m4rcs: Depending on retention time I think is possible to reach the max value of serial. That's why I changed it to bigserial.

m4rcs:
I will try to add some e2e tests.

0xERR0R (approved):
LGTM, Thank you for your work!

0xERR0R:
Please fix the linter error to make build green again, then we can merge

m4rcs:
The linter error is fixed. I'll try to increase coverage later :)

0xERR0R:
> The linter error is fixed. I'll try to increase coverage later :)

Should we merge it now or do you want to add some tests with this PR?

PR #1560, opened 2024-07-28
Update metrics

SuperQ:
* Use Prometheus [naming best practices](https://prometheus.io/docs/practices/naming/).
* Enable native histograms.
* Update dashboards to support new and old names.


0xERR0R (approved):
:+1: LGTM 
thanks for your contribution! :handshake:

SuperQ:
@0xERR0R I was looking at the package layout for these metrics. I was going to open a PR to improve performance by moving the metrics to package local use. This would also eliminate the need for a bunch of inter-package channels and queuing. 

But, I'm not sure this architectural change would be accepted.

0xERR0R:
@SuperQ I'm not sure if I fully understand what you mean, but you can always create a PR with a proposal and we can discuss you solution. If it makes things more performant and simpler -> that would be great

PR #1577, opened 2024-08-19
chore(build): update golangci-lint

0xERR0R:


ThinkChaos (approved):

Findmycrypto (commented):
I hope I can somehow get an answer how my asset disappeared from my metamask wallet I see it in new contract after I bridged from base to Etherium I signed some thing on April 10 or 11 and I'm completely lost just want my asset bridge to have never happened I am not smart enough needed help idk how secure this system can possibly work after such an awful experience for me hoping my asset reverses back to the origin address idk why I did some thing that changed it my trump maga I bought at 3.33 per share is just gone into coding I don't understand

PR #1578, opened 2024-08-19
Use package local metrics

SuperQ:
Improve performance of metrics by moving them to the package that needs them. This reduces the overhead to a simple atomic increment for basic counters like cache hits/misses. This also uses `promauto` to avoid the second step of having to register metrics.

### DIFF
--- a/evt/events.go
+++ b/evt/events.go
@@ -20,12 +20,6 @@ const (
 	// CachingPrefetchCacheHit fires if a query result was found in the prefetch cache, Parameter: domain name
 	CachingPrefetchCacheHit = "caching:prefetchHit"
 
-	// CachingResultCacheHit fires, if a query result was found in the cache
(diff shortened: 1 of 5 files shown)


ThinkChaos (approved):
LGTM, would be nice to move the other metrics too, but @0xERR0R should chime in, as mentioned in #1560, before you do more work :)

ThinkChaos:
Looks like the end to end test for metrics failed though, so it might now fully work at the moment

SuperQ:
Ahh, yes, this is because `metrics/metrics.go` creates an isolated metrics registry instead of using the standard default registry.

Also because it's not created by the `main` or exported it's not easy to pass the registry around.

The simplest solution would be to use the default registry.

SuperQ:
I went another direction and simply exported the registry from the metrics package.

0xERR0R (approved):
:+1:

PR #1595, opened 2024-09-05
refactor: add service package and start splitting HTTP handling

ThinkChaos:
Progress for #1206

Can't reopen 1427, so making a new PR.

EDIT: at least this way we get a codecov patch diff percentage back :smile:


yodog:
is there a testing version of this ? 

i was just trying to disable the api endpoint but keeping DoH enabled.

PR #1605, opened 2024-09-13
refactor: Use package local metrics

SuperQ:
Part 2 (#1578)

Improve performance of metrics by moving them to the package that needs them. This reduces the overhead to a simple atomic increment for basic counters like cache hits/misses. This also uses `promauto` to avoid the second step of having to register metrics.

### DIFF
--- a/lists/downloader_test.go
+++ b/lists/downloader_test.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"github.com/0xERR0R/blocky/config"
-	. "github.com/0xERR0R/blocky/evt"
 	. "github.com/0xERR0R/blocky/helpertest"
 	"github.com/0xERR0R/blocky/log"
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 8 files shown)


SuperQ:
I'm working on an improvement to the tests using the client_golang testutil.

SuperQ:
Not really stale, but I haven't had a chance to work on the coverage change.

SuperQ:
@0xERR0R, how do you feel about the coverage change here?

lists/downloader_test.go
```diff
@@ -22,27 +21,16 @@ import (
...
+		sut        *httpDownloader
+		loggerHook *test.Hook
 	)
 	BeforeEach(func() {
 		var err error
 
 		sutConfig, err = config.WithDefaults[config.Downloader]()
 		Expect(err).Should(Succeed())
 
-		failedDownloadCountEvtChannel = make(chan string, 5)
```
0xERR0R, line 36: now we don't test if metrics counters are being incremented. Maybe we can try to read the counter value and assert it?

0xERR0R (changes requested):

lists/downloader_test.go
SuperQ: Yes, that is something we could test for. But I find it not very useful. The Prometheus client `Inc()` type functions are basically a simple Go atomic increment. There's basically no way for it to go wrong in a way that testing would help with.

SuperQ:
Need to fix up the one duplicate metric.

SuperQ:
Still trying to find time to work on this.

SuperQ:
Sorry, still no time for this one.

PR #1614, opened 2024-09-25
Ability to disable "Special Use Domain Name" blocking

mapl:
https://0xerr0r.github.io/blocky/latest/configuration/#special-use-domain-names

setting enable to false of specialUseDomains will unblock all Special Use Domain Names
```
specialUseDomains:
  enable: false
```

### DIFF
--- a/config/sudn.go
+++ b/config/sudn.go
@@ -16,7 +16,6 @@ type SUDN struct {
 
 // IsEnabled implements `config.Configurable`.
 func (c *SUDN) IsEnabled() bool {
-	
 	return c.Enable
 }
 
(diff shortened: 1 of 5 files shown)


ThinkChaos:
Thanks for making the change!

We'd also need to update the docs (`docs/configuration.md` and `docs/config.yml`).  
And ideally add basic a test for `IsEnabled` (see `Describe("IsEnabled", ...)` in `config/redis_test.go` for an example).  
If that's too much bother, let me know and I can do it.

Can you also explain a bit about your use case?

mapl:
I would love to see you making the subsequent changes to the docs and test if you like
The use case is a dns migration where I successfully used blocky in the past to forward any dns query to the destination dns server. I really liked the exposed Prometheus metrics to track the remaining dns clients.

PR #1624, opened 2024-10-08
fix(docs): mark config mount as read-only

zygm0nt:
This default config results in the following error:

```
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/home/x/blocky/config.yml" to rootfs at "/app/config.yml": mount /home/x/blocky/config.yml:/app/config.yml (via /proc/self/fd/6), flags: 0x5000: not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -84,7 +84,7 @@ services:
       # Optional to synchronize the log timestamp with host
       - /etc/localtime:/etc/localtime:ro
       # config file
-      - ./config.yml:/app/config.yml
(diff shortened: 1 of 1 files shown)


ThinkChaos (approved):
Hi, thanks for the PR!

I'm a bit surprised that's the fix, but it can't hurt :)

Issue #1606, opened 2024-09-16
"CORS or Network issue" error in browser when List refresh blocky 


sborca:
Hello! 
I run blocky app on an lxc container on Nokia router with OpenWRT as host OS, and when I trigger a List refresh action in WEB UI I got the error("CORS or Network issue"):
![CORSErr](https://github.com/user-attachments/assets/bab7f715-c8c6-42fd-9d8c-68d76d07e7a0)
 
All the other API requests works perfectly.
Do you know why? 
My config.yml file looks like this:

blocking:
  denylists:
    ads:
      - https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
      - https://s3.amazonaws.com/lists.disconnect.me/simple_ad.txt
      - http://sysctl.org/cameleon/hosts
      - https://s3.amazonaws.com/lists.disconnect.me/simple_tracking.txt
  clientGroupsBlock:
    default:
      - ads
  loading:
    refreshPeriod: 24h
    downloads:
      timeout: 60s
    strategy: fast
ports:
  dns: 1053
  http: 4000

Thank you very much!

sborca:
Hello! any answer for my question?

TheoTechnicguy:
Hi! I am having difficulties diagnosing and reproducing your issue. Can you provide other useful information like the logs of blocky and your browser?

To help out the community, I would like to kindly request formatting your configuration file. GitHub supports Markdown, so you can delimit your configuration file with triple backticks (see <https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code>)

Doing so will help us save time formatting the configuration file and allow us to invest this time resolving the issue.

The reference configuration file I used for my tests is the following. Can you confirm it reflects yours?

```yaml
bootstrapDns:
- upstream: https://cloudflare-dns.com/dns-query
  ips:
  - 104.16.248.249
  - 104.16.249.249
upstreams:
  timeout: 2s
  strategy: strict
  groups:
    default:
    - tcp+udp:1.1.1.1

blocking:
  denylists:
    ads:
    - https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
    - https://s3.amazonaws.com/lists.disconnect.me/simple_ad.txt
    - http://sysctl.org/cameleon/hosts
    - https://s3.amazonaws.com/lists.disconnect.me/simple_tracking.txt
  clientGroupsBlock:
    default:
[...]

PR #1653, opened 2024-12-04
feat: configurable file write timeouts

TheoTechnicguy:
This pull request introduces changes to the HTTP server configuration to make it more flexible and configurable. The change include adding a new configuration parameter for write timeouts, updating the HTTP server initialization to use this new parameter, and adjusting the server creation logic accordingly.

This PR answers issue #1606.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -405,9 +405,10 @@ func recoverToError(do func(context.Context) error, onPanic func(any) error) fun
 }
 
 type Downloader struct {
-	Timeout  Duration `yaml:"timeout" default:"5s"`
-	Attempts uint     `yaml:"attempts" default:"3"`
-	Cooldown Duration `yaml:"cooldown" default:"500ms"`
+	Timeout      Duration `yaml:"timeout" default:"5s"`
+	WriteTimeout Duration `yaml:"writeTimeout" default:"20s"`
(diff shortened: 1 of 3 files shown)


TheoTechnicguy:
I have some lingering questions. Should the read timeouts that are currently hardcoded also be migrated to configurable values (possibly in a separate PR)?
I also did not find any test cases for the parts I touched. Did I miss them?

0xERR0R:
> I have some lingering questions. Should the read timeouts that are currently hardcoded also be migrated to configurable values (possibly in a separate PR)? I also did not find any test cases for the parts I touched. Did I miss them?

I would say, it is always better to have some parameters configurable for power user and default values which are ok for the most cases ;)

We don't have any tests which cover http timeouts

0xERR0R (approved):
LGTM, Thanks for your contribution! Should I merge your change or do you want to add something?

TheoTechnicguy:
> I would say, it is always better to have some parameters configurable for power user and default values which are ok for the most cases ;)

I'll do that.

> Should I merge your change or do you want to add something?

IMO, this PR is ready for merging; I intend to add both configuration parameters in separate PRs. If you prefer having them in one bigger PR, I can add the commits here.

Issue #1666, opened 2024-12-16
Health check not compatible with containers that use a non standard host IP binding


cosmikwolf:
When I am running blocky, it reports that it is unhealthy, even though it is running correctly:
```bash
ubuntu@server /m/d/d/blocky> docker ps
CONTAINER ID   IMAGE                           COMMAND                  CREATED         STATUS                     PORTS                                                                                                                 NAMES
9227503fa980   ghcr.io/0xerr0r/blocky:latest   "/app/blocky"            4 minutes ago   Up 4 minutes (unhealthy)   192.168.20.96:53->53/tcp, 192.168.50.47:4000->4000/tcp, 192.168.20.96:53->53/udp                                      blocky
```

The health check command is failing:
```bash
ubuntu@miniserver /m/d/d/blocky> docker compose exec blocky /app/blocky healthcheck
NOT OK
Error: dial tcp 127.0.0.1:53: connect: connection refused
```

I believe this is because I am using a non standard host binding port in my docker compose file:

My dockerfile:
```yaml
services:
  blocky:
    image: "ghcr.io/0xerr0r/blocky:latest"
    container_name: blocky
    restart: unless-stopped
    networks:
      - dns
    # Optional the instance hostname for logging purpose
    hostname: "blocky"
    ports:
      - "53:53/tcp"
      - "53:53/udp"
      - "4000:4000/tcp"
    environment:
      - TZ=PST
    volumes:
      # Optional to synchronize the log timestamp with host
      - /etc/localtime:/etc/localtime:ro
      # config file
      - ./config.yml:/app/config.yml

networks:
  dns:
    driver: bridge
    driver_opts:
      com.docker.network.bridge.host_binding_ipv4: "192.168.20.96"
```

If there was an --ip flag on healthcheck, I would be able to set the bound IP address. Maybe there is already an env variable I am unaware of?

PR #1667, opened 2024-12-16
added ip argument to health check

cosmikwolf:
added ip address argument to health check to address the bug I submitted earlier:

https://github.com/0xERR0R/blocky/issues/1666

There may be issues with this pull request that I will resolve, the test suite did not seem to run properly on my M1 Mac, so I am hoping to see any issues that exist with your CI tools

### DIFF
--- a/cmd/healthcheck.go
+++ b/cmd/healthcheck.go
@@ -9,8 +9,8 @@ import (
 )
 
 const (
-  defaultDNSPort = 53
-  defaultIPAddress = "127.0.0.1"
+	defaultDNSPort   = 53
+	defaultIPAddress = "127.0.0.1"
 )
 
 func NewHealthcheckCommand() *cobra.Command {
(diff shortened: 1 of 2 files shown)


0xERR0R (approved):
:+1: thanks

polarathene:
> There may be issues with this pull request that I will resolve

Just chiming in to say this doesn't appear to be documented?

I think the [reference](https://github.com/0xERR0R/blocky/issues/1666#issue-2743430185) example is lacking information on how to reproduce that failure, as there is nothing to indicate the default bind to `127.0.0.1` wouldn't be valid?

You have a custom `dns` network for the container but that's just setting the binding IP on the host, so I'm not sure what the IP would have been set to in [blocky `/app/config.yaml`](https://0xerr0r.github.io/blocky/v0.21/configuration/#ports-configuration)? (_which I assume is where this might be worth drawing attention to in the docs_)

```yaml
ports:
  dns: 198.51.100.42 # IP reserved for documentation (RFC 5737)
```

I could reproduce if I set that to another valid IP in the container (_like a static IP assigned to that custom bridge network_), or if I switched to `network_mode: host` and configured to bind `ports.dns` to my host IP.

Issue #1700, opened 2025-02-02
Old/stale/wrong link in "Additional Information" documentation page


erictelkkala:
The example [allowlist](https://0xerr0r.github.io/blocky/latest/additional_information/#allowlists) shows the link preview to be `https://github.com/anudeepND/whitelist`, but that page does not exist. The link actually points to `https://github.com/anudeepND/allowlist`. Correct page would be `https://github.com/anudeepND/whitelist` [link](https://github.com/anudeepND/whitelist).

I'll make a quick PR for this, since it is a minor fix to the URL

PR #1701, opened 2025-02-02
fix(docs): fix stale link in example Allowlists (#1700)

erictelkkala:
Fixes: #1700

- Accidentally removes redundant white-spaces as well?


0xERR0R (approved):
Thanks!

PR #1722, opened 2025-03-04
[feature] Update Grafana dashboard

maxim-mityutko:
- add datasource selector (fixes a problem when Grafana is used with data source other then Prometheus, eg VictoriaMetrics)
- add the possibility to select specific pod (only affects graphs)
- fix dashboard name
- add tags
- fix alignment of the panels


maxim-mityutko:
Hi @0xERR0R , do you think these changes make sense and can be merged? or is there anything else that is missing?

0xERR0R:
Hey @maxim-mityutko,

sorry, didn't find time to test it on my environment yet. I'll try to do it this week. I saw in the diff, you introduced filtering for "pod", it is not clear for me (from the json diff) if it is mandatory now or not. Since we have non k8s installations, we can introduce a pod filter but it should be optional.

0xERR0R (approved):
Thanks for your contribution! I imported the dashboard and it worked nice. The pod filter is always displayed, even if blocky runs not on k8s. I think Grafana has no feature to conditionally hide empty filters. I'd say, we merge it, this filter will work for k8s installations and has no impact on non-k8s envs.

maxim-mityutko:
Awesome! Thanks  @0xERR0R

PR #1729, opened 2025-03-07
docs(grafana): overview supporting native histograms

krohrsb:
This updates the request duration (upstream) query on blocky-grafana to support native histograms in prometheus. Normal generates the `_bucket` metric, however native do not and manage it all under one metric.


0xERR0R (approved):
thanks!

PR #1732, opened 2025-03-09
feat: add "cache-control" http header to DoH response

### LINKED ISSUE #526: [Feature Request] Add "cache-control" http header on DoH server
This http header is currently not implemented in Blocky DoH server.

Some famous DoH servers like those:

- AdGuard
- Quad9
- Quad101(Powered by https://github.com/m13253/dns-over-https)
- Google
- Applied Privacy DNS
- DNS.SB
- DNS Forge

They do return `cache-control` http header like the example mentioned in [RFC 8484 (DNS Queries over HTTPS (DoH))](https://datatracker.ietf.org/doc/html/rfc8484) [4.2.2 HTTP Response Example](https://datatracker.ietf.org/doc/html/rfc8484#section-4.2.2)

More info: [5.1 Cache Interaction](https://datatracker.ietf.org/doc/html/rfc8484#section-5.1).

Just FYI, Cloudflare, NextDNS and OpenDNS doesn't do that. At least, not yet.

0xERR0R:
closes #526

### DIFF
--- a/server/server_endpoints.go
+++ b/server/server_endpoints.go
@@ -140,7 +140,7 @@ func (r httpMsgWriter) WriteMsg(msg *dns.Msg) error {
 
 	// https://www.rfc-editor.org/rfc/rfc8484#section-5.1
 	// get the smallest TTL from answer
-	r.rw.Header().Set(cacheControlHeader, fmt.Sprintf("max-age=%d", getSmallestTTlFromAnswer(msg)))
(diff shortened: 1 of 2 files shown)


brokoler:
Hello @0xERR0R 
Updated to 0.26 and recognized that Firefox using DoH with blocky still does not use custom caching TTL values.
My configuration.yml includes following values which are working when DoH is not used:
`caching:
  minTime: 24h
  maxTime: 24h`

With DoH being used Firefox does show the default TTL values from websites:

grafik

brokoler:
Probably related to:
https://github.com/0xERR0R/blocky/issues/1777

PR #1753, opened 2025-04-06
New feature: Cache exclude

macsign:
New feature - Cache exclusions rules configuration

Example:
image

Any request with a domain ending with any suffix listed in exclusions rules won't be cached at all


0xERR0R:
Thanks for your contribution. I'm thinking about to make this feature "more flexible". Currently it covers only domain suffixes. Maybe we should introduce regex instead? So you can define one ore more regex to exclude domains from caching. For example `^.*\.lan$` -> every domain which ends with ".lan" or maybe `.googleapi.com` -> not cache google API

What do you thing?

docs/configuration.md
```diff
@@ -621,6 +622,10 @@ With following parameters you can tune the caching behavior:
       minTime: 5m
       maxTime: 30m
       prefetching: true
+      exclude:
+        - lan
+        - local
```
0xERR0R, line 627: 'exclude' property allows only values which starts and ends with a slash -> so this example is not valid, right? Please provide only working example

0xERR0R (changes requested):

docs/configuration.md
macsign: Done

0xERR0R (approved):

PR #1868, opened 2025-09-16
test(metrics): Prometheus metrics

zeilenschubser:
# Description
> Add testing for metrics

I am open to suggestions, used `"testing"` instead of ginkgo because i don't know better.

# What i did
- [x] check if every metric is reachable
- [x] check when a new metric is added, that it has to be tested as well
- [x] updated wrong doc names

### DIFF
--- a/metrics/metrics_test.go
+++ b/metrics/metrics_test.go
@@ -1,0 +1,196 @@
+package metrics_test
+
+import (
+	"context"
+	"testing"
+	"time"
+	"net"
+
+	"github.com/0xERR0R/blocky/config"
+	"github.com/0xERR0R/blocky/evt"
+	"github.com/0xERR0R/blocky/lists"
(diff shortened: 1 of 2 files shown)


metrics/metrics_test.go
```diff
@@ -0,0 +1,196 @@
...
+		"process_max_fds",
+		"process_cpu_seconds_total",
+		"process_virtual_memory_max_bytes",
+		"process_network_transmit_bytes_total",
+		"process_open_fds",
+		"process_resident_memory_bytes",
+		"process_network_receive_bytes_total",
+		// promhttp
+		"promhttp_metric_handler_requests_total",
+		"promhttp_metric_handler_requests_in_flight",
```
0xERR0R, line 103: I would suggest to remove these metrics from the test. We should test only blocky's functionality. Those metrics are generated automatically by the prometheus library and this is not under our control

Issue #1867, opened 2025-09-15
Panic when CNAME points to external domains


Renerick:
When CNAME record in the custom zone file points to external domains, blocky panics and returns no respone

Custom DNS config (it is provisioned by a script, so formatting is not "human")

```yaml
customDNS:
  customTTL: 30m
  rewrite:
    [REDACTED]: internal
  zone: '$ORIGIN internal.

    a 3600 A 1.1.1.1

    c 3600 CNAME example.com.


    '
```

Testing

```sh
$ dig a.internal @10.10.10.10

; <<>> DiG 9.18.39 <<>> a.internal @10.10.10.10
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 7011
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;a.internal.                    IN      A

;; ANSWER SECTION:
a.internal.             3600    IN      A       1.1.1.1

;; Query time: 1 msec
;; SERVER: 10.10.10.10#53(10.10.10.10) (UDP)
;; WHEN: Tue Sep 16 01:55:54 +04 2025
;; MSG SIZE  rcvd: 44

$ dig c.internal @10.10.10.10

; <<>> DiG 9.18.39 <<>> c.internal @10.10.10.10
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 573
;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0
;; WARNING: recursion requested but not available

;; QUESTION SECTION:
;c.internal.                    IN      A

;; Query time: 1 msec
;; SERVER: 10.10.10.10#53(10.10.10.10) (UDP)
;; WHEN: Tue Sep 16 01:55:57 +04 2025
;; MSG SIZE  rcvd: 28
```

Logs:

```log
Sep 15 21:55:54 dns-1 blocky[3760]: [2025-09-15 21:55:54] TRACE new incoming request client_id= client_ip=10.10.0.1 client_request_id=7011 protocol=UDP question=A (a.internal.) req_id=09894299-7520-4f9c-9fd2-f01c87012c99
[...]

PR #1869, opened 2025-09-16
fix: separate resolver for custom CNAME records from the main chain

Renerick:
This a PoC of a fix for #1867


0xERR0R:
replaced with #1884

PR #1876, opened 2025-09-30
Skip Adblock identifier header and comments

### LINKED ISSUE #950: Feature Request: Compressed/Regex Blocklist/Allowlist Syntax Support
With the latest Pi-hole update the developers added basic AdBlock style syntax support for their adlists.

I hope the same can be implemented in blocky, as it remains the only dns software without support for a compressed format.
AdGuard Home, pfblocker, Pi-hole, Technitium, eBlocker all have 1 or more compressed formats supported.

Preferably, I would like the adblock style to be added just like Pi-hole. But really, any is fine (wildcard, adguard, adblock).

Occasionally, blocky will miss domains that AdGuard Home/Pi-hole catches (using the same lists) because it only supports domains/hosts. 

Another problem with domains/hosts is that not all lists support it. Ive run into cases where I cant use a list at all in blocky because the maintainer dropped it/never added it in the first place.
This used to be the case as well with Pi-hole.

Another voice (oisd maintainer): https://www.reddit.com/r/oisd_blocklist/comments/11s6dov/important_update_changes_to_blocklist_syntax/
Pi-hole's implementation: https://pi-hole.net/blog/2023/03/22/pi-hole-ftl-v5-22-web-v5-19-and-core-v5-16-1-released/#page-content

metalinspired:
This should close #950 
I say should because I haven't written anything in Go in over five years 😁

### DIFF
--- a/lists/parsers/lines.go
+++ b/lists/parsers/lines.go
@@ -70,7 +70,7 @@ func (l *lines) Next(ctx context.Context) (string, error) {
 			continue // empty line
 		}
 
-		if strings.HasPrefix(text, "!") || (l.lineNo == 1 && strings.HasPrefix(text, "[Adblock" {
+		if strings.HasPrefix(text, "!") || (l.lineNo == 1 && strings.HasPrefix(text, "[Adblock")) {
 			continue // Adblock comment or identifier header
 		}
 


lists/parsers/lines.go
```diff
@@ -70,6 +70,10 @@ func (l *lines) Next(ctx context.Context) (string, error) {
 			continue // empty line
 		}
 
+		if text[0:1] == "!" || (l.lineNo == 1 && len(text) >= 8 && text[0:8] == "[Adblock") {
```
0xERR0R, line 73:
```suggestion
		if strings.HasPrefix(text, "!") || (l.lineNo == 1 && strings.HasPrefix(text, "[Adblock" {
```

0xERR0R: Imho, "HasPrefix" is more readable than array operators

metalinspired:
Thanks for taking the time to look into the PR.
I may have jumped the gun on closing #950 as I forgot there is also @@ (whitelist) syntax.
Should we ignore @@ in block lists or try to extract them and add them to whitelist?

0xERR0R:
I would first ignore it. Currently there is no support for the adblock format. Maybe sometime we can implement it properly with own parser

metalinspired:
Then this PR is done.
I have a few ideas on how to deal with @@ but I have to first refresh my Go knowledge 😄

0xERR0R:
looks like it is not compiling now

metalinspired:
I rebased instead of merging and it removed two parenthesis 🤦🏻 
Fixed.

PR #1904, opened 2025-10-22
fix: TCP response truncation

TimQuelch:
Currently a TCP response is truncated to the maximum size advertised for UDP. This means that a client cannot effectively fallback to TCP after receiving a truncated UDP response, as the resulting TCP response is also truncated.

This change ensures that only UDP responses are truncated

### DIFF
--- a/e2e/caching_test.go
+++ b/e2e/caching_test.go
@@ -1,0 +1,367 @@
+package e2e
+
+import (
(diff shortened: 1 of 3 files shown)


0xERR0R (approved):
Thank you for your contribution :+1:

PR #1951, opened 2025-12-02
refactor: clean up obsolete Go syntax

sonjek:
During the code review, I noticed several outdated syntax patterns that can be replaced with more modern alternatives.
For example, interface{} can be replaced to any. Array/slice iteration and searching can be simplified.


0xERR0R (approved):
Thanks, LGTM :+1:

PR #1952, opened 2025-12-03
fix: use context based versions of net.Listen and net.Dial

sonjek:
The context based versions of net.Listen and net.Dial allow to cancel operations, set timeouts, and prevent resource leaks


0xERR0R (approved):
:+1:

PR #2080, opened 2026-06-03
feat(querylog): add SQLite query log target

### LINKED ISSUE #585: [Feature Request] Add primary key by default on DNS query log database table
We use Digital Ocean's managed MySQL instances for a lot of our database needs, including wanting to log queries to our existing instance. One issue with this is they require any table to have a Primary Key. Would it be possible to add that natively to logging to MySQL databases? We ended up writing a exporter from file to MySQL to get around this but it would be nice to have natively. 

Some of their documentation: https://docs.digitalocean.com/products/databases/mysql/how-to/create-primary-keys/

0xERR0R:
## What

Adds a new `sqlite` query log target, so query logs can be written to a local SQLite database file with **no external database required**.

- New `queryLog.type: sqlite` with the target being a local file path (e.g. `/var/lib/blocky/querylog.db`); the file and its parent directory are created automatically.
- Switches the SQLite driver from the cgo-based `gorm.io/driver/sqlite` (mattn) to the pure-Go `github.com/glebarez/sqlite` (modernc), so it works in `CGO_ENABLED=0` release/Docker builds.
- Opens the DB in **WAL mode** automatically (with `busy_timeout`) so external tools can read it concurrently while blocky writes.
- Retention (`logRetentionDays`) and bulk flushing work the same as the other DB targets.
- Docs + e2e test + unit tests included.

## Hardening (from review)

- **DSN path safety** — the path is percent-encoded before going into the `file:` URI, so a target containing `?`, `#` or `%` is opened verbatim instead of being truncated (which also silently disabled WAL) or mis-decoded.
- **Single connection** — the sqlite pool is pinned to one connection so blocky's own periodic flush and retention cleanup can't collide into `SQLITE_BUSY`.
- **Typed `QueryLogType`** — `NewDatabaseWriter`/`databaseMigration` take the `config.QueryLogType` enum instead of magic strings.
- Deterministic test ordering and a teardown fix (cancel context before closing the DB).

## ⚠️ Note: incidental PostgreSQL primary-key fix

While threading the typed enum through the migration code, this PR also **restores the `id` primary key on the PostgreSQL query-log table**, which had silently regressed.

- The `id` PK was originally added in #609 (v0.20, closes #585) because some managed-DB providers (e.g. DigitalOcean MySQL) **require every table to have a primary key**. It applied to both mysql and postgres.
- The original code switched on `db.Config.Name()`, which returns `"postgres"` for the postgres dialector — so `case "postgres"` matched and the migration ran.
- In **#1517 (v0.25)** the Timescale work changed the switch to a passed-in `dbType` string to disambiguate postgresql from timescale (they share the postgres dialector). It passed `"postgresql"` but left `case "postgres"`, so from v0.25 on the `ALTER TABLE ... ADD id` **silently stopped running for PostgreSQL** (mysql/timescale were unaffected).
- This PR keys the migration on the typed enum (`QueryLogTypePostgresql`), so the migration matches in production again — restoring the v0.20 behaviour.

**Upgrade impact:** existing PostgreSQL deployments will run a one-time, idempotent `ALTER TABLE log_entries ADD column if not exists id bigserial primary key` on the next startup after upgrading.


docs/configuration.md
```diff
@@ -979,11 +979,33 @@ You can select one of following query log types:
...
+
+Blocky opens the database in **WAL (Write-Ahead Logging) mode** automatically; you do not need to configure this. As a result the database is written as three files next to each other: `querylog.db`, `querylog.db-wal` and `querylog.db-shm`. When running in Docker, mount the **directory** (not just the `.db` file) as a volume so all three files persist, and include all three in any backup.
+
+WAL requires a normal local filesystem. It is **not** supported on network filesystems (NFS, SMB/CIFS); placing the database on a network share can cause lock errors or corruption, so use local/block storage.
+
+Retention (`logRetentionDays`) works the same as for the other database targets: entries older than the limit are deleted. Note that SQLite does not hand freed space back to the operating system on its own, so the `.db` file does not shrink after a deletion — the space is reused for new entries. Run `VACUUM` manually if you need to reclaim disk space.
```
immanuelfodor, line 998: Is it safe to run VACUUM in a cron job periodically eg. overnight?

0xERR0R: It should be safe. Since blocky deletes data periodically (retention), the deleted space will be reused by the sqlite db. So I would assume, if you have roughly the same amount of queries every day, the vacuum is not necessary. The vacuum is only needed if you had a spike of data and you want to reclaim it (basically to shink the data file).

immanuelfodor: This is very good intel, thank you! :pray: 
It might be worth mentioning in the docs, though

PR #2123, opened 2026-06-18
fix(querylog): add loong64 to unsupported list

### LINKED ISSUE #2122: loong64 build failure due to modernc.org/libc v1.22.5
I was working on the upgrade for Alpine aports and ran into build failures on loong64 with modernc.org/libc being pinned to v1.22.5 because of glebarez/sqlite I believe.  This version doesn't have support for loong64.  I noticed that you handle similar cases with querylog/database_writer_sqlite.go and querylog/database_writer_sqlite_unsupported.go and I was able to add a patch to include loong64 for my build.  

I have submitted a PR with the small patch I applied if it is useful.

rtmongold:
Closes #2122

modernc.org/libc v1.22.5 does not support loong64

add loong64 to querylog/database_writer_sqlite.go and querylog/database_writer_sqlite_unsupported.go exclude list.


0xERR0R (approved):
Thanks for your contribution, LGTM 👍

PR #2144, opened 2026-06-25
feat(queryLog): add dnstap as a query log destination

### LINKED ISSUE #2104: querylog: support of dnstap
dnstap (https://dnstap.info/) should be supported as destination

rtmongold:
Closes #2104

Adds `queryLog.type: dnstap` so blocky can stream client query/response events to an external collector over a Unix socket or TCP. Uses golang-dnstap's SocketWriter and emits `CLIENT_RESPONSE` frames with wire-format query and response payloads, client address, timestamps, and blocky's instance identity.

Config looks like the other query log targets:

```yaml
queryLog:
  type: dnstap
  target: unix:/var/run/dnstap.sock
  flushInterval: 30s
```

target also accepts tcp://host:port or a bare /path (treated as Unix). Invalid targets fall back to console logging like the other writers.

A few things worth knowing up front:

fields and logRetentionDays don't apply here — it's a live stream, not stored logs. Retention cleanup is skipped for dnstap (same idea as timescale).
log.privacy obfuscation does not touch dnstap wire payloads; you're getting full DNS messages.
DoH/DoT queries show up as TCP in the socket protocol field since that's what blocky sees at the resolver layer.
Tested locally with dnstap -u /tmp/dnstap.sock -q against a Unix socket target.


resolver/query_logging_resolver.go
```diff
@@ -123,11 +125,16 @@ func NewQueryLoggingResolver(ctx context.Context, cfg config.QueryLog) (*QueryLo
 
 	var writer querylog.Writer
 
-	err := retry.Do(
+	instanceID, err := readInstanceID("/etc/hostname")
```
0xERR0R, line 129: Will it work on windows?

0xERR0R (commented):
I created a branch with some fixes (see review/dnstap-blocker-fixes). This should be only for discussion purposes, following review points were addressed:

1. Non-blocking writer — dnstap's WriteFrame blocks indefinitely when the collector is unreachable, and it ran in the resolver's single writeLog goroutine → logging stalled and ctx.Done() was never seen, so shutdown hung. The writer now owns a dedicated goroutine fed by a bounded channel; Write enqueues without blocking (drops when full), keeping writeLog responsive.
2. Lifecycle close — the socket was never closed (the only Close() path, CleanUp(), is gated off for dnstap), leaking an fd + goroutine and losing buffered frames on shutdown. Added Close() plus an io.Closer-based shutdown hook in writeLog; CleanUp() is now a no-op (no retention for a streaming export). DB/CSV writers unaffected.
3. Visible failures — the dnstap library's Logger was unset (→ its null logger), so a down collector failed silently. Added an adapter onto blocky's logger so reconnect/write errors show up.
4. Target validation — parseDnstapTarget accepted tcp:// (→ :0), unix: (empty path) and relative unix paths without error, surfacing later as #1. Now validates host/port and requires an absolute unix path.

What do you think about it? Should we fix some of them?

resolver/query_logging_resolver.go
rtmongold: This was moved from line 149 before writer creation because NewDnstapWriter needs instanceID at init. readInstanceID still uses  os.Hostname() fallback for Windows, behavior is unchanged.

from line 308-322 on main
```
func readInstanceID(file string) (string, error) {
	// Prefer /etc/hostname over os.Hostname to allow easy differentiation in a Docker Swarm
	// See details in https://github.com/0xERR0R/blocky/pull/756
	hn, fErr := os.ReadFile(file)
	if fErr == nil {
		return strings.TrimSpace(string(hn)), nil
	}

	hostname, osErr := os.Hostname()
	if osErr == nil {
		return hostname, nil
	}

	return "", fmt.Errorf("cannot determine instance ID: %w", errors.Join(fErr, osErr))
}
```

rtmongold: Thanks again for review/dnstap-blocker-fixes — I’d like to incorporate all four robustness fixes from your branch.

On top of that, I’m planning a few small follow-ups that I think add ops value:

Throttled drop logging — same power-of-ten pattern, and prometheus counter as resolver/stats_resolver.go potential further shutdown hardening
I’ll cherry-pick your commit onto my branch and push the follow-ups for you to review.

rtmongold: added a few things for review:
to mimic stats_resolver.go
- Add blocky_dnstap_frames_dropped_total Prometheus counter and register
it in metrics tests; document the metric in prometheus_grafana.md.
- Throttle drop WARN logs to power-of-ten milestones (1, 10, 100, …)
with cumulative drop count, instead of logging on every dropped frame.

and to expand on what you had.
- Harden shutdown: track closed state, ignore post-Close writes, close
the underlying dnstap writer eagerly, and skip run-loop errors after
shutdown.

0xERR0R:
Thanks for your contribution! :+1:

PR #2156, opened 2026-07-06
fix(lists): skip a lone "/" blocklist line instead of panicking

maximilize:
## What

A blocklist line that is just `/` is accepted by the host-list parser as a regex entry and then panics list import with `slice bounds out of range [1:0]`. One such line in a downloaded list crashes the refresh instead of being skipped.

## Why

`isRegex` (`lists/parsers/hosts.go`) treats a string as a `/regex/` entry when it merely starts and ends with `/`. A lone `/` satisfies both, so `validateHostsListEntry` accepts it (`regexp.Compile("/")` succeeds). Consumers then strip both delimiters: `cache/stringcache/string_caches.go:155` does `entry[1 : len(entry)-1]`, i.e. `"/"[1:0]`, which panics. Reproduced: `regexCacheFactory.addEntry("/")` gives `slice bounds out of range [1:0]`.

## Fix

Require both delimiters (`len >= 2`). A lone `/` then falls through to domain validation and is rejected as a normal resumable parse error, so the line is skipped and import continues. Only `/` changes; `//` and real `/regex/` entries are unaffected. `"/"` is added to the parser's invalid-lines test (RED before the fix, GREEN after).


maximilize:
Done in 36b7880. Added a `len(entry) >= 2` guard to `regexCacheFactory.addEntry`, so a lone `/` (which is both a `/` prefix and suffix) no longer reaches `entry[1:len-1]` and panics with `slice bounds out of range [1:0]` — it now falls through as a non-regex entry. A regex-cache regression test asserts `addEntry("/")` returns `false` and does not panic. The hosts-parser change is now belt-and-suspenders rather than the only barrier.

0xERR0R (commented):
Thanks for the clear writeup and for the RED/GREEN discipline on the test — that made this easy to verify. I reproduced the bug end-to-end and the fix is correct. A few notes below, one of which I'd like addressed before merge, plus a correction to the impact description.

## The bug is real, and it's reachable

Confirmed: `regexCacheFactory.addEntry("/")` panics with `slice bounds out of range [1:0]` at `cache/stringcache/string_caches.go:155`, and a lone `/` in a list reaches it. Pre-fix, `isRegex("/")` is true, `validateHostsListEntry` accepts it because `regexp.Compile("/")` succeeds, and the entry is handed to `AddEntry` at `lists/list_cache.go:212`, where the chained cache tries the regex factory first.

## The fix is correct

`len(host) >= 2` is exactly the right guard. `"/"` is the only string where `HasPrefix("/")` and `HasSuffix("/")` can overlap on a single byte, and `""` fails `HasPrefix`, so your claim that only `/` changes behaviour holds. I checked `//`, `/abc/`, `/ /` and `/invalid regex ??/` before and after: all unchanged. Post-fix `/` falls through to `validateDomainName`, produces a resumable `invalid domain name: /`, and the import continues with the remaining entries loaded.

## Please correct the impact description

The PR says the panic "crashes the refresh". It doesn't. `InitStrategy.Do` and `SourceLoading.periodically` both wrap refresh in `recoverToError` (`config/config.go`), so the panic is swallowed and **the entire group's list silently ends up empty**. With the default `maxErrorsPerSource: 5` and a list of `good.com`, `ads.example.com`, `/`, `more.example.com`:

- pre-fix: none of the three domains are blocked, and `NewListCache` returns no error
- post-fix: all three are blocked

So a single `/` line in an upstream list silently disables blocking for that whole group while blocky starts up reporting success. That's a fail-open, which for a DNS blocker is worse than a crash — worth stating accurately in the commit message. (The one path where the panic does escape is `POST /api/lists/refresh` → `RefreshLists` → `ListCache.Refresh`, which isn't wrapped by `recoverToError`.)

## Request: also guard the slice itself

I'd like to second [Copilot's inline suggestion](https://github.com/0xERR0R/blocky/pull/2156#discussion_r3554045670) here, because I reached the same conclusion independently: `regexCacheFactory.addEntry("/")` still panics on this branch. Your fix removes the only caller that can currently feed it a lone slash, so the reported bug is genuinely fixed — but the unguarded slice remains, and `resolver/query_logging_resolver.go` already had to hand-roll a guard against this exact input, with a comment saying so. Two call sites working around one unguarded slice is a good sign the guard belongs in the callee:

```go
func (r *regexCacheFactory) addEntry(entry string) bool {
	if len(entry) < 2 || !strings.HasPrefix(entry, "/") || !strings.HasSuffix(entry, "/") {
		return false
	}
```

The fallthrough is harmless: returning `false` lets the chained cache offer `/` to the wildcard cache (no `*`, declines) and then the string cache, which stores it as a literal no domain can equal.

Could you add that plus a case in `cache/stringcache/string_caches_test.go`? Which brings me to:

## The test doesn't exercise the panic

Worth knowing: with `hosts.go` reverted and your test line kept, the suite fails with

```
[FAILED] Expected an error to have occurred.  Got:
    <nil>: nil
In [It] at: lists/parsers/hosts_test.go:411
```

That's a plain assertion failure, not the panic. It's a good test of the parser contract and I want to keep it, but it would stay green if someone reintroduced the panic in the cache. A `stringcache` test on `addEntry("/")` is the one that actually pins the crash this PR is named after.

## Follow-up, not for this PR

`//` and `/ /` are accepted, get both delimiters stripped, and `regexp.Compile("")` succeeds — an empty regex that matches every domain. A list containing only `//` yields a match for any lookup, i.e. it blocks everything. `newIgnoreDomainsMatcher` already rejects these for exactly this reason; the blocklist path doesn't. Your PR correctly doesn't regress this, so I'll open a separate issue.

Also: the Codecov `-0.10%` is noise (the base is five commits behind and `Lines` is unchanged at 9816) — please ignore it.

Nice catch, and thanks again.

PR #2161, opened 2026-07-09
fix(stats): count only true blocks (BLOCKED) as blocked

### LINKED ISSUE #2151: Possible bug: `blocky stats` Top Blocked table cannot be trusted
## Description

I don't know why or how, but the new `blocky stats` CLI command shows entries that are resolved perfectly, mixed with entries that are indeed blocked. 

Image

When I look at the Blocky UI, I can correctly filter for blocked domains as expected:

Image

However, I'm also not sure if I can "trust" the counters:
- Blocky UI shows all stats parsed from csv log files, so I inherently "trust" it more as it's based on "evidence"
- `blocky stats` CLI seemingly shows a "window" that seem to be aligned with the start of the binary, and therefore the window cannot be modified, and the displayed values cannot be easily cross-referenced with the csv log files to check their correctness

Note: I'd prefer more if stats could be assembled from logs, if possible, not from the in-memory counters. While it has value to see the current runtime values, it cannot be used for eg. daily email reports from cron, that could send the `blocky stats` CLI output as an email at midnight. Setting the window would also be nice, or at least changing the stats source from "in-memory" to "logs".

The stats source/window size could be a separate issue, but at least please ensure that the counters shown in the table are correct, because the domains are apparently not.

## Expected results

- Top Blocked stats via the CLI show only the blocked domains
- Top Blocked stats counters are double-checked to show the correct values
- All other CLI stats tables are tested and reviewed to contain only their intended values

Failing to ensure these, the CLI stats cannot be trusted.

## Additional info

```
$ blocky version
blocky
Version: v0.33.0
Build time: 20260630-080230
Architecture: amd64
```

maximilize:
Fixes #2151.

Following the direction in #2151 (blocked should mean `BLOCKED` only), `categorize()` no longer folds query-type-filtered responses into "blocked":

- `FILTERED` (e.g. an `AAAA` lookup dropped by `filtering.queryTypes`) and `NOTFQDN` now map to a new `filtered` category instead of `blocked`.
- `isBlockedRType` — which feeds the **Top Blocked** table and the per-hour `blocked` series — now matches `BLOCKED` only. Domains that resolve fine on their `A` record but had a filtered `AAAA` query (the case reported in #2151) no longer show up as blocked.

User-facing:
- Stats API `summary` gains a `filtered` field (`api/*.gen.go` regenerated via `go generate`).
- `blocky stats` prints a new **Filtered** row.

Alternative, if a smaller API surface is preferred: drop the separate `filtered` field and let those responses fall into the unnamed remainder instead. That is a one-line change — say which you want.

Tested (`go test ./stats/... ./cmd/... ./api/...` green): new specs assert `Blocked` counts `BLOCKED` only, `Filtered` counts `FILTERED`+`NOTFQDN`, and Top Blocked excludes filtered domains. gofmt / vet / golangci-lint v2.12.2 clean.

### DIFF
--- a/stats/collector.go
+++ b/stats/collector.go
@@ -58,6 +58,7 @@ type Summary struct {
 	Cached        int
 	Forwarded     int
 	Blocked       int
+	Filtered      int
 	Local         int
 	Dropped       int
 	Errors        int
@@ -380,6 +381,7 @@ const (
 	categoryCached
 	categoryForwarded
 	categoryBlocked
+	categoryFiltered
 	categoryLocal
 )
 
@@ -392,9 +394,13 @@ func categorize(rtype string) responseCategory {
 		return categoryCached
 	case model.ResponseTypeRESOLVED.String(), model.ResponseTypeCONDITIONAL.String():
 		return categoryForwarded
-	case model.ResponseTypeBLOCKED.String(), model.ResponseTypeFILTERED.String(),
-		model.ResponseTypeNOTFQDN.String():
+	case model.ResponseTypeBLOCKED.String():
 		return categoryBlocked
+	case model.ResponseTypeFILTERED.String(), model.ResponseTypeNOTFQDN.String():
+		// Query-type filtering (e.g. AAAA via filtering.queryTypes) and NOTFQDN
+		// rejections are not blocklist hits, so they get their own bucket and no
+		// longer inflate "blocked" or the Top Blocked table (#2151).
+		return categoryFiltered
 	case model.ResponseTypeCUSTOMDNS.String(), model.ResponseTypeHOSTSFILE.String(),
 		model.ResponseTypeSPECIAL.String(), model.ResponseTypeSYNTHESIZED.String():
 		return categoryLocal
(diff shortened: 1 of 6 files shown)


stats/collector.go
```diff
@@ -392,9 +394,13 @@ func categorize(rtype string) responseCategory {
 		return categoryCached
 	case model.ResponseTypeRESOLVED.String(), model.ResponseTypeCONDITIONAL.String():
 		return categoryForwarded
-	case model.ResponseTypeBLOCKED.String(), model.ResponseTypeFILTERED.String(),
-		model.ResponseTypeNOTFQDN.String():
+	case model.ResponseTypeBLOCKED.String():
 		return categoryBlocked
+	case model.ResponseTypeFILTERED.String(), model.ResponseTypeNOTFQDN.String():
```
0xERR0R, line 399: here is a small problem: filtered is overloaded and can be "filtered out by type" or also rebind protection. I think, I will introduce new dedicated response type later, we can leave it as is for now

0xERR0R (approved):
Thanks for your contribution!

PR #2190, opened 2026-07-26
fix(bootstrap): fall back to other resolved addresses when a dial fails

### LINKED ISSUE #2188: List downloads waste retry attempts dialing IPv6 on IPv4-only hosts (bootstrapDns dialer has no address-family fallback)
**Blocky version:** v0.33.0 (container, Kubernetes, arm64)

**Setup:** IPv4-only network (the host has no global IPv6 route, only link-local). Relevant config:

```yaml
bootstrapDns:
  - "1.1.1.1"
  - "1.0.0.1"
blocking:
  loading:
    strategy: failOnError
    downloads:
      timeout: 15s
      attempts: 5
      cooldown: 2s
```

**Observed:** during blocklist download, an attempt is consumed by dialing a resolved AAAA even though the host has no IPv6 connectivity:

```
WARN list_cache: Can't download file: HTTP request to 'https://raw.githubusercontent.com/hagezi/dns-blocklists/main/adblock/tif.txt' failed: Get "https://raw.githubusercontent.com/hagezi/dns-blocklists/main/adblock/tif.txt": failed to dial '[2606:50c0:8000::154]:443' (resolved from 'raw.githubusercontent.com'): dial tcp [2606:50c0:8000::154]:443: connect: network is unreachable attempt=1/5 link=https://raw.githubusercontent.com/hagezi/dns-blocklists/main/adblock/tif.txt
```

**Analysis:** when `bootstrapDns` is configured, the list downloader's HTTP transport uses the bootstrap dialer (`Bootstrap.NewHTTPTransport` → `dialContext`), which resolves the hostname via the bootstrap resolvers and dials one resolved address directly. If that address is an AAAA on a v4-only host, the dial fails with `ENETUNREACH` and the entire download attempt is consumed — there is no fallback to the remaining resolved addresses (the other family) within the attempt, unlike Go's default dual-stack dialing.

With `failOnError` + a bounded `attempts` budget, these wasted attempts matter: combined with a slow/large list they can exhaust the budget and fail startup on a host where an immediate v4 retry would have succeeded.

**Expected:** on dial failure, try the remaining resolved addresses (preferring the other address family) before consuming the attempt — or sort resolved addresses by local route availability (RFC 6724-style) before dialing.

**Workaround:** `connectIPVersion: v4` works (it restricts bootstrap resolution to A records), but it requires knowing/declaring the deployment's address family in config where a fallback in the dialer would handle it automatically.

maximilize:
Fixes #2188.

### Problem

When `bootstrapDns` is configured, the list downloader dials through `Bootstrap.dialContext`, which resolves the host itself and then dials a single randomly-picked resolved address (`resolver/bootstrap.go:223`). On failure it returns immediately (`:230-233`) without trying the other resolved addresses. On an IPv4-only host a picked AAAA fails with `ENETUNREACH` and the whole download attempt is consumed — with `loading.strategy: failOnError` and a bounded `attempts` budget this can fail startup where an immediate v4 retry would have succeeded. The standard dialer does this dual-stack fallback for us, but we bypass it by resolving the host ourselves and dialing one address.

### Fix

Shuffle the resolved addresses (keeping the load-spreading the previous random pick provided) and try each in turn, returning the first successful connection and only failing when every address fails.

### Test

`resolver/bootstrap_test.go`: with a dual-stack resolve and a dialer that fails the first dial (mimicking an unreachable address family) and succeeds on the next, the transport now falls back and connects. Against the previous code the single dial fails. Deterministic — no dependence on shuffle order or family.

### Validation (isolated `golang:1.26` container)

- `go test -race ./resolver/` — green (697 specs)
- `gofmt -l`, `go vet ./resolver/`, `golangci-lint v2.12.2 run ./resolver/...` — clean


0xERR0R (approved):
👍

PR #2191, opened 2026-07-27
fix(resolver): match custom DNS PTR queries case insensitively

maximilize:
`reverseAddresses` is keyed by the output of `dns.ReverseAddr` (`resolver/custom_dns_resolver.go:63-68`), which is always lower case. `handleReverseDNS` looked the question name up verbatim (`resolver/custom_dns_resolver.go:109`), so a PTR query that uses a different case missed the mapping and fell through to the next resolver — no custom DNS answer.

That happens with upper case hex nibbles in `ip6.arpa`, an upper case `IN-ADDR.ARPA` suffix, or a client that applies 0x20 encoding.

The forward path already normalizes both sides through `util.ExtractDomainOnly` (keys at `custom_dns_resolver.go:48,57`, lookup via `util.ExtractDomain`), and its case insensitivity is covered by an existing spec (`ABC.CUSTOM.DOMAIN.`). This brings the reverse path in line, using the same `strings.ToLower` idiom as `sudn_resolver.go:133`.

The response still echoes the question name as sent, since `util.CreateHeader` builds the header from the original question.

**Test:** a new spec mirrors the existing "Reverse DNS request is received" case with upper case names, for both `IN-ADDR.ARPA` and `IP6.ARPA`. It fails before the change (answer is `nil`) and passes after.

Verified in a `golang:1.26` container: `go test -race ./resolver/` green (697 specs), `gofmt -s` and `go vet` clean, `golangci-lint v2.12.2` reports 0 issues.


0xERR0R (approved):
Thank you for your contribution

PR #2197, opened 2026-07-28
fix(resolver): count down cached authority and additional TTLs

maximilize:
`setTTLInCachedResponse` rewrites the TTLs of a cached message on every cache hit, but it only walks `resp.Answer`. Records in the authority and additional sections keep the TTL the upstream sent and are served with that same value for the entire lifetime of the cache entry. For NXDOMAIN and NODATA the answer section is empty, so the loop never runs and nothing counts down at all.

Observed on `main` (5e7e844) with a mock upstream, querying twice through the caching resolver:

```
NXDOMAIN with SOA in the authority section
  1st (upstream):   AUTHORITY  example.com.  ttl=3600
  2nd (cached):     AUTHORITY  example.com.  ttl=3600     <-- unchanged

NOERROR with answer + NS + glue
  1st (upstream):   ANSWER 3600  AUTHORITY 3600  ADDITIONAL 3600
  2nd (cached):     ANSWER 3597  AUTHORITY 3600  ADDITIONAL 3600
```

Why it matters beyond cosmetics: a stub resolver derives its negative caching time from the SOA record (RFC 2308), so it re-arms that timer to the full upstream TTL every time it refreshes. A domain that stays NXDOMAIN for blocky's `cacheTimeNegative` window can stay NXDOMAIN downstream considerably longer, counted from the client's last query rather than from when blocky first cached it.

Scope note: downstream negative caching is `min(SOA.MINIMUM, SOA.TTL)`, and only the TTL is aged here — the MINIMUM rdata field is not a lifetime and cannot sensibly be counted down. The improvement therefore lands in the case where MINIMUM is the larger of the two.

### Change

Age every section by the time the entry has already spent in the cache, instead of rebasing only the answer section:

* For the answer section the arithmetic is unchanged — `rr.Ttl - minTTL + remaining` is exactly `rr.Ttl - elapsed`, so existing behaviour and the existing specs stay as they are.
* The baseline mirrors `putInCache`, which picks it by **rcode**: only a successful, non-empty answer counts down from the smallest answer TTL. Everything else — NODATA and any NXDOMAIN, including one that carries a CNAME chain — was stored with `cacheTimeNegative` and counts down from that.
* OPT records are skipped: their TTL field carries flags and the extended rcode, not a lifetime. `packForCache` already strips OPT before caching, so this is belt-and-braces for future callers.

### Tests

New specs in `caching_resolver_test.go` covering the cases above — the SOA TTL of a cached NXDOMAIN, an NXDOMAIN carrying a CNAME chain, and the authority/additional TTLs alongside the answer. Each fails on `main` (`Expected <uint32>: 3600 to be < 3600`) and passes with the change.

Verified in a container: `go test -race ./resolver/` → 702 of 702 specs pass, `gofmt -s`/`go vet` clean, `golangci-lint v2.12.2` → 0 issues.

### Known limitation, left for a follow-up

`adjustTTLs` clamps only the answer section by `minTime`/`maxTime`. With e.g. `maxTime: 5m`, an upstream NS record at 3600 now ages 3600 → 3300 and then resets — still well above the entry's own lifetime. Same class of problem one level down, not addressed here.


0xERR0R (changes requested):
Thanks for the thorough writeup — the diagnosis is correct and the RFC 2308 angle is the part that actually matters here, not the cosmetics. I reproduced it: your two specs fail against `main`'s implementation and pass on the branch, and `go test -race ./resolver/` is green (700/700), as is CI.

One change I'd like before merging.

### Please change: key the negative baseline off the rcode, not the empty answer

`putInCache` decides "store with `CacheTimeNegative`" by **rcode** — any NXDOMAIN goes down that path, whether or not it carries an answer section. `setTTLInCachedResponse` instead keys off `len(resp.Answer) == 0`, so an NXDOMAIN that carries a CNAME chain (a CNAME pointing at a name that doesn't exist — very common) takes the positive branch: `baseTTL` comes from the CNAME, `remaining` exceeds it, `elapsed` clamps to 0, and nothing ages.

On this branch, with defaults, `example.com. 60 IN CNAME gone.example.net.` plus an SOA at 3600:

```
1st: rcode=3  answer(CNAME)=60  authority(SOA)=3600
2nd (CACHED): answer(CNAME)=60  authority(SOA)=3600   <-- frozen for the full 30 min
```

That's the exact case the PR is aiming at. Mirroring `putInCache` closes it:

```go
	// Mirrors putInCache: only a successful, non-empty answer is stored with the smallest
	// answer TTL (see adjustTTLs). Everything else -- NXDOMAIN (even one carrying a CNAME
	// chain) and NODATA -- is stored with the negative cache time. That baseline is what
	// the remaining lifetime has been counting down from.
	baseTTL := r.cfg.CacheTimeNegative.SecondsU32()

	if resp.Rcode == dns.RcodeSuccess && len(resp.Answer) > 0 {
		baseTTL = uint32(math.MaxInt32)
		for _, rr := range resp.Answer {
			baseTTL = min(baseTTL, rr.Header().Ttl)
		}
	}
```

I ran this against the full resolver suite plus a spec for the CNAME case — all green. Could you add that spec alongside your other two?

### Not blocking, just noting

- `adjustTTLs` still clamps only the answer section by `minTime`/`maxTime`, so with e.g. `maxTime: 5m` an upstream NS at 3600 ages 3600→3300 and resets — still well above the entry's own lifetime. Same class of problem one level down; fine to leave for a follow-up.
- Downstream negative caching is `min(SOA.MINIMUM, SOA.TTL)`, and this ages the TTL only (the MINIMUM rdata field can't sensibly be aged), so the improvement lands when MINIMUM is the larger of the two. Worth a sentence in the description so expectations match.

The OPT guard is unreachable today — `packForCache` strips EDNS0 and `getFromCache` unpacks those bytes — but I'm happy to keep it as defence for future callers, and the comment says as much.

Nice work overall: the comments explain the *why*, and the specs are real regression tests rather than change detectors.

maximilize:
Done — your baseline adopted verbatim in `1cdee92`.

I checked the CNAME case against the code before changing anything: `putInCache` (`caching_resolver.go:392-400`) routes by rcode, so `RcodeNameError` takes the `CacheTimeNegative` put regardless of what the answer section carries, while `setTTLInCachedResponse` keyed off `len(resp.Answer)`. Your reading is exact.

New spec `when a cached NXDOMAIN response carries a CNAME chain` (`example.com. 60 IN CNAME gone.example.net.` plus SOA 3600), asserting that both the SOA and the CNAME age. Against the old baseline it is RED and it is the only failure of the 702 specs; green with the rcode baseline.

Verified in a container (golang:1.26, offline): `go test -race ./resolver/` 702/702, `gofmt -s` and `go vet` clean, `golangci-lint v2.12.2` 0 issues.

The description now carries the `min(SOA.MINIMUM, SOA.TTL)` sentence, and the `minTime`/`maxTime` clamp in `adjustTTLs` is written up there as a known limitation for a follow-up rather than folded into this PR.

0xERR0R (approved):
Thank you, LGTM 👍

Issue #1741, opened 2025-03-22
DNS Cache ignores DNSSEC flag


LekkerKofi:
When a query with the DNSSEC flag is sent, Blocky will reply without the RRSIG record if the same query without the DNSSEC flag has been made and cached. The same also applies the other way around. A client that doesn't request DNSSEC information can prevent other clients that do verify DNSSEC from successfully resolving domains.

<details>
<summary>Query without DNSSEC first</summary>

```
$ dig @blocky example.com +short
23.215.0.136
23.192.228.80
96.7.128.175
23.215.0.138
23.192.228.84
96.7.128.198

$ dig @blocky example.com +short +dnssec
23.215.0.136
23.192.228.80
96.7.128.175
23.215.0.138
23.192.228.84
96.7.128.198
```

</details>

<details>
<summary>Query with DNSSEC first</summary>

```
$ dig @blocky example.com +short +dnssec
23.192.228.84
23.215.0.136
96.7.128.175
96.7.128.198
23.215.0.138
23.192.228.80
A 13 2 300 20250406061732 20250316122333 15496 example.com. yDJ7Nv+MCtCLvobPIEUgV0e8gQ3/j0xL5hXC+XIXv63g+CKptUuZ1rUU RZPclonSymV/nQtPzTyoSigBQwgU9w==

$ dig @blocky example.com +short
23.192.228.84
23.215.0.136
96.7.128.175
96.7.128.198
23.215.0.138
23.192.228.80
A 13 2 300 20250406061732 20250316122333 15496 example.com. yDJ7Nv+MCtCLvobPIEUgV0e8gQ3/j0xL5hXC+XIXv63g+CKptUuZ1rUU RZPclonSymV/nQtPzTyoSigBQwgU9w==
```

</details>

PR #2203, opened 2026-08-02
fix(dnssec): don't return DNSSEC records to clients with the DO bit clear

### LINKED ISSUE #2202: DNSSEC validation returns RRSIG records to clients that didn't set the DO bit
With `dnssec.validate: true`, blocky includes DNSSEC records (RRSIG) in answers to clients that did not set the EDNS DO bit. RFC 4035 §3.2.1 ("The DO Bit") requires the opposite: for a query with no OPT pseudo-RR, or with DO clear, the name server must treat RRSIG/NSEC/DNSKEY as ordinary RRsets and must not add them to the response.

Repro (blocky v0.33.0, `dnssec.validate: true`, any upstream):

```
$ dig @<blocky> isc.org A +nodnssec        # EDNS present, DO=0 — a normal stub query

;; ANSWER SECTION:
isc.org.  300  IN  A      151.101.2.217
isc.org.  300  IN  A      151.101.66.217
isc.org.  300  IN  RRSIG  A 13 2 300 20260806185109 ... isc.org. ...   # returned although DO=0
```

The RRSIG comes back for every signed name, with or without an OPT in the query (`+noedns` behaves identically). A conformant validating resolver (1.1.1.1 / 8.8.8.8 / 9.9.9.9) returns only the A records for the same query.

Impact: strict stub resolvers reject the unsolicited DNSSEC RRs. macOS `mDNSResponder` (used by every app via getaddrinfo, including Safari and Chrome) stalls ~30s per signed domain before returning — so enabling `dnssec.validate` silently breaks resolution of signed domains for macOS clients. Pointing the same client at a conformant resolver makes it instant; `dig` tolerates the extra records, which hides the problem during testing.

It looks like the validated RRset — fetched from the upstream with DO=1, RRSIGs included — is cached and served verbatim regardless of the downstream client's DO bit. The response path should strip DNSSEC RRs (RRSIG/NSEC/NSEC3/DNSKEY) when the request's DO bit is clear.

kevinpark1217:
Fixes #2202.

### Problem

A validating resolver must query upstream with the DO bit set, so `DNSSECResolver` sets it on every request regardless of what the client asked for (`resolver/dnssec_resolver.go:71-89`). Nothing puts it back: `resolver/dnssec/validator.go` only classifies the response, and `server.resolve` normalized the OPT record and the message size but not the DNSSEC records. A client that left DO clear therefore received the RRSIG (and NSEC/NSEC3/DNSKEY/DS) records blocky fetched for its own validation, which RFC 4035 §3.2.1 forbids: the name server side "MUST strip any authenticating DNSSEC RRs from the response but MUST NOT strip any DNSSEC RR types that the initiating query explicitly requested".

Strict stub resolvers reject the unsolicited records; per the issue, macOS `mDNSResponder` stalls ~30s per signed domain, so `dnssec.validate: true` breaks resolution of signed domains for macOS clients.

Because the client's DO bit is only known before the chain mutates `request.Req`, this is fixed where the other client-facing normalization already happens. The first commit is a behavior-preserving refactor collecting that normalization into a `clientQuery` value captured up front; the existing server specs pass unchanged.

### Fix

`clientQuery.normalizeResponse` now also normalizes against the client's DO bit:

- **Strip the DNSSEC records** when DO is clear, keeping any type the query asked for explicitly, so a DNSKEY or DS query is still answered — just without the signatures over it (RFC 4035 §3.2.1). RFC 3225 §3 extends the exception to `QTYPE=ANY` ("Security records that match an explicit SIG, KEY, NXT, or ANY query [...] are included whether or not the DO bit was set"), so an ANY query keeps everything.
- **Copy the query's DO bit into the response's OPT record** (RFC 3225 §3).
- **Clear AD unless the client set DO or AD** (RFC 6840 §5.8: "validating resolvers SHOULD only set the AD bit when a response both meets the conditions listed in Section 3.2.3 of [RFC4035], and the request contained either a set DO bit or a set AD bit").

Stripping runs before `Truncate`, so a signed answer that only exceeded the client's buffer because of the signatures is no longer truncated.

**Why the server and not the DNSSEC resolver:** the cache key is qtype and name only (`util/common.go:223`) and `packForCache` keeps the signatures, so an entry populated by a DO client is served to every other client of that name — including with `dnssec.validate: false`, where `DNSSECResolver.Resolve` is a no-op. Stripping has to be per client at the point the response leaves, which is also where the OPT-record and truncation normalization already live.

The cache is unaffected by the stripping: it stores packed bytes and unpacks a fresh message per hit (`resolver/caching_resolver.go:266-282`, `:343-351`), so mutating one client's response cannot touch the shared entry.

### Test

- `util/dnssec_test.go`: `StripDNSSECRecords` over signed answers, NSEC and NSEC3 denial proofs, unrequested key material, the explicitly-requested exception per type, `QTYPE=ANY`, multi-type requests, all three sections, and that the OPT record survives.
- `server/server_test.go`: through `s.resolve` with a chain that sets DO and returns signatures — stripping for an EDNS0 client with DO clear and for a client without EDNS0, the DO bit mirrored, AD cleared, an explicit DNSKEY query keeping its answer, ANY keeping everything, a chain that returns signatures without validating (the `validate: false` cache case), a 5×(A+RRSIG) answer no longer truncated to 512 bytes, DO=1 clients still getting records + AD, and AD-only clients keeping AD while still being stripped.
- `e2e/dnssec_test.go`: against a real container with `dnssec.validate: true` and a valid chain of trust, a DO-clear query gets the A record without RRSIG and without AD; and after the upstream is terminated, a DO=1 query is still served the signatures from the cache.

The e2e spec fails against an image built from `main` (`RRSIG must not be returned to a client that didn't set the DO bit`) and passes with this change. Each normalization step and the ANY carve-out were individually reverted to confirm the specs catch them.

### Validation (isolated `golang:1.26` container)

- `go tool ginkgo --label-filter="!e2e" -r` — green, 21 suites
- `go tool ginkgo --label-filter="dnssec" e2e` — green, 12 specs
- `gofumpt -l -extra`, `goimports -l`, `go vet ./...`, `golangci-lint v2.12.2 run` — clean

### Notes for review

- **The three normalizations are unconditional**, not gated on `dnssec.validate` — an upstream can return DNSSEC records either way, and the cache case above needs them applied with validation off. The docs bullets are worded accordingly. The visible consequence with `validate: false` is that an upstream-supplied `AD=1` is no longer passed to clients that asked for neither DO nor AD, and the `/api/query` endpoint (which builds a DO-clear message) no longer shows RRSIGs.
- **DO mirroring only applies to responses that carry an OPT record.** Cache hits are served from bytes packed without one (`packForCache`), so an EDNS0 client gets no OPT there at all — pre-existing on `main` and left alone here, but it means RFC 6891 §6.1.1 is not currently satisfied on that path.
- **The query log and dnstap record the pre-normalization response**, since `QueryLoggingResolver` sits above the server. Stripped records and the cleared AD bit will still appear there. This divergence already existed for OPT removal and truncation; this change widens it.
- **Not addressed:** the reverse cache-key problem, where a DO=1 client is served an unsigned entry populated by a DO=0 client (#1741, closed as stale). That needs a cache-key change and cannot happen while `dnssec.validate` is enabled, because the chain then always fetches with DO set.


0xERR0R (approved):
thanks for this fix

PR #2221, opened 2026-08-14
fix(cmd): healthcheck follows ports.dns from the config

### LINKED ISSUE #2218: Docker healthcheck ignores configured ports.dns
### Description

  The healthcheck embedded in the Docker image always probes
  `127.0.0.1:53`, even when Blocky is configured to listen on a different
  DNS port through `ports.dns`.

  This causes a correctly running container to be reported as unhealthy.

  ### Configuration

  ```yaml
  ports:
    dns: 3000
    http: 3001
  ```

  The container publishes host port 53 to container port 3000:

  ```yaml
  ports:
    - "192.168.178.39:53:3000/tcp"
    - "192.168.178.39:53:3000/udp"
  ```

  ### Actual result

  Blocky starts successfully:

  ```text
  DNS = [:3000]
  UDP server is up and running on address :3000
  TCP server is up and running on address :3000
  ```

  However, the healthcheck embedded in the image continues to probe port 53:

  ```text
  Error: healthcheck failed for 127.0.0.1:53:
  dial tcp 127.0.0.1:53: connect: connection refused
  ```

  As a result, Docker reports the running container as unhealthy.

  ### Workaround

  Overriding the image healthcheck with the configured internal DNS port works:

  ```yaml
  healthcheck:
    test: ["CMD", "/app/blocky", "healthcheck", "-p", "3000"]
  ```

  ### Expected result

  When neither `--port` nor `--bindip` is specified explicitly, the
  healthcheck should derive its target from `ports.dns` in the loaded
  configuration.

  Explicit CLI flags should continue to take precedence.

  ### Proposed behavior

  1. Explicit `--port` and `--bindip` flags take precedence.
  2. Otherwise, derive the healthcheck target from the first configured
     `ports.dns` listener.
  3. For wildcard listeners such as `3000` or `:3000`, probe
     `127.0.0.1:3000`.
  4. For listeners bound to a concrete IP address, probe that address and port.

  ### Environment

  - Blocky: v0.34.0
  - Image: `ghcr.io/0xerr0r/blocky:v0.34.0`
  - Docker Compose: v5.4.0

  ### Related issues

  - #1666
  - #1125
  - #1304

  I would be happy to submit a pull request for this. Please let me know
  whether the proposed behavior fits the project's intended design.

maximilize:
Fixes #2218.

The image's own `HEALTHCHECK` runs `/app/blocky healthcheck` without flags (`Dockerfile:82`), but the command defaults to port 53 and never loads the configuration. A container listening on a different DNS port is therefore reported unhealthy while it is serving correctly.

**`healthcheck` was the only subcommand that does not load the config.** `lists`, `cache`, `stats`, `blocking`, `serve` and `query` all wire `PersistentPreRunE: initConfigPreRun`, and `validate` calls `initConfig()` directly. `initConfig()` also already derives `apiHost`/`apiPort` from `ports.http` — this change does the same for `ports.dns`, so the existing mechanism now covers the DNS side too. The image already exports `BLOCKY_CONFIG_FILE=/app/config.yml` (`Dockerfile:78`), which `initConfig()` reads, so the healthcheck finds the config without any change to the Dockerfile.

Behaviour kept intact:

- An explicitly given `--port`/`--bindip` still wins over the config (checked via `Flags().Changed`), so the documented workaround `healthcheck -p 3000` keeps working.
- A missing or unreadable config does **not** fail the healthcheck — the error is ignored and the previous flag defaults (`127.0.0.1:53`) apply, exactly as before.
- Only the new `ports.dns` path uses `net.SplitHostPort`; the existing `ports.http` parsing is untouched.

**Tests** (`cmd/issue_2218_test.go`, ginkgo): three specs — port taken from `ports.dns` when no flag is given (this one fails without the fix), explicit `--port` beating the config, and no-config-file still working. The latter two are regression guards and pass both before and after.

**Verification** (isolated container, `golang:1.26`):

```
go test ./cmd/                                   # RED before the fix: 1 of 57 specs fails
                                                 # GREEN after: ok
go test -race $(go list ./... | grep -v /e2e)    # 20 packages ok
gofmt -l ./cmd ./config                          # clean
go vet ./cmd/... ./config/...                    # clean
golangci-lint v2.12.2 run                        # 0 issues
```

### DIFF
--- a/cmd/healthcheck.go
+++ b/cmd/healthcheck.go
@@ -18,7 +18,16 @@ func NewHealthcheckCommand() *cobra.Command {
 	c := &cobra.Command{
 		Use:   "healthcheck",
 		Short: "performs healthcheck",
-		RunE:  healthcheck,
+		// Load the configuration like every other subcommand, so the defaults below can
+		// follow `ports.dns`. A missing or unreadable config must not break the
+		// healthcheck: it then keeps working with the flag defaults, exactly as before.
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			_ = args
+			_ = initConfig()
+
+			return nil
+		},
+		RunE: healthcheck,
 	}
 
 	c.Flags().Uint16P("port", "p", defaultDNSPort, "blocky port")
@@ -31,6 +40,15 @@ func healthcheck(cmd *cobra.Command, args []string) error {
 	_ = args
 	port, _ := cmd.Flags().GetUint16("port")
 	bindIP, _ := cmd.Flags().GetString("bindip")
+
+	// An explicitly given flag always wins; otherwise follow the loaded configuration.
+	if !cmd.Flags().Changed("port") {
+		port = dnsPort
+	}
+
+	if !cmd.Flags().Changed("bindip") {
+		bindIP = dnsHost
+	}
 
 	c := new(dns.Client)
 	c.Net = "tcp"
--- a/cmd/issue_2218_test.go
+++ b/cmd/issue_2218_test.go
@@ -1,0 +1,98 @@
+package cmd
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/0xERR0R/blocky/helpertest"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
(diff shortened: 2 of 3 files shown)


0xERR0R (approved):
lgtm 👍

PR #2222, opened 2026-08-16
feat(resolver): add blocky_client_response_total metric

### LINKED ISSUE #2198: blocky_query_total can't be split by blocklist reason — no way to alert on "real" vs blocked query rate
\`blocky_query_total{client,type}\` counts every incoming query, blocked or not.
\`blocky_response_total{reason,response_code,response_type}\` has the blocklist/reason detail, but is a
separate counter with no shared dimension to join on cheaply in a rate() alert (they only correlate
via time window, not a label).

Concretely: a client repeatedly querying an already-denylisted domain (e.g. mDNS/Bonjour meta-queries
like lb._dns-sd._udp, blocked via a custom denylist) inflates blocky_query_total identically to genuine
upstream traffic. There's no way to write "alert when a client's *unblocked* query rate is high" without
also catching noisy-but-harmless blocked traffic.

Proposal: add the same reason/response_code info blocky_response_total already computes to
blocky_query_total's labels (or split by a low-cardinality "blocked" bool + reason=blocklist-group).
Both counters are populated in the same Resolve() call in resolver/metrics_resolver.go, after the
response is known — so no extra resolution work, just recording reason alongside client+type on the
existing counter.

Cardinality: reason is already the low-cardinality group name (per code comment: "ReasonLabel... low-
cardinality... group names only"), so client × type × reason stays bounded by config, not by traffic
volume.

Would a PR in this direction be welcome, or is there a preferred way to correlate these two counters
that I'm missing?

jvandenbroek:
Closes #2198.

Adds `blocky_client_response_total{client, response_type}` — a per-client counter of query outcomes (resolved/blocked/cached/etc.), as discussed in #2198. `response_type` is the existing bounded 12-value enum, so this avoids the cardinality problems a `reason` label on `blocky_query_total` would have introduced.

Also documents (in `docs/prometheus_grafana.md`) that the `client` label itself is unbounded — derived from rDNS, not config — which matters on networks with high device turnover (public/guest wifi), as flagged in the issue discussion.

## Test plan
- [x] `go build ./...`
- [x] `go test ./resolver/...` — 702/702 passing, including two new specs covering the new counter on both the success and error paths


0xERR0R (approved):
👍

0xERR0R (approved):
👍

PR #2223, opened 2026-08-17
fix(resolver): restore fallbackUpstream for rewritten queries

maximilize:
`fallbackUpstream` has had no effect since #1884.

That PR removed `RewriterResolver` and moved rewriting into `CustomDNSResolver` and `ConditionalUpstreamResolver`, but the fallback the rewriter performed did not come along — its implementation and its tests were deleted with the file. The config field, its `LogConfig` line and the documentation stayed, so a config with `fallbackUpstream: true` still loads, still logs `fallbackUpstream = true`, and does nothing.

The removed code was:

```go
fallbackCondition := err != nil || (response != NoResponse && response.Res.Answer == nil)
if r.cfg.FallbackUpstream && fallbackCondition {
	// Inner resolver had no answer, configuration requests fallback, continue with the normal chain
	return r.next.Resolve(ctx, request)
}
```

It ran after the request had been reverted to the original name, which is what the documentation promises: *"If not found and if `fallbackUpstream` was set to `true`, the original query "blog.example.com" will be sent upstream."*

This restores it as `shouldFallbackUpstream` in `rewrite_helper.go`, called from both resolvers that carry a `RewriterConfig`. It only applies when rewrites are configured — as before, since `NewRewriterResolver` returned the inner resolver unwrapped for an empty `rewrite` map.

Two details where the new structure needs more care than the old one, because these resolvers now continue the chain themselves instead of being wrapped:

- `ConditionalUpstreamResolver` only falls back when the mapping matched (`resolved`), so a query that never hit the conditional mapping is not resolved twice. The error case is included, as before.
- `CustomDNSResolver` falls back only for a response it produced itself (`ResponseTypeCUSTOMDNS` with an empty answer, i.e. the `filterUnmappedTypes` case). A response it merely passed through from the next resolver must not be sent there again, so unlike the old code this branch does not include errors.

Tests: one spec per resolver, both failing on master (`should ask the next resolver with the original name`) and passing here, plus a spec pinning that nothing changes when `fallbackUpstream` is off. Verified in a container: `go test ./resolver/` 704 → 706 specs, `-race` green, full `go test ./...` green over 20 packages, `gofmt` clean, `golangci-lint v2.12.2` (the version pinned in the Makefile) reports 0 issues.

While reading this I noticed something adjacent that this PR does *not* change: when a rewrite matched but no mapping did, both resolvers now hand the **rewritten** name to the next resolver, whereas the old rewriter reverted first. That is a separate behaviour question from the same refactor, so it seemed better kept out of a fix that only restores the documented flag.


0xERR0R (approved):
👍

PR #2239, opened 2026-08-25
feat(resolver): add refused block type

### LINKED ISSUE #2187: Respond with REFUSED for blocked domains
For parity with AdGuard DNS:

Image

ChrisJr404:
Closes #2187.

Adds `refused` as a blockType option, so blocked queries are answered with the DNS REFUSED rcode for parity with AdGuard DNS. It sits right next to `nxDomain`: the handler sets the rcode and returns no records, so blockTTL has no effect in this mode.

Added a test for it and a row in the config docs.

### DIFF
--- a/config/blocking.go
+++ b/config/blocking.go
@@ -23,7 +23,8 @@ type Blocking struct {
 	ListSchedules map[string][]string `yaml:"listSchedules"`
 	// Maps client identifiers (name, IP, CIDR) to the list groups that apply to them.
 	ClientGroupsBlock map[string][]string `yaml:"clientGroupsBlock"`
-	// Response for blocked A/AAAA queries: zeroIP (0.0.0.0/::), nxDomain, or custom IPs; other types get NXDOMAIN.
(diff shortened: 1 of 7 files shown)


0xERR0R:
Thanks for the PR! I pushed a follow-up commit directly to the branch:

- `refused` was missing from the `blockType` description in `config/blocking.go`, which generates `docs/config.schema.json` (served by blocky and used for editor autocomplete), and from the reference config in `docs/config.yml`. The old wording "other types get NXDOMAIN" was also wrong for `refused`, which ignores the query type.
- Scoped two now-inaccurate statements in `docs/configuration.md`: the "only for A and AAAA queries" section intro and "blockTTL applies to all blocking modes". Also documented that stub resolvers commonly treat REFUSED as a server failure and fail over to another configured DNS server, which bypasses blocking — worth knowing before enabling it.
- Added coverage for query types other than A/AAAA (the behaviour that distinguishes `refused` from the other handlers) plus an e2e case in the "Block types" suite.

The handler itself is unchanged — it was correct as written.

PR #2240, opened 2026-08-26
fix(server): always answer an EDNS0 query with an OPT record

jackboykin:
Responses served from the cache carry no OPT record, even when the query had one. `packForCache` strips the OPT before storing (#1150), and `normalizeResponse` only ever removes an OPT, never adds one. So the first answer for a name has an OPT and every later answer doesn't.

RFC 6891 §6.1.1 requires an OPT record in the response to any EDNS0 query. systemd-resolved treats its absence as a server without EDNS0 support, downgrades to plain UDP, and re-probes about every 45 minutes. Lookups that land inside a probe window return `SERVFAIL` to applications. The journal on such a client shows the loop:

```
Grace period over, resuming full feature set (UDP+EDNS0) for DNS server 192.168.1.161
Using degraded feature set UDP instead of UDP+EDNS0 for DNS server 192.168.1.161
```

In this commit when the client sends an OPT record and the response has none, `normalizeResponse` adds one with the client's DO bit. Responses to clients without EDNS0 unchanged.

### DIFF
--- a/server/client_query.go
+++ b/server/client_query.go
@@ -76,17 +76,17 @@ func (q clientQuery) normalizeResponse(res *dns.Msg) {
 		util.StripDNSSECRecords(res, q.qType)
 	}
 
-	if opt := res.IsEdns0(); opt != nil {
-		// RFC 3225 §3: the DO bit of the query is copied into the response. The OPT record here is
-		// the upstream's, which answered a query whose DO bit the chain may have set. Responses
-		// blocky assembles itself carry no OPT and none is added here, so there is nothing to
-		// mirror onto - notably cache hits, which are served from bytes packed without the OPT.
-		opt.SetDo(q.wantsDNSSEC)
-	}
-
 	if !q.hadEdns0 {
 		// don't return an OPT record to a client that didn't use EDNS0 (RFC 6891 section 7)
 		util.RemoveEdns0Record(res)
+	} else if opt := res.IsEdns0(); opt != nil {
+		// RFC 3225 §3: the DO bit of the query is copied into the response
+		opt.SetDo(q.wantsDNSSEC)
+	} else {
+		// RFC 6891 §6.1.1: a response to an EDNS0 query must carry an OPT record. Cache hits
+		// are served from bytes packed without one; resolvers such as systemd-resolved read
+		// its absence as a server without EDNS0 support and downgrade.
+		res.SetEdns0(dns.DefaultMsgSize, q.wantsDNSSEC)
 	}
 
 	// truncate if necessary; Truncate also disables compression when the message already fits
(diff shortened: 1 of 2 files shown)


0xERR0R:
Thanks for tracking this down — the diagnosis is spot on. The `packForCache` /
`normalizeResponse` asymmetry is real, and the systemd-resolved downgrade loop it
causes is a genuine user-visible bug in the default configuration.

I've reviewed this closely and I'm happy to merge it. I pushed one commit to your
branch (hope that's OK — shout if you'd rather revert it and apply it yourself).

### What I pushed

There's one configuration where the fix doesn't fire: with `ede.enable`, cache hits
and blocked answers *already* carry an OPT record, so they take the mirroring branch
and never reach the new `else`. That OPT comes from `util.SetEdns0Option`
(`util/edns0.go:110-113`), which builds `new(dns.OPT)` setting only `Hdr.Name` and
`Hdr.Rrtype` — leaving `Hdr.Class`, the UDP payload size, at zero.

RFC 6891 §6.2.4 makes a peer read that zero as 512, so blocky was advertising a
512-byte receive buffer while its listener accepts 65535 (`server.go:38`). The commit
repairs the size in that branch and leaves a valid upstream OPT untouched.

I also added a DO=false spec. The existing "client left the DO bit clear" cases all
use `chainAddingEdns0` / `chainValidatingDNSSEC`, which put an OPT on the response —
so they exercise the mirroring branch and left the new one uncovered. Hardcoding
`SetEdns0(…, true)` kept every spec green before this.

### Follow-ups — not for this PR

These are all pre-existing and equally broken before and after your change, so I
don't want to grow the scope. Filing them separately:

- **`normalizeResponse` isn't actually the last step before the wire.** `handleReq`
  writes the SERVFAIL response directly (`server.go:803`), as does `OnHealthCheck`
  (`server.go:866`). So on an upstream timeout or a recovered panic the reply still
  goes out with no OPT and `RecursionAvailable` unset — the same downgrade signal,
  on the path where a struggling upstream makes it most likely to fire. The invariant
  wants to live on a shared write path.
- **Extended rcodes break when the OPT is stripped.** In the `!q.hadEdns0` branch,
  `util.RemoveEdns0Record` on a response with Rcode > 15 makes `Msg.Pack()` fail with
  `dns: bad extended rcode` and return a zero-length buffer. An upstream answering
  BADVERS or BADCOOKIE to a non-EDNS0 client therefore produces no reply at all.
- **The DO bit isn't part of the cache key.** `GenerateCacheKey` (`util/common.go:223`)
  keys on qtype + name only, so a DO=0 client can populate the cache with an
  RRSIG-less answer that a later DO=1 client now receives with `DO=1, AD=0`. This one
  *is* new behavior from this PR — but those clients were already getting the EDNS0
  downgrade you're fixing, so it trades one broken behavior for another in a narrow
  subcase rather than regressing something that worked. The root cause is the cache key.

Smaller things I noticed and am not going to block on: the synthesized OPT doesn't
mirror an ECS option back on cache hits (RFC 7871 §7.3 — not a regression, the OPT was
absent entirely before), the extra 11 bytes can flip borderline answers to TC=1 for
clients advertising 512 (correct per RFC 6891, just untested at that boundary), and we
now have five separate declarations of the EDNS0 buffer size across the resolver
packages.

Nice catch on this one.

0xERR0R (approved):

Issue #1980, opened 2026-01-28
'Upstream Groups' configuration


G-Huber:
Under https://0xerr0r.github.io/blocky/latest/configuration/#upstream-groups documentation states:

> Each resolver must be defined as a string in following format: `[net:]host:[port][/path][#commonName]`.

But resolvers defined like this without port cannot be parsed. Actually it has to be `[net:]host[:port][/path][#commonName]`.
Also the `path` parameter is not described in the table below at all.

I'd happily create a PR to fix that - but I'm not quite clear on one thing:
I believe that `path` only applies to `net` type `https` but can currently be specified for other types, where it will be ignored.

Is this correct? Startup logs don't make this clear with a warning.

PR #2241, opened 2026-08-28
docs: Fix upstream notation and mention config folder use with container

G-Huber:
Mention that the container env var also accepts a config directory
Fix multiple instances of slightly wrong upstream notation description - as mentioned in #1980

### DIFF
--- a/config/upstream.go
+++ b/config/upstream.go
@@ -87,7 +87,7 @@ func (u *Upstream) UnmarshalText(data []byte) error {
 	return nil
 }
 
-// ParseUpstream creates new Upstream from passed string in format [net]:host[:port][/path][#commonname]
+// ParseUpstream creates new Upstream from passed string in format [net:]host[:port][/path][#commonname]
 // or DNS Stamp format: sdns://...
 func ParseUpstream(upstream string) (Upstream, error) {
 	// Check if it's a DNS stamp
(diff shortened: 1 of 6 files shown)


0xERR0R:
Thanks for this — the notation fix is correct, I verified it against the parser (`1.1.1.1` parses, `1.1.1.1:` does not, so the old `host:[port]` was indeed wrong).

One part was missing: `docs/config.schema.json` is a generated file (see the `go:generate` directive in `config/config.go`), so the edit there would have been reverted by `go generate` and `make generate-check` would have failed in CI. I've pushed a commit that moves the fix to its source in `tools/schemagen/overrides.go` — the generated schema is unchanged, it just reproduces correctly now.

While in there I enhanced the docs a bit:

- `docs/config.yml` listed `tcp` and `udp` as valid `net` values, but neither actually parses — replaced with the real set (tcp+udp, tcp-tls, https, quic) and added quic's default port.
- Documented the `path` parameter, the second point you raised in #1980. To answer your question there: yes, it only applies to `https` — it's consumed by the DoH client, while the tcp/tls and quic clients discard it. So it is silently ignored elsewhere, as you suspected.
- Fixed the same stale notation in the `ParseUpstream` doc comment.

`go build`, `go test ./config/` and `generate-check` all pass. Happy to have you review the extra commit.

0xERR0R (approved):

PR #2242, opened 2026-08-28
fix(querylog): exclude sqlite on Solaris and illumos

drscream:
At the moment there is no pure-Go driver for SQLite available on illumos. For that reason disable it.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1155,7 +1155,13 @@ You can select one of following query log types:
 The `sqlite` target stores the query log in a single local file (set via `queryLog.target`, e.g. `/var/lib/blocky/querylog.db`) — no external database is required. Blocky creates the file and its parent directory automatically.
 
 !!! note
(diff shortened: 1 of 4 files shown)


0xERR0R:
Thanks! Reproduced the build failure and confirmed the fix — `modernc.org/libc` has no illumos port, so the driver chain cannot compile there.

I pushed three follow-ups to your branch:

- **fac9b203** — `database_writer_test.go` carries its own copy of the constraint and was missed by #2123, so `go vet ./querylog/` still failed on `linux/loong64`. Pre-existing, unrelated to your change.
- **d8e0e900** — widened `illumos` to `solaris`. `modernc.org/sqlite` has no Solaris port either, and Go satisfies the `solaris` build tag for `GOOS=illumos` as well, so one term covers both platforms. Applied to the test file too.
- **487a0d72** — brought the platform list in `docs/configuration.md` in line with the build constraint.

Verified `build` + `vet` clean on illumos, solaris, linux/mips, linux/loong64 and every current release target, with exactly one of the two implementation files selected per platform.

0xERR0R (approved):

PR #2260, opened 2026-09-11
fix(redis): recover optional connections after startup failure

SisyphusMD:
When Redis is unavailable during startup with `required: false`, Blocky continues serving DNS but discards the Redis client. Starting Redis afterward does not restore cache sharing or blocking-state synchronization.

Retain the client and start the cache and event subscriptions so they can reconnect. Skip the initial cache load after a failed connection check, and preserve startup errors when Redis is required.

Add regression tests for cache writes and blocking-state events after Redis starts, optional startup without repeated blocking checks, and client cleanup when a required connection check or subscription fails.

Validation on Linux/amd64 with Go 1.26.2: build, regular tests, race tests, lint and generated-file checks passed. GoReleaser 2.18.1 configuration validation passed. The required-Ping cleanup regression also fails when the client close is removed.

Docker build and end-to-end targets could not run locally: unchanged upstream and the patch both stop at the missing Docker prerequisite. Those checks still need CI.

### DIFF
--- a/cache/redis.go
+++ b/cache/redis.go
@@ -50,6 +50,8 @@ type RedisOptions[T any] struct {
 	FlushInterval time.Duration
 	// SendBufSize is the capacity of the internal send buffer channel.
 	SendBufSize int
+	// SkipInitialLoad avoids another blocking dial when Redis already failed its startup check.
+	SkipInitialLoad bool
 }
 
 // ReloadPublishable is implemented by inner caches that reload entries internally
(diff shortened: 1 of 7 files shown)


SisyphusMD:
Addressed all three review findings in separate commits:

- Preserved the two-argument `NewEventBusBridge` constructor and added an options constructor for background connection.
- Cancel and close acquired Redis resources when later server initialization fails.
- Move Redis publication out of the local event callback, with one worker and one pending latest state. Queued timed disables retain only their remaining duration; expired ones are discarded.

Regression tests cover constructor compatibility, startup cleanup, slow publication, pending-state replacement, duration expiry, and cancellation.

